PR #12, opened 2015-02-26
Adding collectinfo to asadm

kportertx:
Features are currently at par with asmonitor's collectinfo


lib/util.py
```diff
@@ -122,3 +123,29 @@ def shell_command(command):
...
+# ## Many of the view functions just do a print without any valid returns
+# ## This is to capture such prints so that they can be used as returns
+def capture_stdout(func,line=[''],printMsg=True):
+    name = ''
+    if printMsg is True:
+        try:
+            name = func.func_name
+        except Exception:
+            pass
+        print "Data collection for " + name +str(line) + " in progress.."
```
kportertx: Please keep this function generic by not having prints specific to collectinfo

lib/util.py
```diff
@@ -122,3 +123,29 @@ def shell_command(command):
...
+        sys.stdout.flush()
+    old = sys.stdout
+    capturer = StringIO.StringIO()
+    sys.stdout = capturer
+    print '===ASCOLLECTINFO==='
```
kportertx: Please keep this function generic by not having prints specific to collectinfo

kportertx: Please keep this function generic by not having prints specific to collectinfo

lib/util.py
```diff
@@ -122,3 +123,29 @@ def shell_command(command):
...
+    print name + str(line)
```
kportertx: Please keep this function generic by not having prints specific to collectinfo

lib/util.py
```diff
@@ -122,3 +123,29 @@ def shell_command(command):
...
+    o = func(line)
+    if isinstance(o,object):
```
kportertx: Should be a space after the comma

lib/util.py
```diff
@@ -122,3 +123,29 @@ def shell_command(command):
     else:
         return out, err
 
+# ## Many of the view functions just do a print without any valid returns
```
kportertx: I think all view functions behave this way, it is a CLI application the views are supposed to do the rendering.

lib/util.py
```diff
@@ -122,3 +123,29 @@ def shell_command(command):
...
+# ## This is to capture such prints so that they can be used as returns
```
kportertx: Could you please follow the function documentation pattern in this file?

lib/controller.py
```diff
@@ -521,3 +523,97 @@ def do_dun(self, line):
...
         results = self.cluster.infoUndun(self.mods['line'], nodes=self.nodes)
         self.view.dun(results, self.cluster, **self.mods)
+
+@CommandHelp('"collectinfo" is used to collect system stats on the local node.')
+class CollectinfoController(CommandController):
+    def collect_local_file(self,src,dest_dir):
+        shutil.copy2(src, dest_dir)
+        
+    def write_log(self,collectedinfo,src_file='/var/log/aerospike/*log'):
+        aslogdir = '/tmp/as_log_' + str(time.time())
```
kportertx: Should we be using https://docs.python.org/2/library/tempfile.html

lib/controller.py
```diff
@@ -14,6 +14,7 @@
 
 from lib.controllerlib import *
 from lib import util
+import time,os,sys,platform,shutil
```
kportertx: There should be spaces after the commas

lib/controller.py
```diff
@@ -521,3 +523,97 @@ def do_dun(self, line):
...
+@CommandHelp('"collectinfo" is used to collect system stats on the local node.')
+class CollectinfoController(CommandController):
+    def collect_local_file(self,src,dest_dir):
+        shutil.copy2(src, dest_dir)
+        
+    def write_log(self,collectedinfo,src_file='/var/log/aerospike/*log'):
+        aslogdir = '/tmp/as_log_' + str(time.time())
+        aslogfile = aslogdir + '/ascollectinfo.log'
+        os.mkdir(aslogdir)
+        f = open(str(aslogfile), 'w')
```
kportertx: We are supporting python 2.6.6+ which is shipped with centos6, so you can use the with operator here.

lib/controller.py
```diff
@@ -521,3 +523,97 @@ def do_dun(self, line):
...
+        f.close()
+        self.collect_local_file(src_file,aslogdir)
+        util.shell_command(["tar -czvf " + aslogdir + ".tgz " + aslogdir])
+        sys.stderr.write("\x1b[2J\x1b[H")
+        print "\n\n\nFiles in " + aslogdir + " and " + aslogdir + ".tgz saved. Please send tgz archive to support@aerospike.com"
+        print "END OF ASCOLLECTINFO"
+
+    def do_collectinfo(self, line):
+        capture_stdout = util.capture_stdout
+        collect_output = ''
```
kportertx: Making 2 assignments to collect_output, first assignment isn't used.

lib/controller.py
```diff
@@ -521,3 +523,97 @@ def do_dun(self, line):
...
+                      'dmesg',
+                      'iostat -x 1 10',
+                      'vmstat -s',
+                      'vmstat -m',
+                      'iptables -L',
+                      'cat /etc/aerospike/aerospike.conf',
+                      'cat /etc/citrusleaf/citrusleaf.conf',
+                      ]
+
+        def collect_sys(self):
```
kportertx: Why is this a nested function?

lib/controller.py
```diff
@@ -521,3 +523,97 @@ def do_dun(self, line):
...
+                      'ls /sys/block/{sd*,xvd*}/queue/rotational |xargs -I f sh -c "echo f; cat f;"',
+                      'ls /sys/block/{sd*,xvd*}/device/model |xargs -I f sh -c "echo f; cat f;"',
+                      'lsof',
+                      'dmesg',
+                      'iostat -x 1 10',
+                      'vmstat -s',
+                      'vmstat -m',
+                      'iptables -L',
+                      'cat /etc/aerospike/aerospike.conf',
+                      'cat /etc/citrusleaf/citrusleaf.conf',
```
kportertx: asadm currently only supports 3.x, we could probably leave out the 2.x files.

kportertx:
@anshprat Cleaned up your commits here, please review these notes.

lib/controller.py
anshprat: >>> aslogdir
'as_log_1425024563.54'
>>> tempfile.mkdtemp('',aslogdir)
'/tmp/as_log_1425024563.54_iu1Uq

The suffix makes it non-standard to be used.

lib/controller.py
anshprat: Will we have any specific gain from moving back to with?

lib/util.py
anshprat: done, changed printMsg to collectinfo_msg and moved the other print within the if check block.

lib/util.py
anshprat: done

lib/util.py
anshprat: done

lib/util.py
anshprat: done

lib/util.py
anshprat: updated comment.

lib/util.py
anshprat: done

lib/controller.py
anshprat: done

lib/controller.py
anshprat: yes, this was more erring on side of precaution, just in there is a failure later on. fixed it now.

lib/controller.py
anshprat: don't need it outside of this.

lib/controller.py
anshprat: I would still leave the 2.x stuff around. helps in cases where 2x->3x migration has happened.

lib/util.py
kportertx: It still shouldn't be here.

lib/controller.py
kportertx: https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile

You can pass a suffix.

lib/controller.py
kportertx: Not really, just know before you wanted to use it, but our python version to support was 2.4 or 2.5 in that project.

lib/controller.py
kportertx: Even so we should do it in a way that doesn't generate not founds. Otherwise it is confusing to the users. I've seen this a couple times with asmonitor's version.

lib/controller.py
kportertx: No problems, just bad style (pep20).

lib/controller.py
anshprat: The above output is after the suffix is passed, the empty string is the suffix.

lib/controller.py
anshprat: generate not founds? where does it generate not found except for in the log file?

lib/util.py
anshprat: well, function can be used now outside of collectinfo usage without any collectinfo specific prints. the other way would be to pass the whole msg all the time. but considering that its collectinfo specific usage and function right now, I would say its good.

lib/controller.py
kportertx: It generates the not found in the collectinfo file. I have seen it as a point of confusion for users.

PR #15, opened 2015-07-31
AER-3455 added dmesg with timestamp in collectinfo

PavanGupta01:
added -T with dmesg in controller.py for printing timestamp with kernel logs in collectinfo
Unfortunately timestamp can not be printed in Centos with dmesg, storing dmesg logs without timestamp for this OS.
code tested on ubuntu(12.04, 14.04), Debian 6, CentOS(5,6).
Please review the code and share your thoughts for this code change.


kportertx:
@PavanGupta01 Had to revert this change, please do pull requests against develop, not master. Master is for released changes.

PR #17, opened 2015-08-03
AER-3587 made cluster dun all command version aware

PavanGupta01:
Added version awareness for cluster dun/undun command.
Tested on ubuntu so far, testing with other flavours of Linux.
Meanwhile could you please have a look and let me know if we can improve the code further.

### DIFF
--- a/lib/node.py
+++ b/lib/node.py
@@ -19,6 +19,7 @@ from telnetlib import Telnet
 from time import time
 import socket
 import threading
+from distutils.version import StrictVersion
 
 def getfqdn(address, timeout=0.5):
     # note: cannot use timeout lib because signal must be run from the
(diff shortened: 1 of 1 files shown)


lib/node.py
```diff
@@ -453,9 +454,17 @@ def _parseDunList(self, dun_list):
         c = lib.cluster.Cluster(self.ip)
         lookup = c.node_lookup
         result = set()
-
+        
         if 'all' in dun_list:
-            result.add('all')
+            _serviceIP, _servicePort = self._serviceIPPort.split(':')
+            as_version = citrusleaf.citrusleaf_info(_serviceIP, _servicePort)['build']
```
kportertx: You should use self.info instead.

PavanGupta01:
Thanks Kevin, self.info instead of citrusleaf_info makes more sense, I have updated the code.

PR #21, opened 2015-09-23
[AER-3587] Replaced set comprehension for compatibility

PavanGupta01:
Replaced set comprehension with loop for python version compatibility.


samkarande:
Fix for asadm crash

PR #30, opened 2017-04-05
Adding support for non tty input stream

volmarl:
asadm uses getpass as one of its component. In order to support automation where password could come from a script and not a terminal, we can also check for the type used as input stream. For tty we keep using getpass() and for non tty we can use sys.stdin.readline()

This will allow scripts to be able to send password outside of terminal and clear text passwords not to be visible through a ps.


citrusraj:
Fixed in release 0.1.10

PR #40, opened 2018-04-12
Set default connect host to 127.0.0.1

kustodian:
If IPv6 is enabled and there is no IPv6 address set on a host, `asadm` cannot connect to `localhost`, because it probably tries to connect to an IPv6 host which isn't possible. More details in this thread https://discuss.aerospike.com/t/asadm-h-localhost-doesnt-work-in-version-0-1-17/5016/8

I set the default to 127.0.0.1, I also removed [::1] because if Aerospike isn't listening on IPv6 it displays the IPv6 nodes as offline.

### DIFF
--- a/pkg/astools.conf
+++ b/pkg/astools.conf
@@ -19,7 +19,7 @@
 #------------------------------------------------------------------------------
 #
 [global]
-host = "localhost:3000,[::1]:cluster_a:3000"                     # host = [ <host>[:<tls-name>][:<port>], ...]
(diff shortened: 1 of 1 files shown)


hbpatre:
We have already fixed this issue internally. We have commented all entries in conf file. So no new parameter will get assigned through conf without users knowledge. And default values will continue to work as old behaviour. host default is 127.0.0.1. These changes will be published in next release once done with all testing.

PR #47, opened 2021-06-21
Fix a sed command in Make to be GNU/BSD and zsh compatible

jfwm2:
- The sed command in question did not appear to work with zsh
 - Also, the previous version contained two versions of the same command, one seemed BSD sed compatible, the other one with gsed; this was replaced by a single command that should work with both sed options and also with zsh

Signed-off-by: Jean-Francois Weber-Marx <jf.webermarx@criteo.com>
Signed-off-by: Jean-Francois Weber-Marx <jfwm@hotmail.com>


jdogmcsteezy:
@jfwm2, I apologize for taking so long to thank you.  As you may know, our development takes place in a private repo so this did not receive attention.  However, going forward we will develop in the public repo

Thanks,
Jesse

PR #51, opened 2022-04-26
2.7.3 bugfix

jdogmcsteezy:
- Fixed issue with parsing the `service` response from the server.
- A hidden `--debug` mode was also added.


dwelch-spike (approved):
LGTM

PR #58, opened 2022-06-30
Include all obvious keyboard chars as symbols

rbotzer:


jdogmcsteezy (commented):
We can't allow ';' since that is how asadm understands the separation of cmds.  Ill have to look at the implications of the rest of this.  

Another concern is that the user may have to do some escaping to login with some of these possible passwords.  I am not sure if this matters but it will catch somebody up

jdogmcsteezy:
Closing.  Included in #62

PR #60, opened 2022-08-02
TOOLS-2079 Support dns load-balancer

jdogmcsteezy:
Asadm currently supports a proxy load-balancer in which an LB routes each request to a new node and proxies the node response.  When a DNS LB is used it is the responsibility of the DNS server to return a new IP every time the lookup is done.  The former was handled by making a single request to the node for all the needed connection information.  However, we were less careful when getting the IP and fqdn.  This caused the initial connection to happen on one node while the DNS lookup responded with a different node.  This eventually caused a later node to appear already in the cluster, triggering an eviction.

### DIFF
--- a/lib/live_cluster/client/cluster.py
+++ b/lib/live_cluster/client/cluster.py
@@ -351,6 +351,7 @@ class Cluster(AsyncObject):
                         if node is not None
                     ]
                 )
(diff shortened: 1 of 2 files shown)


dwelch-spike (approved):
LGTM, remember to ask QE for a scenario for this

PR #62, opened 2022-08-19
Tools 2063 sindex cdt

jdogmcsteezy:
[TOOLS-2066](https://aerospike.atlassian.net/browse/TOOLS-2066) (ASADM) Add sindex cardinality to `show sindex` and `info sindex` commands

[TOOLS-2063](https://aerospike.atlassian.net/browse/TOOLS-2063) (ASADM) Add the ability to create sindexs on CDTs by providing a context to the `manage sindex create` command.

[TOOLS-2061](https://aerospike.atlassian.net/browse/TOOLS-2061) (ASADM) Accept all characters for passwords except `;` and whitespace.

### DIFF
--- a/lib/live_cluster/client/ctx.py
+++ b/lib/live_cluster/client/ctx.py
@@ -1,0 +1,106 @@
+from typing import Any, Generic, TypeVar
+
+T = TypeVar("T")
+
+
+class ASValue(Generic[T]):
+    def __init__(self, value: T):
+        self.value = value
+
+
+class ASValues:
+    class ASBool(ASValue[bool]):
+        def __init__(self, value: bool):
+            super().__init__(value)
+
+    class ASInt(ASValue[int]):
+        def __init__(self, value: int):
+            super().__init__(value)
+
+    class ASString(ASValue[str]):
+        def __init__(self, value: str):
+            super().__init__(value)
+
+    class ASList(ASValue[list[ASValue]]):
+        def __init__(self, value: list[ASValue]):
+            super().__init__(value)
+
+    class ASMap(ASValue[dict[ASValue, ASValue]]):
+        def __init__(self, value: dict[ASValue, ASValue]):
+            super().__init__(value)
+
+    class ASPair(ASValue[tuple]):
+        def __init__(self, value: tuple):
+            super().__init__(value)
+
+    class ASBytes(ASValue[bytes]):
+        def __init__(self, value: bytes):
+            super().__init__(value)
+
+    class ASDouble(ASValue[float]):
+        def __init__(self, value: float):
+            super().__init__(value)
+
+    class ASGeoJson(ASValue[str]):
+        def __init__(self, value: str):
+            super().__init__(value)
+
+    class ASWildCard(ASValue[None]):
+        def __init__(self):  # maybe could be dict?
+            super().__init__(None)
+
+    # class ASUndef(ASValue):
+    #     def __init__(self):
+    #         super().__init__(None)
+
+    # class ASNil(ASValue):
+    #     def __init__(self):
+    #         super().__init__(None)
+
+    # class ASRec(ASValue):
+    #     def __init__(self, value):
+    #         super().__init__(value)
+
+
+class CTXItem(Generic[T]):
+    def __init__(self, val: T):
+        self.value = val
+
+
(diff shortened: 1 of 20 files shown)


lib/live_cluster/client/ctx.py
```diff
@@ -0,0 +1,122 @@
...
+    def __str__(self):
+        return "{}({})".format(type(self), self.value)
+
+
+class CTXIntItem(CTXItem[int]):
+    def __init__(self, val: int):
+        if not isinstance(val, int):
+            raise TypeError("CTX value must of type int")
+
+        super().__init__(val)
```
dwelch-spike, line 87: why have int and ASInt ctx types?

lib/live_cluster/client/msgpack.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def pack(self, obj):
+        if isinstance(obj, ASValue):
+            self._pack_as_value(obj)
+            return
+        elif isinstance(obj, CDTContext):
+            self._pack_as_cdt_ctx(obj)
+            return
+
+        super().pack(obj)
```
dwelch-spike, line 37: what gets handled by this default case?

lib/live_cluster/client/msgpack.py
```diff
@@ -0,0 +1,107 @@
...
+            val = obj.value
+            val = chr(AS_BYTES_STRING) + val
+            self.pack(val)
+            return
+
+        if isinstance(obj, ASValues.ASBytes):
+            val = obj.value
+            val = chr(AS_BYTES_BLOB) + val.decode("utf-8")
+            self.pack(val)
+            return
```
dwelch-spike, line 81: Why are there so many ASValue types defined, i.e. bool, int, float, etc if only ASString and ASBytes are used?

lib/live_cluster/client/types.py
```diff
@@ -248,3 +249,6 @@ class ASInfoNotAuthenticatedError(ASInfoError):
 
 class ASInfoClusterStableError(ASInfoError):
     pass
+
+
+SIndexBinType = Union[Literal["NUMERIC"], Literal["GEO2DSPHERE"], Literal["STRING"]]
```
dwelch-spike, line 254: Where is this used? Is it needed?

dwelch-spike (commented):
This is cool 😎 Looks good I just have some questions for my understanding.

lib/live_cluster/client/ctx.py
jdogmcsteezy: There two types of Ctx items.  Ones that accept ints and ones that accpet asvalues (which includes asints).  There is no ASInt ctx type

lib/live_cluster/client/msgpack.py
jdogmcsteezy: All msgpack native types.  Almost all roads lead to this base case since self.pack() is recursive.

lib/live_cluster/client/msgpack.py
jdogmcsteezy: It is about being very explicit about what asvalues we except and which ones we do not. This is also important if we except other types like geo json or maps in the future.

jdogmcsteezy: I understand that the is a way to do this without all these as types but the way it is done is very explicit

lib/live_cluster/client/types.py
jdogmcsteezy: It is not used.  I was trying to use it somewhere, but it caused some IntelliSense problems.  It is just a typing

lib/live_cluster/client/msgpack.py
jdogmcsteezy: Also seems weird to me to have ASString and ASBytes but not the others.  There also might be a reason for some of this in the future.  Also the c-client differentiates these so I figured it might be a good call

jdogmcsteezy: O no, 3 responses.  That aint good

dwelch-spike: Ok I think I'm just having trouble telling the layers of types apart. This works for me though.

dwelch-spike (approved):
LGTM

Issue #63, opened 2022-08-22
Please fix processing hash from filename for success to add log_file


kan-fumihito:
I try using log-analyzer by following command.
```bash
$ file -i aerospike.log
aerospike.log: text/plain; charset=us-ascii
$ asadm -lf aerospike.log
Seed:        [('localhost', 3000, None)]
Config_file: /home/ubuntu/.aerospike/astools.conf, /etc/aerospike/astools.conf
Aerospike Log Analyzer Shell, version 2.4.0

ERROR: Unicode-objects must be encoded before hashing
```
I modified
```python
# https://github.com/aerospike/aerospike-admin/blob/master/lib/log_analyzer/log_handler/log_handler.py#L471
file_key = "MD5_" + str(hashlib.md5(log_file).hexdigest())
```
to
```python
file_key = "MD5_" + str(hashlib.md5(log_file.encode("utf-8")).hexdigest())
```
then succeeded it.

Could you fix this?
thank you.

jdogmcsteezy:
Thanks for the issue.  I will fix this.  Keep an eye out for the next release

jdogmcsteezy:
@kan-fumihito, you can download the latest version here https://aerospike.com/download/#aerospike-tools

PR #64, opened 2022-08-22
Bugfix 2.8.1

jdogmcsteezy:
[TOOLS-2097](https://aerospike.atlassian.net/browse/TOOLS-2097) (ASADM) LogAnalyzer fails when trying to hash logfile before it is converted to bytes.

Collectinfo cmd does not correctly restore prompt when finished.


jdogmcsteezy:
#63

PR #70, opened 2022-10-18
TOOLS-2220 Bug in the `with` modifier logic causes ip to not match

jdogmcsteezy:
When the exact ip of a node is provided to the with modifier and there exists another node with an almost identical ip except for the last octet where the provided ip is a prefix of the other then an error will occur indicating that no matching nodes could be found.

Example:
User provides `with 1.1.1.1` and the list of available nodes is (1.1.1.11:3000, 1.1.1.1:3000) the correct response is  [1.1.1.1:3000] but with this bug an empty list would be returned.

### DIFF
--- a/lib/live_cluster/client/cluster.py
+++ b/lib/live_cluster/client/cluster.py
@@ -76,7 +76,7 @@ class Cluster(AsyncObject):
 
         # self.node_lookup is a dict of (fqdn, port) -> Node
         # and (ip, port) -> Node, and node.node_id -> Node
-        self.node_lookup = LookupDict(LookupDict.PREFIX_MODE)
+        self.node_lookup: LookupDict[Node] = LookupDict(LookupDict.PREFIX_MODE)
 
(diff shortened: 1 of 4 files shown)


lib/live_cluster/client/cluster.py
```diff
@@ -415,25 +415,40 @@ def get_node(self, node) -> list[Node]:
...
+                    match = n
+                else:
+                    # more than one match
+                    return []
 
-        return []
+        if match is None:
+            return []
+        else:
+            # more than one match
```
dwelch-spike, line 450: this is "one or more matches" right?

dwelch-spike (approved):
LGTM, one minor question

lib/live_cluster/client/cluster.py
jdogmcsteezy: This comment should be changed to, "only one exact match found"

PR #75, opened 2022-12-08
ci: Add dependabot

jdogmcsteezy:


arrowplum (approved):
While I have some strong concerns I think this looks acceptable

PR #81, opened 2022-12-09
asadm 2.11.1

jdogmcsteezy:
### [TOOLS-2286](https://aerospike.atlassian.net/browse/TOOLS-2286) \(ASADM\) Bundle asadm with Python 3.10.


jdogmcsteezy:
unittest2 is a backport of unittest which is included in the python stdlib.  The backport is to support older python versions but because we are able to lockdown our python version it is no longer required.  Also, unittest2 does not support 3.10 so it had to be removed

dwelch-spike (approved):
LGTM

PR #82, opened 2022-12-09
2.11.1 bugfix

jdogmcsteezy:
### [TOOLS-2286](https://aerospike.atlassian.net/browse/TOOLS-2286) \(ASADM\) Bundle asadm with Python 3.10.


dwelch-spike (approved):
LGTM

PR #86, opened 2022-12-15
TOOLS-2302 fix Mapping type import changed in python 3.10

jdogmcsteezy:


lib/utils/conf.py
```diff
@@ -21,6 +20,7 @@
 import toml
 import copy
 from jsonschema import validate
+from typing import Mapping
```
dwelch-spike, line 23: It looks like most of typing was deprecated in Python 3.9. https://docs.python.org/3/library/typing.html#typing.Mapping I think the preferred method is `from collections.abc import Mapping` which will fail for anything older than Python 3.9 So how about something like...
```
try:
    from collections.abc import Mapping
except ImportError:
    from collections import Mapping
```

dwelch-spike (changes requested):
Importing from typing is deprecated. Maybe we should use a different pattern. https://peps.python.org/pep-0585/

lib/utils/conf.py
jdogmcsteezy: Ah, did not realize.  We only use Python 3.10

jdogmcsteezy, line 23:
```suggestion
from collections.abc import Mapping
```

PR #87, opened 2022-12-15
docs: Specify python 3.10 requirement

jdogmcsteezy:


dwelch-spike (approved):
Should be against test-ready I think but otherwise LGTM

PR #103, opened 2023-02-03
TOOLS-2310 TOOLS-2353 Ignore trailing commas in list. Have consistent close commands with aql.

jdogmcsteezy:


lib/base_controller.py
```diff
@@ -480,9 +509,16 @@ def parse_modifiers(self, line, duplicates_in_line_allowed=False):
                 if duplicates_in_line_allowed or word not in groups[mod]:
                     groups[mod].append(word)
 
-        if "with" in mods and "all" in groups["with"]:
-            groups["with"] = "all"
+        # Remove ',' from input since it needs to be filtered
+        # out in many cases.
```
dwelch-spike, line 513: Suggestion: this comment makes me think all commas are removed, but it's only the trailing commas in a modifier list, right? Maybe update the comment?

dwelch-spike (approved):
LGTM

PR #105, opened 2023-02-11
fix: remove debug print, allow modifiers to pass between commands

jdogmcsteezy:
Issues found by QE.

### DIFF
--- a/lib/live_cluster/manage_controller.py
+++ b/lib/live_cluster/manage_controller.py
@@ -1,6 +1,8 @@
 import asyncio
 import base64
 import binascii
+import copy
+import inspect
 import os
 import logging
 from datetime import datetime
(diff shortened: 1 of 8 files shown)


lib/live_cluster/manage_controller.py
```diff
@@ -32,7 +32,47 @@
...
 
-class ManageLeafCommandController(LiveClusterCommandController):
+class LiveClusterManageCommandController(LiveClusterCommandController):
+    def pre_controller(self, line):
+        self._init_controller_arg()
+
+        if self.controller_arg is None:
+            return
+
+        mod = self.context[-1]
```
dwelch-spike, line 42: This only supports 1 mod right? Is there a possibility 2 could be passed in? like "for' and "with"?

lib/live_cluster/manage_controller.py
```diff
@@ -32,7 +32,47 @@
...
+
+        # Used as the key to reference arg, normally the controllers "command".
+        try:
+            arg = line.pop(0)
+        except IndexError:
+            raise ShellException("{} is required".format(mod))
+
+        if arg not in self.modifiers | self.required_modifiers:
+            if mod not in self.mods[mod]:
+                self.mods[mod] = []
```
dwelch-spike, line 55: Why does self.mods[mod] have a default value of {} 1st and then []?

lib/live_cluster/manage_controller.py
dwelch-spike: Food for thought, if modifiers being the second to last argument in context is common maybe you can make a getMods function out of this? Or make context a class? That way it is easier to tell what data is in context and where it goes.

dwelch-spike (commented):
A few questions mostly for my understanding. I had a hard time following LiveClusterManageCommandController.pre_controller(). If the comments require no attention that this LGTM.

lib/live_cluster/manage_controller.py
jdogmcsteezy: This would take some explaining beyond what is reasonable in a chat I think.  To be clear, I did not add this behavior.  I only relocated it as to not pollute other classes inheriting from `CommandController`.  Essentially, this mechanism is allowing required args (controller arg) provided to high-level controllers to be passed down to their children via mods.

lib/live_cluster/manage_controller.py
jdogmcsteezy: fixed. There were a number of things wrong here.  Not sure when that was added

jdogmcsteezy: There was actually a larger problem here. I am glad you pointed it out @dwelch-spike or I would have missed it.

PR #113, opened 2023-02-15
Tools 2329 snyk results

jdogmcsteezy:


.github/workflows/snyk-test.yml
```diff
@@ -0,0 +1,34 @@
+name: "Snyk Test"
```
dwelch-spike, line 1:
```suggestion
name: "Snyk Report"
```

dwelch-spike (changes requested):
I think we should rename this action, otherwise looks good.

dwelch-spike (approved):
LGTM

PR #121, opened 2023-02-28
Tools-2350 Tools-2424 add stop write size to info set

jdogmcsteezy:


lib/view/templates.py
```diff
@@ -69,24 +69,44 @@ def project_xdr_req_shipped_errors(s, esc, ess):
...
+    """
+    values_total = 0.0
+    weighted_total = 0.0
 
     for w, v in zip(weights, values):
-        weighted_avg += w * v
-        total += v
+        weighted_value = w * v
+        weighted_total += weighted_value
+        values_total += v
```
dwelch-spike, line 104: I thought weighted averages divide by the sum of the weights in the end. Please double check that this is correct.

lib/view/templates.py
```diff
@@ -505,6 +569,49 @@ def set_index_projector(enable_index, index_populating):
...
+                ),
+                Field(
+                    "Used%",
+                    Projectors.Div(
+                        Projectors.Sum(
+                            Projectors.Number(
+                                "set_stats", "memory_data_bytes", "n-bytes-memory"
+                            ),
+                            Projectors.Number(
+                                "set_stats", "device_data_bytes", "n-bytes-memory"
```
dwelch-spike, line 592: should this be "n-bytes-device"?

lib/view/templates.py
```diff
@@ -1458,8 +1560,8 @@ def numeric_sum_aggregator_selector(key, is_numeric):
 
 
 def latency_weighted_avg(edatas: list[EntryData]):
-    pcts = map(lambda edata: edata.value, edatas)
-    ops_sec_values = map(lambda edata: edata.record["ops/sec"], edatas)
+    ops_sec_values: map[float] = map(lambda edata: edata.record["ops/sec"], edatas)
```
dwelch-spike, line 1563: is this a valid type hint? isn't the result of map() a list?

lib/view/templates.py
```diff
@@ -263,67 +290,103 @@ def usage_weighted_avg(edatas: list[EntryData]):
...
             ),
         ),
         Subgroup(
             "Memory",
             (
+                Field(
+                    "Total",
+                    Projectors.Number("ns_stats", "memory-size", "total-bytes-memory"),
+                    hidden=True,
+                ),
```
dwelch-spike, line 336: Why is this hidden?

dwelch-spike (changes requested):
A couple questions, mostly want to make sure the averaging is correct.

lib/view/templates.py
```diff
@@ -69,24 +69,44 @@ def project_xdr_req_shipped_errors(s, esc, ess):
...
 
+def _ignore_zero(num: int):
+    if num == 0:
+        return None
+
+    return num
+
+
+def _ignore_null(s: str):
+    if s == "null":
```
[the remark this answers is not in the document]
jdogmcsteezy: We do not but I can make it that way.

lib/view/templates.py
jdogmcsteezy: renamed the variables to match the equation you saw

lib/view/templates.py
jdogmcsteezy: good catch

lib/view/templates.py
jdogmcsteezy: It is valid. A map returns a generator like object called a map

lib/view/templates.py
jdogmcsteezy: So that it is not displayed. I am following precedence here. If somebody wants the total they can open a ticket

jdogmcsteezy: The value is needed for the weighted average calculation.  That does not mean it should be displayed

lib/view/templates.py
jdogmcsteezy: That's what happens when you derive the equation yourself haha.  I was thinking the pct should be considered the weight and the bytes should be considered the values but in more abstract terms the values should be the same unit of the result you are expecting

dwelch-spike (approved):
LGTM

PR #122, opened 2023-02-28
fix: `info sindex` formatting

jdogmcsteezy:
The stats need to be flipped one more time for them to be correctly displayed.  Instead of adding another flip I removed a previous flip in live and collectinfo modes which had the same effect.


dwelch-spike (approved):
LGTM

PR #123, opened 2023-03-03
Tools-2319 Tools-2434 add sindex usage and type to `info sindex` and `info namespace`

jdogmcsteezy:


lib/collectinfo_analyzer/get_controller.py
```diff
@@ -147,31 +153,25 @@ class GetStatisticsController:
...
+        self.log_handler = log_analyzer
+
+    def get_namespace(self):
+        return self.log_handler.info_statistics(stanza=constants.STAT_NAMESPACE)
+
+    def get_sindex(self):
+        return self.log_handler.info_statistics(stanza=constants.STAT_SINDEX)
 
     # TODO might be a good place to add support for the with modifier to filter nodes
-    def get_all(self):
```
jdogmcsteezy, line 153: not used

lib/view/templates.py
```diff
@@ -220,25 +220,6 @@ def usage_weighted_avg(edatas: list[EntryData]):
         namespace_field,
         node_field,
         hidden_node_id_field,
-        Field(
```
jdogmcsteezy, line 223: removed because it is also in `info namespace object` where it probably fits better

lib/view/templates.py
```diff
@@ -220,25 +220,6 @@ def usage_weighted_avg(edatas: list[EntryData]):
...
-                Projectors.Number("ns_stats", "master_tombstones"),
-                Projectors.Number("ns_stats", "prole_objects", "prole-objects"),
-                Projectors.Number("ns_stats", "prole_tombstones"),
-                Projectors.Number("ns_stats", "non_replica_objects"),
-                Projectors.Number("ns_stats", "non_replica_tombstones"),
-            ),
-            converter=Converters.scientific_units,
-            aggregator=Aggregators.sum(),
-        ),
-        Field(
```
jdogmcsteezy, line 236: diddo

lib/view/templates.py
```diff
@@ -361,6 +346,51 @@ def usage_weighted_avg(edatas: list[EntryData]):
...
+                            "sindex_flash_used_bytes",
+                            "sindex_pmem_used_bytes",
+                            "memory_used_sindex_bytes",
+                        ),
+                        Projectors.Number("ns_stats", "sindex-type.mounts-size-limit"),
+                    ),
+                    converter=Converters.round(2),
+                    aggregator=ComplexAggregator(
+                        create_usage_weighted_avg("Secondary Index"),
+                        converter=Converters.round(2),
```
dwelch-spike, line 378: I'm guessing this will need to change to use what is in https://github.com/aerospike/aerospike-admin/pull/121 ?

lib/view/view.py
```diff
@@ -299,19 +299,32 @@ def info_XDR(stats, xdr_enable, cluster, timestamp="", **ignore):
...
+        if not sindex_stats:
             return
 
+        # Insert sindex-type from ns_confgs into sindex_stats. Sheets can't handle
+        # two objects with differing schemas.
+        for node, node_stats in sindex_stats.items():
+            for key in node_stats:
+                ns = key.split()[0]
+
+                sindex_type = ns_configs.get(node, {}).get(ns).get("sindex-type", None)
```
dwelch-spike, line 315: No alternative value for .get(ns)?

dwelch-spike (approved):
This looks good. Just a few comments to keep in mind.

lib/view/templates.py
jdogmcsteezy: Correct

lib/view/view.py
jdogmcsteezy: It's a tradeoff between future proofing and backwards compatibility

PR #124, opened 2023-03-04
Minor fixes

jdogmcsteezy:
A few minor bugs. One of them does not have a ticket associated because I am unaware of a case where it would be visible to the user.


dwelch-spike (approved):
LGTM

PR #126, opened 2023-03-09
fix: live health check not checking namespace and sindex configs

jdogmcsteezy:
Issues found by Ramya in health check.


dwelch-spike (approved):
LGTM

PR #127, opened 2023-03-10
Tools 2372 dont use sudo if root

jdogmcsteezy:
There are two areas where system cmds are run. One is for running the cmds and parsing the output to json so the information can later be ingested into collectinfo mode. The other is just cmds pretty printed into sysinfo.log. 

I refactored the former area mentioned.  When you have a list of tuples with that many fields it is probably time to be an object.  I also made the object directly take a parsing function. Before, it was unclear that every new entry in the list also needed a parser function in sys_cmd_parser unless you read through the code a bit.  A lot of the logic removed was from a time when code unparsed cmd output needed to be parsed and then injected into the json object.  I think this was only done in collectinfo mode though.


lib/live_cluster/client/node.py
```diff
@@ -101,6 +102,56 @@ def wrapper(*args, **kwargs):
...
+
+    def parse(self, command_raw_output):
+        result = self._parse_func(command_raw_output)
+        sys_cmd_parser.type_check_basic_values(
+            result
+        )  # Not sure if this is still needed.
+        return result
+
+    def _is_root(self) -> bool:
+        return self._uid == 0
```
dwelch-spike, line 136: Should this be checking the class variable?

lib/live_cluster/client/node.py
```diff
@@ -165,58 +216,110 @@ async def __init__(
...
         self.sys_default_ssh_key = None
-        self.sys_cmds = [
-            # format: (command name as in parser, ignore error, command list)
-            ("hostname", False, ["hostname -I", "hostname"]),
-            ("top", False, ["top -n1 -b", "top -l 1"]),
-            (
+
+        _SysCmd.set_uid(os.getuid())
+        self.sys_cmds: list[_SysCmd] = [
+            # format: (command name as in parser, ignore error, sudo, command list)
```
dwelch-spike, line 222: This format looks wrong. Isn't it command_name, ignore_error, command_list, parser?

lib/utils/common.py
```diff
@@ -2340,31 +2339,40 @@ def _zip_files(dir_path, _size=1):
...
-        alt_dmesg = "sudo dmesg"
+        netstat_cmds = ["netstat"]
 
     # cmd and alternative cmds are stored in list of list instead of dic to
     # maintain proper order for output
 
-    sys_shell_cmds = [
+    sys_shell_cmds: list[list[str]] = [
         ["hostname -I", "hostname"],
         ["top -n3 -b", "top -l 3"],
```
dwelch-spike, line 2369: differs with the command used in node.py, not sure if that is ok.

lib/utils/common.py
```diff
@@ -2431,6 +2439,18 @@ def get_system_commands(port=3000):
...
+
+    """
+    Some distros and most containers do not have sudo installed by default. If running
+    as root don't use it. 
+    """
+    if uid == 0:
+        for cmd_list in sys_shell_cmds:
+            for idx, cmd in enumerate(cmd_list):
+                if "sudo " in cmd:
+                    cmd_list[idx] = cmd.replace("sudo ", "")
```
dwelch-spike, line 2452: Are these just here to document the commands used in node.py? Seems like having a single source of truth would be nice. Like a list of _sysCmd objects defined somewhere that can be used by node.py and be rendered as strings for functions like get_system_commands().

dwelch-spike (changes requested):
Want to make sure class/instance variables are being set correctly, see comment.

I don't understand this well but why is there a list of commands in node.py and common.py? If they are supposed to be the same list they should be defined in one place.

lib/live_cluster/client/node.py
jdogmcsteezy: Should not make a difference.

lib/live_cluster/client/node.py
jdogmcsteezy: Forgot to delete the comment

lib/utils/common.py
jdogmcsteezy: I explained in my first comment why they are different. Yeah, merging them could be a possibility one day but that day is not today

lib/utils/common.py
jdogmcsteezy: no

lib/live_cluster/client/node.py
dwelch-spike: If self._uid is ever set it will differ from _SysCmd._uid. Since this is not happening it should work but seems worth noting.

dwelch-spike (approved):
LGTM

PR #129, opened 2023-03-11
Tools 2380 rhel9

jdogmcsteezy:
These spec files are really just python files.  I did two things. The most important is that I now bundle `/usr/lib64/libcrypt.so.1` with asadm. The location of that file is system dependent and is only needed for rhel9 so I only add it if we are linux.  We build our Linux bundles on centos7 and that is where the file is located. The second thing I did was remove the entire `lib` directory from the distribution.  When I first created the files I could not figure out how to only include the schemas and excluded the all the source files while still keeping the paths as asadm expects them.  Anyway, I figured it out hence the addition to the array `datas`.

### DIFF
--- a/asadm-asinfo-one-dir.spec
+++ b/asadm-asinfo-one-dir.spec
@@ -13,11 +13,11 @@ from PyInstaller.utils.hooks import collect_all
 #
 
 datas = []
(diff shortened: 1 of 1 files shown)


dwelch-spike (approved):
LGTM

PR #132, opened 2023-03-21
Tools 2353 revert ctrl-c behavior

jdogmcsteezy:
There was a bug that caused an error to be displayed when a user used `quit` to exit. We are also reverting the change to how ctrl-c behaved to no anger users and to match expected interactive cmd line behavior


dwelch-spike (approved):
LGTM

PR #135, opened 2023-04-08
TOOLS-2448 `Security` feature not displayed in `features` and `summary` commands

jdogmcsteezy:


lib/utils/common.py
```diff
@@ -235,36 +98,391 @@ def _check_value(data={}, keys=()):
...
+
+        for field_check in self._field_checks:
+            if field_check.check(sources):
+                return True
+
+        return False
+
+    def __str__(self):
+        for field_check in self._field_checks:
+            return True
```
dwelch-spike, line 237: Is this incomplete?

lib/utils/common.py
```diff
@@ -235,36 +98,391 @@ def _check_value(data={}, keys=()):
...
+                "si_query_long_basic_complete",
+                "si_query_long_basic_error",
+                "si_query_short_basic_complete",
+                "si_query_short_basic_error",
+                "si_query_aggr_complete",
+                "si_query_aggr_error",
+                "si_query_udf_bg_complete",
+                "si_query_udf_bg_error",
+                "si_query_ops_bg_complete",
+                "si_query_ops_bg_error",
```
dwelch-spike, line 342: I don't think it will ever be the only query stats but there is also pi/si_query_abort

dwelch-spike (commented):
Just a couple non-blocking comments. If they seem fine to you then this LGTM

lib/utils/common.py
jdogmcsteezy: yes, good catch. Ill remove it. I was going to automate the internal docs creation but decided against it

dwelch-spike (approved):
LGTM

PR #143, opened 2023-04-17
fix: TOOLS-2484 `show latencies` aggregation issue

jdogmcsteezy:


lib/view/templates.py
```diff
@@ -1638,9 +1638,9 @@ def numeric_sum_aggregator_selector(key, is_numeric):
 
 
 def latency_weighted_avg(edatas: list[EntryData]):
-    ops_sec_values: map[float] = map(lambda edata: edata.record["ops/sec"], edatas)
-    pcts = map(lambda data: data[0].value * data[1], zip(edatas, ops_sec_values))
```
jdogmcsteezy, line 1642: This is the issue. This pre-processing step should have been removed when the `weighted_avg` func was changed. It is applying the weights beforehand.

dwelch-spike (approved):
LGTM

PR #159, opened 2023-05-30
chore: TOOLS-2470 add description to `show users` command.

jdogmcsteezy:
Also, change Scan to PI Query


dwelch-spike (approved):
LGTM
Please check tests before merging

PR #160, opened 2023-06-02
fix: TOOLS-2538 not able to parse collectinfo files

jdogmcsteezy:
Asadm 2.15.0 can't run commands that print a 'Node' column if the collectinfo file was not created with 2.15.0

### DIFF
--- a/test/e2e/live_cluster/test_all.py
+++ b/test/e2e/live_cluster/test_all.py
@@ -97,8 +97,8 @@ class TableRenderTests(unittest.TestCase):
         ("show statistics xdr dc"),
         ("show statistics xdr namespace"),
         ("show latencies -v"),
-        # ("show distribution time_to_live"), # TODO: Causing issues on github actions
-        # ("show distribution object_size"),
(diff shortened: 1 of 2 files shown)


test/e2e/live_cluster/test_all.py
```diff
@@ -97,8 +97,8 @@ class TableRenderTests(unittest.TestCase):
         ("show statistics xdr dc"),
         ("show statistics xdr namespace"),
         ("show latencies -v"),
-        # ("show distribution time_to_live"), # TODO: Causing issues on github actions
-        # ("show distribution object_size"),
+        ("show distribution time_to_live"),  # TODO: Causing issues on github actions
+        ("show distribution object_size"),
```
dwelch-spike, line 101: Do these still cause issues in Github actions?

jdogmcsteezy: I am hoping it is fixed. Only time will tell. The failures were intermittent before

dwelch-spike: ok

dwelch-spike (approved):
LGTM

PR #165, opened 2023-06-07
Tools 2325 keyboard interrupt except

jdogmcsteezy:
The main change involves configuring a signal handler on the event loop to cancel tasks when a keyboard interrupt is received. In the original code it wasn't easy to guarantee the KeyboardInterrupt exception would propagate up as you would expect a normal exception to in a sync environment. Setting a CancelError exception makes the exception propagate as you'd expect rather than boil up to the event loop. We can play with this more in a code review if you'd like.

The last change is a small refactor of the watch controller and view logic. I moved stripped out some of the execution code from the view.watch method and put it in the controller to match the rest of the code design. Now the view code only displays, and the controller loops and executes the desired function. There is also a slight refactor of the arg/flag parsing. You might want to choose the split view when diffing.

If you would like to test out the watch function just run asadm -e 'watch 1 info net'

### DIFF
--- a/lib/live_cluster/live_cluster_root_controller.py
+++ b/lib/live_cluster/live_cluster_root_controller.py
@@ -121,7 +121,6 @@ class LiveClusterRootController(BaseController, AsyncObject):
         "           interrupted",
         "           watch 5 info namespace",
     )
-    @DisableAutoComplete()
     async def do_watch(self, line):
         await self.view.watch(self, line)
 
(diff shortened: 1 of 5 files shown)


lib/live_cluster/live_cluster_root_controller.py
```diff
@@ -121,9 +125,59 @@ def do_help(self, line):
...
+                count += 1
+                await asyncio.sleep(sleep)
+
+        except asyncio.CancelledError:
+            return
+        finally:
+            sys.stdout = real_stdout
+            print("")
+
+        # await self.view.watch(self, line)
```
dwelch-spike, line 180: comment needed?

dwelch-spike (commented):
LGTM as long as the tests pass.

dwelch-spike (approved):
LGTM

PR #174, opened 2023-06-28
feat: TOOLS-2516 add check for stop writes at startup

jdogmcsteezy:
In the AerospikeShell init I no longer call logger.critical. Instead I set self.connected = False and let the caller handle the exit code. Other than that most the changes are related to type annotations


asadm.py
```diff
@@ -138,7 +140,8 @@ async def __init__(
                         "You have not specified any collectinfo path. Usage: asadm -c -f <collectinfopath>"
                     )
                     await self.do_exit("")
```
jdogmcsteezy, line 142: make sure connected is used in -cf mode

asadm.py
```diff
@@ -194,25 +196,41 @@ async def __init__(
...
-                        self.intro += (
-                            terminal.fg_red()
-                            + "Extra nodes in alumni list: %s"
+                        logger.warning(
+                            terminal.fg_yellow()
+                            + "Some nodes have become unreachable by other nodes in the cluster. Check their peers lists: %s"
                             % (", ".join(cluster_down_nodes))
                             + terminal.fg_clear()
-                            + "\n"
                         )
```
jdogmcsteezy, line 215: Check on coroutine RuntimeWarning un-awaited

jdogmcsteezy (commented):
Some changes needed after first review with @dwelch-spike

dwelch-spike (approved):
LGTM

dwelch-spike:
holding for tools 8.6

PR #176, opened 2023-07-01
feat: TOOLS-2446 support multiple privileges in the `manage acl create role` command

jdogmcsteezy:
This one is a bit of a doozy.  Hopefully, you can review it while I am gone. The old interface and the new interface were not really compatible, and Adam did not want any breaking changes since he believed this is a heavenly scripted API.  There is some hackery to make this possible that I am not very proud of mostly to do with the poor decision I made long ago to make `read` and `write` the keywords for providing read and write quotas even though there are `read` and `write` privileges.

There is a bit of work here to identify when someone is trying to provide a read or write privilege or a read or write quota, but it mostly boils down to if it is read/write followed by an int, then it is a quota. If it is a read/write followed by no int and it comes after the word `privs` and before any other "modifier" then it is a privilege.  

I also had to do some extra work to change the keyword `priv` to `privs` as well as still allow the user to provide the `ns` and `set` modifiers.

Anyway. . . good luck! 

P.S. luckily the parsing is pretty well tested. Please see below

### DIFF
--- a/lib/live_cluster/manage_controller.py
+++ b/lib/live_cluster/manage_controller.py
@@ -511,17 +511,13 @@ class ManageACLRolesLeafCommandController(ManageLeafCommandController):
 
 @CommandHelp(
     "Create a role",
-    usage="role <role-name> priv <privilege> [ns <namespace> [set <set>]] [allow <addr1> [<addr2> [...]]] [read <read-quota>] [write <write-quota>]",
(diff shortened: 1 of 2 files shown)


jdogmcsteezy:
Closing. I don't think this should be added. There is no way of determining the different between a read/write priv and a read/write quota followed by a mistypes int. The server would send back an error that would likely confuse the user.

PR #188, opened 2023-08-22
fix: Execute mode error does not have non-zero return code

jdogmcsteezy:
The problem was because of the import order in `logger.py`.  I found another potential future problem where certain ASInfoErrors and ASProtocolErrors should result in the Node being marked dead so I added a new ASProtocolError type and used multiple `except` statements.


dwelch-spike (approved):
LGTM

PR #192, opened 2023-08-30
Tools 2607 collect remote logs

jdogmcsteezy:


lib/live_cluster/logfile_downloader.py
```diff
@@ -86,13 +84,8 @@ def __init__(
...
         Keyword Arguments:
-            enable_ssh {bool} -- Use ssh to gather the logs from Aerospike nodes. This
-            assumes either the required config is defined in the ssh/ssh*_confg files or
-            all nodes require the same credentials and are defined by ssh_config.
-            (default: {False})
-            ssh_config {SSHConnectionConfig | None} -- The ssh config to use the
-            authenticate on the remote node. This overrides the config found in the
-            ssh_config if found. (default: {None})
+            ssh_factory {SSHConnectionFactory | None} -- An optional factory for
+            creating #TODO: finish this comment
```
dwelch-spike, line 88: Can/should this be finished now?

dwelch-spike (commented):
non blocker, incomplete doc string

dwelch-spike (approved):
LGTM, I left a nit in a previous review. Have a look but I don't think it is blocking.

PR #201, opened 2023-09-27
fix: add step in confgen to handle network.tls[0] configs

jdogmcsteezy:


lib/utils/conf_gen.py
```diff
@@ -457,10 +418,52 @@ async def __call__(self, context_dict: dict[str, Any]):
...
+               "name": "tls-name"
+            }
+        }
+    "network":
+        {
+            NamedSection("tls", "tls-name"): {
+               . . .
+            }
+        }
+    should be
```
dwelch-spike, line 435:
```suggestion
```
nit: is this needed?

lib/utils/conf_gen.py
```diff
@@ -457,10 +418,52 @@ async def __call__(self, context_dict: dict[str, Any]):
             self._helper(config_dict)
 
 
-class ConvertIndexedToList(ConfigPipelineStep):
-    def __init__(self):
-        super().__init__()
+class ConvertIndexesSubcontextsToNamedSection(ConfigPipelineStep):
```
dwelch-spike, line 421:
```suggestion
class ConvertIndexedSubcontextsToNamedSection(ConfigPipelineStep):
```
nit: typo?

lib/utils/conf_gen.py
```diff
@@ -457,10 +418,52 @@ async def __call__(self, context_dict: dict[str, Any]):
...
+
+                if isinstance(config, InterUnnamedSectionKey):
+                    config_split = config.type.split("[")
 
+                    if len(config_split) > 1:
+                        if "name" in config_dict[config]:
+                            name = config_dict[config]["name"]
+                            del config_dict[config]["name"]
+
+                            del config_dict[config]
```
dwelch-spike, line 453:
```suggestion
                            del config_dict[config]
```
I think only 1 delete is needed

dwelch-spike (commented):
A few nitpicks. Also, is the pattern to unit test each class in the pipeline? If so this should have tests.

lib/utils/conf_gen.py
jdogmcsteezy: It is needed. Ill rewrite it so it makes more sense though

dwelch-spike (commented):
LGTM

PR #210, opened 2023-10-05
Tools 2679 show stop writes bug

jdogmcsteezy:
Renamed `device_available_pct` to `device_avail_pct` and renamed the threshold metrics to include `storage-engine.` The 7.0 metrics are included as well, but that is ok. There is a slight refactor in some other places around grabbing metrics from dicts.

### DIFF
--- a/lib/utils/common.py
+++ b/lib/utils/common.py
@@ -1672,26 +1672,58 @@ def _is_stop_writes_cause(
 
 
 @staticmethod
+def _get_first_value_from_dict_with_key(
+    dict_: dict[str, Any],
+    key: str | tuple,
+    default_value: Any = None,
+    return_type: type = str,
+) -> Any:
+    if isinstance(key, str):
+        key = (key,)
+
+    for key in key:
+        if key in dict_:
+            return key, util.get_value_from_dict(
+                dict_, key, default_value=default_value, return_type=return_type
+            )
+
+    return None, None
+
+
+@staticmethod
 def _format_ns_stop_writes_metrics(
     stop_writes_metrics: StopWritesDict,
     service_stats,
     ns_stats,
 ):
     for node in service_stats:
-        cluster_clock_skew_ms = service_stats[node].get("cluster_clock_skew_ms", None)
-        cluster_clock_skew_stop_writes_sec = service_stats[node].get(
-            "cluster_clock_skew_stop_writes_sec", None
+        cluster_clock_skew_ms: int | None = util.get_value_from_dict(
+            service_stats[node],
+            "cluster_clock_skew_ms",
+            None,
+            return_type=int,
         )
-        system_free_mem_pct = service_stats[node].get("system_free_mem_pct", None)
+        cluster_clock_skew_stop_writes_sec: int | None = util.get_value_from_dict(
+            service_stats[node],
+            "cluster_clock_skew_stop_writes_sec",
+            None,
+            return_type=int,
+        )
+        system_free_mem_pct: int | None = util.get_value_from_dict(
+            service_stats[node],
+            "system_free_mem_pct",
+            None,
+            return_type=int,
+        )
 
         for ns, stats in ns_stats.get(node, {}).items():
             # There is no config for this trigger
             strong_consistency: str | None = stats.get("strong-consistency", None)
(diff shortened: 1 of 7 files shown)


lib/utils/common.py
```diff
@@ -1649,39 +1649,81 @@ def _create_stop_writes_entry(
...
+            else False
+        )
     return (
         True
         if usage >= threshold and (stop_writes is None or stop_writes.lower() == "true")
         else False
     )
 
 
+@staticmethod
```
dwelch-spike, line 1674: Does the staticmethod decorator do anything if this function isn't the member of a class? I don't see the class this is attached to.

lib/utils/common.py
```diff
@@ -1649,39 +1649,81 @@ def _create_stop_writes_entry(
...
+def _get_first_value_from_dict_with_key(
+    dict_: dict[str, Any],
+    key: str | tuple,
+    default_value: Any = None,
+    return_type: type = str,
+) -> Any:
```
dwelch-spike, line 1680: Should this return type be `(Any, Any)`? Looks like this returns a tuple

lib/utils/common.py
```diff
@@ -1740,70 +1780,87 @@ def _format_ns_stop_writes_metrics(
...
+                return_type=int,
+            )
+            bytes_total: int | float | None = util.get_value_from_dict(
+                stats,
+                ("data_total_bytes", "device_total_bytes", "pmem_total_bytes"),
+                None,
+                return_type=int,
+            )
 
             if usage is not None and threshold is not None and bytes_total is not None:
```
dwelch-spike, line 1830: Does this also need a check for `metric`? like on line 1799?

lib/utils/common.py
```diff
@@ -1740,70 +1780,87 @@ def _format_ns_stop_writes_metrics(
...
+            )
+            threshold = util.get_value_from_dict(
+                stats, config, default_value=None, return_type=int
+            )
 
             if usage is not None and threshold is not None and bytes_total is not None:
-                use = int(usage)
-                thresh = int(bytes_total) * (int(threshold) / 100)
-                sw = _is_stop_writes_cause(use, thresh, stop_writes)
+                threshold = int(bytes_total) * (int(threshold) / 100)
```
dwelch-spike, line 1856: Are the `int()` casts still needed? They are removed on line 1831

dwelch-spike (changes requested):
Mostly double checking some things, please have a look.

lib/utils/common.py
```diff
@@ -1649,39 +1649,81 @@ def _create_stop_writes_entry(
...
+    invert: bool = False,
 ):
     if threshold == 0:
         return False
 
+    if invert:
+        return (
+            True
+            if usage <= threshold
+            and (stop_writes is None or stop_writes.lower() == "true")
```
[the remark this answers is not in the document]
jdogmcsteezy: It would be if `data_avail_pct` = `50` and `storage-engine.stop-writes-avail-pct` = 55. This is the only case where `invert=True`

lib/utils/common.py
```diff
@@ -1740,70 +1780,87 @@ def _format_ns_stop_writes_metrics(
...
+                ),
+                default_value=None,
+                return_type=int,
+            )
 
-            if usage is not None and threshold is not None:
-                use = int(usage)
-                thresh = int(threshold)
-                sw = _is_stop_writes_cause(use, thresh, stop_writes)
+            if metric and usage is not None and threshold is not None:
```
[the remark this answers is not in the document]
jdogmcsteezy: metric should be string with length > 0

lib/utils/common.py
jdogmcsteezy: Should not matter. The other check for metric is also a bit redundant

jdogmcsteezy: Ill remove the check from the above comment

lib/utils/common.py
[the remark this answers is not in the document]
jdogmcsteezy: removing check

lib/utils/common.py
jdogmcsteezy: nope, good catch

dwelch-spike (approved):
LGTM

PR #220, opened 2023-10-17
build: TOOLS-2703 consolidate pyinstaller build specs

jdogmcsteezy:
pyinstaller-build.spec is a consolidation of the 4 deleted spec files


dwelch-spike (approved):
LGTM

PR #231, opened 2023-11-06
fix:  TOOLS-2717 `info namespace object` displays error when connecting to CE server

jdogmcsteezy:


dwelch-spike (approved):
LGTM

lib/live_cluster/client/node.py
```diff
@@ -91,11 +91,11 @@ async def wrapper(*args, raise_exception=False, **kwargs):
         except (ASInfoNotAuthenticatedError, ASProtocolConnectionError) as e:
             args[0].alive = False
             exception = e
-        except (ASInfoError, ASProtocolError) as e:
-            exception = e
-        except Exception as e:
+        except OSError as e:
             args[0].alive = False
             exception = e
+        except (ASInfoError, ASProtocolError, Exception) as e:
```
kportertx, line 97: `Exception` should cove both `ASInfoError` and `ASProtocolError`. Are you sure you want to handle all exceptions this way?

PR #244, opened 2023-12-08
TOOLS-2783 TOOLS-2782 `generate config` bugs

jdogmcsteezy:


lib/utils/conf_gen.py
```diff
@@ -530,7 +551,10 @@ def _helper(
 
             # Handles ldap scenarios with configs that allow commas in the config and
             # are returned with commas. e.g. "query-base-dn" and "user-dn-pattern"
-            if len(split_value) > 1 and not ("-dn" in config and "ldap" in context):
+            if len(split_value) > 1 and not (
+                ("-dn" in config and "ldap" in context)
+                or config in {"dcs", "namespaces"}
```
dwelch-spike, line 556: What is the `or config in {"dcs", "namespaces"}` handling?

lib/utils/conf_gen.py
```diff
@@ -959,6 +984,28 @@ async def _generate_intermediate(
...
+            elif isinstance(key, InterListKey):
+                list_keys.append(key)
+            elif isinstance(key, InterUnnamedSectionKey):
+                unnamed_keys.append(key)
+            elif isinstance(key, InterNamedSectionKey):
+                named_keys.append(key)
+
+        str_keys.sort()
+        list_keys.sort(key=lambda x: x.name)
+        named_keys.sort(key=lambda x: x.type + x.name)
```
dwelch-spike, line 1005: Why no sort for unnamed_keys?

lib/utils/conf_gen.py
```diff
@@ -996,4 +1044,20 @@ async def generate(
         lines = []
 
         self._generate_helper(lines, list(intermediate_dict.values())[0])
-        return "\n".join(lines)
+        return "\n".join(lines).replace("{\n\n", "{\n")
```
dwelch-spike, line 1047: Why does this happen? Is it possible to fix it when `lines` is generated?

lib/utils/conf_gen.py
```diff
@@ -996,4 +1044,20 @@ async def generate(
...
+    "ignored-bins": "ignore-bin",
+    "ignored-sets": "ignore-set",
+}
+
+
+def convert_response_to_config_key(key: str) -> str:
+    if key in server_to_config_key_map:
+        return server_to_config_key_map[key]
+
+    return key
```
dwelch-spike, line 1063: nit:
```suggestion
    return server_to_config_key_map.get(key, key)
```

dwelch-spike (approved):
A few suggestions and understanding questions but nothing blocking.

lib/utils/conf_gen.py
jdogmcsteezy: The info responses return `dcs=DC1,DC2` and I want to make sure we don't split those. Once they are split, and turned into a list it is more difficult to remove them later in the step that deletes invalid keys since `dcs` is an invalid key

lib/utils/conf_gen.py
jdogmcsteezy: I suppose I could. I just did not see it as necessary like the others. The reason I did this is because when requesting dc and ns configs they are called concurrently and the results could be in any order. Dicts in python iterate by insertion order. The responses that provide the unamed sections are always in the same order and come from a single info request. 

This was mostly done to make comparisons easier in testing

lib/utils/conf_gen.py
jdogmcsteezy: fixed

PR #262, opened 2024-01-26
Tools 2831 Deprecate `generate conf` cmd

jdogmcsteezy:
Please look over the wording. I suspect since we only recently added this feature and have marked it has BETA we can remove it without issuing a breaking change release.

Ignore the ssh change. That is already in master


dwelch-spike (approved):
LGTM

PR #290, opened 2024-05-28
build: TOOLS-2932 use github actions to build m-series macos

jdogmcsteezy:


.github/workflows/mac-artifact.yml
```diff
@@ -16,7 +16,10 @@ on:
 
 jobs:
   build:
-    runs-on: macos-11
+    strategy:
+      matrix:
+        os: [macos-11, macos-14]
```
dwelch-spike, line 21:
```suggestion
        os: [macos-12, macos-14]  
```
Macos-11 runners are no longer supported.

dwelch-spike:
@jdogmcsteezy the conflicts need to be fixed and macos-11 runners should be bumped to macos-12.

dwelch-spike (approved):

PR #306, opened 2024-09-06
fix: TOOLS-2964 Error Response from 7.2 for Show Latencies

a-spiker:
Error Syntax has changed to ERROR:

~Checking what can be added as test for this.~

### DIFF
--- a/lib/live_cluster/client/node.py
+++ b/lib/live_cluster/client/node.py
@@ -2087,7 +2087,8 @@ class Node(AsyncObject):
                 hist_info.append(await self._info(cmd))
             except Exception:
                 return data
-            if hist_info[-1].startswith("error"):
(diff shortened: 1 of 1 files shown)


jdogmcsteezy (approved):
Seems like a good initial task! LGTM. Only nit would be 

Fix: (TOOLS-2964) Error Response from 7.2 for Show Latencies

should be 

fix: TOOLS-2964 Error Response from 7.2 for Show Latencies

of course when you officially take it over you can do whatever you want but that is the current formatting rule.

After this is merged, make sure to update the aerospike-tools package to point to the latest master of aerospike-admin. Normally the commit message for that is "update admin TOOLS-2963. This helps someone looking at the ticket to see that a change has been places at the individual tool repo and then at the meta-package level.

PR #311, opened 2024-10-01
update download-artifact action to v4, fixes compatibility with the s…

dwelch-spike:
…nyk-test action which uses artifact-upload v4


a-spiker (approved):
LGTM

PR #315, opened 2024-10-22
fix: TOOLS-2978 & TOOLS-2977 federal edition in build col

a-spiker:
Build column should have F- for Federal edition

Fixes: TOOLS-2977 & TOOLS-2977

Simplifying the logic, Info Edition Calls to Server will not have true / false or CE / EE. Valid responses are 
- Aerospike Federal Edition build 6.4.0.15
- Aerospike Enterprise Edition
- Aeropsike Community Edition

### DIFF
--- a/lib/utils/util.py
+++ b/lib/utils/util.py
@@ -687,18 +687,17 @@ def find_delimiter_in(value):
     return ";"
 
 
-def convert_edition_to_shortform(edition):
-    """Convert edition to shortform Enterprise or Community or Federal or N/E"""
-
-    if "enterprise" in edition.lower():
-        return "Enterprise"
-
-    if "community" in edition.lower():
-        return "Community"
+def convert_edition_to_shortform(edition: str) -> str:
+    """Convert edition to shortform: Enterprise, Community, Federal, or N/E"""
+    edition_lower = edition.lower()
     
-    if "federal" in edition.lower():
+    if "enterprise" in edition_lower:
+        return "Enterprise"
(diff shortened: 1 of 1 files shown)


dwelch-spike (approved):
LGTM

a-spiker:
Verified the behaviour with 
- aerospike:4.4.0.6
- aerospike:4.9.0.8
- aerospike:ee-5.6.0.5
- aerospike:ee-6.2.0.2
- aerospike:ce-6.2.0.2
- aerospike:ce-7.1.0.3

sunilvirus:
Ok. That’s nice.
Let’s simplify the code like we discussed yesterday.

a-spiker:
@sunilvirus Already done based on our discussion.

lib/utils/util.py
```diff
@@ -688,19 +688,16 @@ def find_delimiter_in(value):
...
 
 def convert_edition_to_shortform(edition):
-    """Convert edition to shortform Enterprise or Community or N/E"""
+    """Convert edition to shortform Enterprise or Community or Federal or N/E"""
 
-    if (
-        edition.lower() in ["enterprise", "true", "ee"]
-        or "enterprise" in edition.lower()
-    ):
+    if "enterprise" in edition.lower():
```
sunilvirus, line 693: Shall we simply this logic further. Worst case scenario, we are doing the edition.lower() 3 times. We should just do it once and do a switch statement instead of 3 if conditions.

PR #332, opened 2025-01-20
fix: TOOLS-2986 string match should support older 7.2 version

a-spiker:
With this PR https://github.com/aerospike/aerospike-admin/pull/329, we updated the server error response strings `unstable-cluster` to `unstable cluster`, which dropped the support for the server running before 7.2 https://aerospike.atlassian.net/browse/AER-6764

This PR enables back support for older error string too so that asadm remains compatible with 7.1 version and before.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,17 +12,14 @@ jobs:
         with:
           submodules: true
       - name: Get Python version from Pipfile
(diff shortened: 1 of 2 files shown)


a-spiker:
Closing this in preference to https://github.com/aerospike/aerospike-admin/pull/331

PR #337, opened 2025-02-20
fix: TOOLS-3012 standardise error handling string for manage roster

a-spiker:
- removes special string handling

### DIFF
--- a/test/unit/live_cluster/client/test_node.py
+++ b/test/unit/live_cluster/client/test_node.py
@@ -2053,34 +2053,6 @@ class NodeTest(asynctest.TestCase):
         )
 
     async def test_info_cluster_stable_with_errors(self):
-        self.info_mock.return_value = "ERROR::cluster not specified size"
-        expected = ASInfoResponseError(ErrorsMsgs.INFO_SERVER_ERROR_RESPONSE, "ERROR::cluster not specified size")
-
-        actual = await self.node.info_cluster_stable(cluster_size=3, namespace="bar")
-
-        self.info_mock.assert_called_with(
-            "cluster-stable:size=3;namespace=bar", self.ip
-        )
-        self.assertEqual(
-            actual,
-            expected,
-            "info_cluster_stable did not return the expected result",
-        )
-
-        self.info_mock.return_value = "ERROR::unstable cluster"
-        expected = ASInfoResponseError(ErrorsMsgs.INFO_SERVER_ERROR_RESPONSE, "ERROR::unstable cluster")
-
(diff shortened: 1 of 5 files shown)


test/unit/live_cluster/client/test_node.py
```diff
@@ -2067,7 +2067,7 @@ async def test_info_cluster_stable_with_errors(self):
         )
 
         self.info_mock.return_value = "ERROR::unstable cluster"
-        expected = ASInfoClusterStableError("ERROR::unstable cluster")
+        expected = ASInfoResponseError("Server returned an error response for info command", "ERROR::unstable cluster")
```
kportertx, line 2070: This looks like it would still fail since it would be `ERROR:7:unstable cluster`

a-spiker: @kportertx We have removed special string handling with this PR and treat every error generically. With this change, we always handle the Excpetion which goes to a happy path where the User can choose what to do next. 

We were earlier abstracting the Error thrown from the server. Here s a complete experience example.

```
Admin+> manage roster stage observed ns sc
ERROR: Server returned an error response for info command : unstable cluster.
WARNING: It is advised that you do not manage the roster. Run 'info network' for more information.
You are about to set the pending-roster for namespace sc to: A2, A1, A0
Confirm that you want to proceed by typing x5d8ee, or cancel by typing anything else.
x5d8ee
Pending roster now contains observed nodes.
Run "manage recluster" for your changes to take effect.
```

a-spiker: This path was broken due to unstable-cluster and unstable cluster change from the server where asadm was handling these special strings viz `unstable cluster`, `cluster not specified size`

kportertx: K, just looked like this test would still fail.

test/unit/live_cluster/client/test_node.py
```diff
@@ -2053,7 +2053,7 @@ async def test_info_cluster_stable(self):
 
     async def test_info_cluster_stable_with_errors(self):
         self.info_mock.return_value = "ERROR::cluster not specified size"
-        expected = ASInfoClusterStableError("ERROR::cluster not specified size")
+        expected = ASInfoResponseError("Server returned an error response for info command", "ERROR::cluster not specified size")
```
sud82, line 2056: Can we directly use INFO_SERVER_ERROR_RESPONSE here rather than the hardcoded string?

a-spiker: Done

sud82 (approved):

PR #346, opened 2025-05-02
feat: TOOLS-3044 TOOLS-3084 Remove reference and logic to compute summary from UDA data

a-spiker:
Changes:
- Removes reference / options to add agent
- Remove logic to compute license_data
- Removes testcases around license data


dwelch-spike (approved):
Code LGTM. This is a breaking change so asadm and the tools package should increment their major versions when this is released.

PR #349, opened 2025-05-28
feat: TOOLS-3054 add auth mode field to user info and templates

a-spiker:
With this PR, it enables a new column for show users called Auth Mode.

Screenshot 2025-06-05 at 4 54 30 PM

The same will be available in CollectInfo Mode

Screenshot 2025-06-05 at 4 57 58 PM

Update. the column is moved to the end.


a-spiker:
As discussed internally, moved the column at the end, and removed from user statistics command [which is more based on dynamic properties]

PR #351, opened 2025-06-02
feat: TOOLS-3055 - use the admin port instead of service port based on flag

a-spiker:
# Add --use-seed-address flag to force connection via seed address only

## Overview
This PR introduces a new `--use-seed-address` CLI flag that forces the Aerospike admin tool to connect exclusively through the provided seed address, bypassing service address discovery. This feature is particularly useful in network environments with load balancers, NAT, or complex routing where service address discovery may return unreachable addresses.

### 1. CLI Argument Addition and Validation
- **Added `--use-seed-address` CLI flag** with proper help text
- **Added validation** to ensure `--use-seed-address` can only be used with `--single-node`
- **Added configuration support** in conf.py

**Files Modified:**
- conf.py - Added CLI argument definition and configuration support
- asadm.py - Added validation logic and parameter passing to shell

### 2. Core Connection Logic Enhancement
- **Modified `Node._node_connect()` method** to handle seed-only connections
- **When `use_seed_address=True`**: Only retrieves node ID and features, skips service/peer discovery
- **When `use_seed_address=False`**: Performs full service and peer discovery (existing behavior)
- **Streamlined logic** to avoid unnecessary network calls in seed-only mode

**Files Modified:**
- node.py - Core connection logic with conditional behavior
- cluster.py - Parameter passing to Node constructor  
- live_cluster_root_controller.py - Controller parameter handling

### 3. Comprehensive Testing
- **Added unit tests** for both seed-only and full discovery modes
- **Tests verify** correct command execution and return values
- **Ensures** proper mocking and validation of connection behavior

**Files Modified:**
- test_node.py - Added `test_node_connect_seed_mode` and `test_node_connect_full_mode`

### Normal Operation (default: `use_seed_address=False`)
1. Connect to seed address
2. Discover service addresses and peers from server  
3. Use discovered service addresses for subsequent connections
4. Build full cluster topology

### With --use-seed-address (requires --single-node)
1. Connect to seed address
2. **Skip service discovery** - only get node ID and features
3. **Skip peer discovery** - return empty peers list
4. **Force all connections** to use only the original seed address
5. **Maintain single-node connection mode**

### Optimized Connection Logic
The `_node_connect()` method now has conditional logic:

```python
if self.use_seed_address:
    # Minimal commands - no service/peer discovery
    commands = ["node", "features"]
    # Return seed address as the only service address
    return node_id, [(self.ip, self.port, self.tls_name)], features, []
else:
    # Full discovery (existing behavior)
    commands = ["node", service_info_call, "features"] + peers_info_calls
    # Return discovered service addresses and peers
```

### Validation Logic
```python
if cli_args.use_seed_address and not cli_args.single_node:
    logger.critical("--use-seed-address can only be used with --single-node. Please enable --single-node or remove --use-seed-address.")
```

## Use Cases
- **Load Balancers**: Connect through load balancer without switching to backend servers
- **NAT/Firewall**: When only the seed address is accessible
- **Container Environments**: When service discovery returns unreachable internal IPs
- **Network Isolation**: Scenarios requiring explicit address control

## Example Usage
```bash
# Valid usage
asadm --host 192.168.1.100 --single-node --use-seed-address

# Invalid usage (exits with error)
asadm --host 192.168.1.100 --use-seed-address
```

## Backward Compatibility
- **Fully backward compatible** - new flag is opt-in only
- **Default behavior unchanged** when flag is not specified
- **No impact** on existing deployments or scripts

## Testing
- **Unit tests added** for both connection modes
- **Validation logic tested** for CLI argument combinations
- **Maintains existing test coverage** for default behavior

### DIFF
--- a/asadm.py
+++ b/asadm.py
@@ -106,7 +106,7 @@ class AerospikeShell(cmd.Cmd, AsyncObject):
         execute_only_mode=False,
         privileged_mode=False,
         timeout=1,
-        use_seed_node=False,
+        use_seed_address=False,
     ):
         # indicates shell created successfully and connected to cluster/collectinfo/logfile
         self.connected = True
@@ -177,7 +177,7 @@ class AerospikeShell(cmd.Cmd, AsyncObject):
                     only_connect_seed,
                     timeout=timeout,
                     asadm_version=admin_version,
-                    use_seed_node=use_seed_node,
+                    use_seed_address=use_seed_address,
                 )
 
                 if not self.ctrl.cluster.get_live_nodes():
@@ -254,8 +254,6 @@ class AerospikeShell(cmd.Cmd, AsyncObject):
         for command in commands:
             if command != "help":
                 self.commands.add(command)
-
-        self.use_seed_node = use_seed_node
 
     async def active_stop_writes(self, cluster: Cluster):
         """
@@ -762,6 +760,11 @@ async def main():
     ):
         logger.critical("TLS is required for authentication mode: " + cli_args.auth)
 
+    if cli_args.use_seed_address and not cli_args.single_node:
+        logger.critical(
+            "--use-seed-address can only be used with --single-node. Please enable --single-node or remove --use-seed-address."
+        )
+
     ssl_context = parse_tls_input(cli_args)
 
     if cli_args.asinfo_mode:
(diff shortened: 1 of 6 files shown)


dwelch-spike (approved):
I left a nit, and I think the title of the PR and the ticket does not match the contents, but the code LGTM

kportertx (changes requested):
https://aerospike.atlassian.net/browse/SERVER-263 should render this PR unnecessary - please hold.

PR #363, opened 2025-07-08
fix: TOOLS-3064 Fix license data calculation for mixed-version clusters and add safety clamping

a-spiker:
## Problem

The license data calculation in `_manually_compute_license_data_size()` had one critical issues:

**Incorrect global-level overhead calculation**: The function calculated record overhead at the namespace level, but this fails in mixed-version clusters where different hosts may run different Aerospike versions with different overhead requirements.

## Defensive Programming
**Missing safety bounds**: License calculations could result in negative values when record overhead exceeded actual data size, particularly with small records, high compression ratios, or test environments.

### 1. **Per-Host Overhead Calculation**
Changed from namespace-level to per-host calculation logic:

**Before (Incorrect):**
```python
# Applied single overhead logic across entire namespace
ns_unique_data = (total_data / repl_factor) - (total_objects * overhead)
```

**After (Correct):**
```python
# Each host contributes based on its individual version
for each host:
    if host_version >= 8.0:
        host_contribution = host_data / repl_factor  # No overhead
    else:
        host_contribution = (host_data / repl_factor) - (host_objects * overhead)
    
    ns_total += host_contribution
```

This correctly handles mixed-version clusters where some nodes use overhead subtraction (< 8.0) while others don't (>= 8.0).

### 2. **Safety Clamping**
Added bounds checking to prevent negative license values:

```python
# Ensure license data is never negative
ns_unique_data_clamped = max(0, ns_unique_data)
```

## Key Improvements

- ✅ **Mixed-version support**: Correctly handles clusters with different Aerospike versions per host
- ✅ **Version-specific logic**: Applies appropriate overhead calculation (35B, 39B, or none) per host
- ✅ **Safety bounds**: Prevents negative license values that could occur with small records
- ✅ **Accurate reporting**: License usage now reflects actual per-host storage calculations

## Example

**Mixed-version namespace calculation:**
```
Namespace "test":
  - Host 1.1.1.1 (v7.0.0): (40000/2) - (200×35) = 13000
  - Host 2.2.2.2 (v8.0.0): (40000/2) = 20000  
  - Total: 33000

Namespace "bar" (overhead exceeds data):
  - Calculated: -5800 → Clamped to: 0
```

This ensures robust and accurate license calculation across all deployment scenarios while preventing invalid negative values.

### DIFF
--- a/lib/utils/common.py
+++ b/lib/utils/common.py
@@ -991,7 +991,6 @@ def _manually_compute_license_data_size(
             continue
 
         ns_unique_data = 0.0
-        ns_master_objects = 0
         ns_repl_factor = 1
 
         for host_id, host_stats in ns_stats.items():
@@ -1001,16 +1000,6 @@ def _manually_compute_license_data_size(
 
             if not host_stats or isinstance(host_stats, Exception):
                 continue
-            
-            host_build_version = util.get_value_from_dict(
-                server_builds,
-                host_id,
-                default_value=None,
-                return_type=str,
-            )
-
-            if host_build_version is None:
-                raise Exception("could not find host %s in build responses" % host_id)
 
             repl_factor = util.get_value_from_dict(
                 host_stats,
@@ -1075,6 +1064,16 @@ def _manually_compute_license_data_size(
                     return_type=float,
                 )
 
+            host_build_version = util.get_value_from_dict(
+                server_builds,
+                host_id,
+                default_value=None,
+                return_type=str,
+            )
+
+            if host_build_version is None:
+                raise Exception("could not find host %s in build responses" % host_id)
+            
             host_record_overhead = 35
 
             if version.LooseVersion(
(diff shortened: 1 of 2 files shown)


lib/utils/common.py
```diff
@@ -1084,21 +1082,23 @@ def _manually_compute_license_data_size(
...
+                # For 8.0 and above - don't subtract overhead for this host
+                host_license_contribution = host_unique_data / ns_repl_factor
+            else:
+                # For versions below 8.0 - subtract overhead for this host
+                host_license_contribution = (host_unique_data / ns_repl_factor) - (host_master_objects * host_record_overhead)
+            
+            ns_unique_data += host_license_contribution
         
+        # Clamp namespace license data to ensure it's never negative
+        ns_unique_data_clamped = max(0, ns_unique_data)
```
dwelch-spike, line 1097: If the ns_unique_data is negative, issuing a warning or debug log would be good.

dwelch-spike (commented):
The code LGTM but are we sure that Product only wants the new unique data calculation to apply to server 8.0 and newer? Did you check with Ronen to be sure?

dwelch-spike (approved):
LGTM!

PR #364, opened 2025-07-09
fix: Add 'Auth Mode' to expected user table headers in tests

a-spiker:
Updated the expected headers in TestShowUsers to include 'Auth Mode' for improved test coverage and alignment with current output.


dwelch-spike (approved):
LGTM

PR #365, opened 2025-07-09
fix: TOOLS-3063 fix Index Usage Calculation with Fallback Logic for Aerospike Version Compatibility

a-spiker:
## Summary

This PR updates the `info namespace usage` view to support metrics introduced in Aerospike server versions 7.0 and later, while maintaining backward compatibility with older versions. The primary and secondary index usage calculations have been enhanced to handle different storage configurations (e.g., flash/pmem with mounts vs. in-memory) and prioritize newer metrics when available.

**Key Changes:**

*   **`lib/view/templates.py`**:
    *   Modified `info_namespace_usage_sheet` to use `Projectors.Any` for calculating `Used%` for both "Primary Index" and "Secondary Index".
    *   The logic now correctly prioritizes `index-type.mounts-budget` (for flash/pmem) and `indexes-memory-budget` (for in-memory indexes, added in 7.1) over older metrics.
    *   It now uses `index_mounts_used_pct` and `sindex_mounts_used_pct` when available for a more accurate usage percentage, converting it from a 0-100 value to a ratio.
    *   For older servers, it falls back to calculating usage by dividing the used bytes by the size limit.

*   **`test/unit/view/test_view.py`**:
    *   Added a comprehensive suite of unit tests for `info_namespace_usage` to validate the new logic.
    *   Test cases cover various scenarios, including:
        *   Server 7.0 with `mounts-budget` metrics.
        *   Server 7.1 with memory-only `indexes-memory-budget`.
        *   Pre-7.0 servers with legacy metrics.
        *   Mixed environments to ensure correct metric prioritization.
        *   Cases with partial or missing metrics.

These changes ensure that `asadm` provides an accurate and reliable view of index usage across different Aerospike server versions and configurations.

### DIFF
--- a/lib/view/templates.py
+++ b/lib/view/templates.py
@@ -296,9 +296,9 @@ info_namespace_usage_sheet = Sheet(
                     Projectors.Number(
                         "ns_stats",
                         "index_used_bytes",  # flash, pmem, and memory metrics were consolidated in 7.0
-                        "index_flash_used_bytes",
-                        "index_pmem_used_bytes",
-                        "memory_used_index_bytes",
+                        "index_flash_used_bytes", # pre 7.0
+                        "index_pmem_used_bytes", # pre 7.0
+                        "memory_used_index_bytes", # pre 7.0
                     ),
                     converter=Converters.byte,
                     aggregator=Aggregators.sum(),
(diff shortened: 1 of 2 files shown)


dwelch-spike (changes requested):
This looks good! If possible, can you add tests? This is complex enough that I think it is worth having tests for the different configurations we may see, such as, on flash, in mem, and older and newer servers.

sunilvirus (commented):
I need to understand this better. lets discuss offline for me to get some idea.

a-spiker:
@dwelch-spike I had a review session with Sunil, the logic is now further optimised.  I have added tc which will assert on the table values for various versions where metrics may be missing.

PR #367, opened 2025-07-10
refactor: TOOLS-3060 add debug log for decoded user agent

a-spiker:
A debug log statement was added to output the decoded user agent for each node, aiding in troubleshooting and monitoring user agent decoding.


dwelch-spike (approved):
Nice addition 👍  LGTM

PR #369, opened 2025-07-14
fix: Update E2E Tests: Pin Docker Image to 8.1 and Add Environment Variable Support for Mac

a-spiker:
### Summary

1. **Docker Image Version Pinning**
   - Updated default Docker tag from `latest` to `"8.1"`
   - I will later use the internal image, which points to build [requires docker auth setup] nightly

2. **Environment Variable Support**
   - Added `E2E_TEST_USE_LOCALHOST` environment variable for macOS compatibility
   - When set to `true`, tests use `127.0.0.1` instead of container IP addresses
   - This resolves Docker networking issues on macOS where bridge network IPs are not accessible

### Testing

The changes maintain backward compatibility while improving test reliability:
- Default behavior remains unchanged for Linux environments
- macOS users can now run tests successfully by setting `E2E_TEST_USE_LOCALHOST=true`
- Docker image version is now predictable and stable

### Usage

For macOS users experiencing Docker networking issues:
```bash
E2E_TEST_USE_LOCALHOST=true FEATKEY=$(base64 -i path/to/features.conf) make integration
```

### Files Modified

- lib.py: Updated Docker tag default and added localhost environment variable support
- README.md: Added documentation for the new environment variable

This PR improves test stability and cross-platform compatibility without breaking existing functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,11 @@ to test the bundled app run:
 ASADM_TEST_BUNDLE=true FEATKEY=(base64 -i path/to/features.conf) make integration
 ```
 
+**Note for Mac users**: If you encounter Docker networking issues on macOS, enable localhost networking:
+```
+ASADM_TEST_USE_LOCALHOST=true FEATKEY=(base64 -i path/to/features.conf) make integration
+```
+
 ### Running all tests with coverage
 ```
 FEATKEY=(base64 -i path/to/features.conf) make coverage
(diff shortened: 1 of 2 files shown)


dwelch-spike (approved):
We'll need to make sure to update the docker image used in testing when 8.1 is no longer needed, LGTM

PR #372, opened 2025-07-23
fix: TOOLS-3097 Improve CLI Error Handling: Reject Unknown Subcommands for info Command

a-spiker:
This PR improves the user experience for the `info` command in both live-cluster and collectinfo modes by explicitly rejecting unknown subcommands. Previously, entering an invalid subcommand (e.g., `info random`) would fall back to the default `info`, which could confuse users. Now, the CLI immediately returns a clear error message indicating the subcommand is invalid and suggests using help info for available options.

Additionally, an end-to-end test has been added to verify that unknown subcommands to info produce the correct error and do not show the default summary. This change ensures more predictable and user-friendly CLI behavior, aligning with standard practices in modern command-line tools.

### DIFF
--- a/lib/live_cluster/info_controller.py
+++ b/lib/live_cluster/info_controller.py
@@ -20,6 +20,7 @@ from lib.live_cluster.get_controller import (
 from lib.utils import util, version, constants
 from lib.base_controller import CommandHelp, ModifierHelp
 from .live_cluster_command_controller import LiveClusterCommandController
+from lib.base_controller import ShellException
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


lib/live_cluster/info_controller.py
```diff
@@ -54,6 +55,11 @@ async def _do_default(self, line):
         # 'info namespace object', but since it is not correct command it should print output for partial correct
         # command, in this case it should print data for 'info'. To keep consistent output format, we are passing empty
```
sunilvirus, line 56: These comments need update as they are describing the current behaviour. We should explain the new behaviour in the comments.

PR #374, opened 2025-07-28
fix: TOOLS-3065 Refine validation for mutually exclusive modifiers for sindex-create

a-spiker:
Improves the validation logic to separately check for conflicts between 'ctx' and 'exp_base64', and 'ctx_base64' and 'exp_base64', providing more precise error messages for each case.

More on this https://aerospike.atlassian.net/browse/TOOLS-3065?focusedCommentId=165372

### DIFF
--- a/lib/live_cluster/manage_controller.py
+++ b/lib/live_cluster/manage_controller.py
@@ -1341,10 +1341,15 @@ class ManageSIndexCreateController(ManageLeafCommandController):
                 )
 
         # Validate mutually exclusive modifiers - ctx (including ctx_base64) and exp_base64
-        if (ctx_list or cdt_ctx_base64) and exp_base64 is not None:
-            raise ShellException(
-                "Cannot use both 'ctx' and 'exp_base64' modifiers together. Use either 'ctx' to specify how to index into a CDT, or 'exp_base64' to specify an expression to be evaluated."
-            )
+        if exp_base64 is not None:
+            if ctx_list:
+                raise ShellException(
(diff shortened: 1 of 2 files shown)


lib/live_cluster/manage_controller.py
```diff
@@ -1341,10 +1341,15 @@ async def _do_create(self, line, bin_type: str):
                 )
 
         # Validate mutually exclusive modifiers - ctx (including ctx_base64) and exp_base64
-        if (ctx_list or cdt_ctx_base64) and exp_base64 is not None:
-            raise ShellException(
-                "Cannot use both 'ctx' and 'exp_base64' modifiers together. Use either 'ctx' to specify how to index into a CDT, or 'exp_base64' to specify an expression to be evaluated."
-            )
+        if exp_base64 is not None:
+            if ctx_list:
```
srivarun91, line 1345: what is ctx_list? just curious why there is no equivalent for exp_base64.

a-spiker: We have a parsing logic for ctx, and one can pass  raw context which asadm parses.. eg. ctx list_index(-1) map_key(first-name). we don't have a parsing logic for exp yet.

a-spiker: A list of context items describing how to index into a CDT. Possible values include: list_index(<int>) list_rank(<int>) list_value(<value>), map_index(<int>), map_rank(<int>) map_key(<value>), and map_value(<value>). Where <value> i <string>, int(<int>), bool(<bool>), or bytes(<base64>) a base64 encoded byte array (no quotes).

srivarun91 (commented):
LGTM.

PR #383, opened 2025-09-06
SERVER-216

cinterloper:
Built [packages with links to JFrog](https://github.com/aerospike/aerospike-admin/actions/runs/18727326399) are available in the `Actions` tab above

[Integration tests are here](https://github.com/aerospike/aerospike-admin/tree/SERVER-216/.github/docker/test). 

Test cases download the current build from JFrog and install it in a fresh container, then execute it with the --help argument 

You need a JFrog username and token to execute the test cases, an example is in the README.sh

### DIFF
--- a/lib/live_cluster/client/node.py
+++ b/lib/live_cluster/client/node.py
@@ -3389,11 +3389,14 @@ class Node(AsyncObject):
         if set_ is not None:
             masking_show_req += "set={};".format(set_)
 
+        resp = await self._info(masking_show_req)
+
+        if resp.startswith("error") or resp.startswith("ERROR"):
+            raise ASInfoResponseError("Failed to list masking rules", resp)
+
         return [
             client_util.info_to_dict(v, ";")
-            for v in client_util.info_to_list(
-                await self._info(masking_show_req), delimiter=":"
-            )
+            for v in client_util.info_to_list(resp, delimiter=":")
             if v != ""
         ]
 
--- a/lib/live_cluster/collectinfo_controller.py
+++ b/lib/live_cluster/collectinfo_controller.py
@@ -621,6 +621,7 @@ class CollectinfoController(LiveClusterCommandController):
                 dignostic_show_params.append("pmap")
 
             dignostic_aerospike_info_commands = [
+                "connection",
                 "service-clear-std",
                 "service-clear-alt",
                 "service-tls-std",
--- a/test/unit/live_cluster/client/test_node.py
+++ b/test/unit/live_cluster/client/test_node.py
@@ -4597,6 +4597,16 @@ class NodeTest(asynctest.TestCase):
 
         self.assertEqual(result, [])
 
+    async def test_info_masking_list_rules_error(self):
+        """Test masking rules listing with ERROR response"""
+        self.info_mock.return_value = "ERROR::masking not supported"
+
+        result = await self.node.info_masking_list_rules()
+
+        self.assertIsInstance(result, ASInfoResponseError)
+        self.assertEqual(result.message, "Failed to list masking rules")
+        self.assertEqual(result.response, "masking not supported")
+
 
 class SyscmdTest(unittest.TestCase):
     def setUp(self) -> None:


.github/workflows/build-artifacts.yml
```diff
@@ -0,0 +1,76 @@
...
+    secrets:
+      gpg-private-key: ${{ secrets.GPG_SECRET_KEY }}
+      gpg-public-key: ${{ secrets.GPG_PUBLIC_KEY }}
+      gpg-key-pass: ${{ secrets.GPG_PASS }}
+
+  upload-artifacts:
+    strategy:
+      matrix:
+        distro: [ debian12, debian13, ubuntu20.04, ubuntu22.04, ubuntu24.04, el8, el9, el10, amzn2023 ]
+        host: [ ubuntu-24.04-arm, ubuntu-24.04 ]
```
pvinh-spike: Should it be "ubuntu-24.04-arm64"

.github/packaging/project/build_package.sh
```diff
@@ -0,0 +1,28 @@
+
+function build_packages(){
+  if [ "$ENV_DISTRO" = "" ]; then
+    echo "ENV_DISTRO is not set"
+    return
+  fi
+  chown -R root:root .
```
dwelch-spike: why is this chown needed?

.github/packaging/project/install_deps.sh
```diff
@@ -0,0 +1,271 @@
...
+      mkdir -p /opt/golang && tar -zxvf /tmp/go1.24.6.linux-amd64.tar.gz -C /opt/golang
+  elif [ "$(uname -m)" = "aarch64" ]; then
+      curl -L https://go.dev/dl/go1.24.6.linux-arm64.tar.gz -o /tmp/go1.24.6.linux-arm64.tar.gz
+      mkdir -p /opt/golang && tar -zxvf /tmp/go1.24.6.linux-arm64.tar.gz -C /opt/golang
+  else
+      echo "unknown arch $(uname -m)"
+      exit 1
+  fi
+  /opt/golang/go/bin/go install github.com/asdf-vm/asdf/cmd/asdf@v0.18.0
+  install /root/go/bin/asdf /usr/local/bin/asdf
```
dwelch-spike: Where is the use of "asdf" coming from?

dwelch-spike: I see pyenv (or nowadays uv) used for python version management more often

.github/workflows/mac-artifact-jfrog.yml
```diff
@@ -0,0 +1,123 @@
+name: Mac Artifact
```
dwelch-spike: why doesn't asconfig have a mac-artifact-jfrog workflow? Is it's mac build done differently?

dwelch-spike (changes requested):
Some versoning and clarifying questions

.github/packaging/project/build_package.sh
pvinh-spike: I've committed needed changes for makefiles and with that this file should be changed to:

function build_packages(){
  if [ "$ENV_DISTRO" = "" ]; then
    echo "ENV_DISTRO is not set"
    return
  fi
  chown -R root:root .
  GIT_DIR=$(git rev-parse --show-toplevel)
  cd "$GIT_DIR"
  
  echo "building package for $BUILD_DISTRO"

  if [[ $ENV_DISTRO == *"ubuntu"* ]]; then
    make deb
  elif [[ $ENV_DISTRO == *"debian"* ]]; then
    make deb
  elif [[ $ENV_DISTRO == *"el"* ]]; then
    make rpm
  elif [[ $ENV_DISTRO == *"amzn"* ]]; then
    make rpm
  else
    make tar
  fi

  mkdir -p /tmp/output/$ENV_DISTRO
  cp -a $GIT_DIR/pkg/target/* /tmp/output/$ENV_DISTRO
}

.github/workflows/build-artifacts.yml
```diff
@@ -0,0 +1,76 @@
...
+      retention-days: 7
+      dry-run: false
+      artifactory-url: https://aerospike.jfrog.io # default
+      artifactory-oidc-provider-name: database-gh-aerospike # default
+      artifactory-oidc-audience: database-gh-aerospike # default
+  sign-artifacts:
+    strategy:
+      matrix:
+        distro: [ debian12, debian13, ubuntu20.04, ubuntu22.04, ubuntu24.04, el8, el9, el10, amzn2023 ]
+        host: [ ubuntu-24.04-arm, ubuntu-24.04 ]
```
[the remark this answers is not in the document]
cinterloper: No, this is the name of the github image they provide for arm64
https://docs.github.com/en/actions/reference/runners/github-hosted-runners

cinterloper:
> The PR have 2 set of Dockerfile, one for build and one for test. Should we combine them?

No, they take different build arguments and have different purposes.

cinterloper:
> We need to add this line in all *.yaml : git config --global --add safe.directory "$GITHUB_WORKSPACE"

What does that do?

PR #392, opened 2025-09-15
fix: TOOLS-3115 Show actual server errors instead of generic validation messages in manage config commands

a-spiker:
### Problem
When running `manage config` commands (namespace, service, security, etc.), role violations and other server errors were incorrectly displayed as generic validation messages like "Parameter is not dynamically configurable" instead of showing the actual server error.

**Example:**
```bash
manage config namespace test param replication-factor to 2
# Before: "Parameter is not dynamically configurable"  
# After:  "role violation - requires permission 'set-config'"
```

### Root Cause
The `ASInfoConfigError` class was checking validation conditions (like parameter being non-dynamic) before processing the actual server response, causing real server errors to be hidden.

### Solution
Modified `ASInfoConfigError.__init__()` to check server responses first:
- If server returns an error response (not "ok" or empty) → process and display the actual server error
- Otherwise → fall back to existing validation logic

### Impact
- ✅ **Fixes role violations, authentication failures, and permission errors** across all `manage config` commands
- ✅ **Maintains backward compatibility** for validation errors when no server error exists  
- ✅ **Works with both modern and legacy Aerospike servers**
- ✅ **Universal fix** - covers namespace, service, security, network, XDR, and logging config commands

### Testing
Added comprehensive unit tests covering:
- Server errors taking precedence over validation errors
- Backward compatibility for validation scenarios
- Different error response formats
- Empty response handling

**Files changed:**
- `lib/live_cluster/client/types.py` - Core error handling logic
- `test/unit/live_cluster/client/test_types.py` - Test cases and fixes

### DIFF
--- a/test/unit/live_cluster/client/test_types.py
+++ b/test/unit/live_cluster/client/test_types.py
@@ -248,9 +248,11 @@ class ASInfoConfigErrorTest(unittest.TestCase):
         ]
         # Configure a non-dynamic parameter to trigger the old behavior
         self.node_mock.config_type.return_value = BoolConfigType(False)  # non-dynamic
-        
+
         # Server response indicates role violation
-        role_violation_resp = "ERROR:81:role violation - requires permission 'set-config'"
+        role_violation_resp = (
+            "ERROR:81:role violation - requires permission 'set-config'"
+        )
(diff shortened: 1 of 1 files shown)


dwelch-spike (approved):
LGTM

PR #456, opened 2026-03-20
feat: TOOLS-3251 Add support for set-based secondary indexes in ManageSIndexController

a-spiker:
## Summary
Adds support for creating set-based secondary indexes (server >= 8.1.2) via
`manage sindex create`. A set-based index covers an entire set rather than a
specific bin, so no bin name or bin type is required.

**New command form:**
```
manage sindex create <index-name> ns <ns> set <set>
```
The existing bin-based form is unchanged:
```
manage sindex create <bin-type> <index-name> ns <ns> [set <set>] bin <bin-name> ...
```

### `lib/live_cluster/manage_controller.py`
- Added `_do_create_set` — handles the set-based index path: validates namespace and set name, rejects unsupported modifiers with targeted hints, enforces server version >= 8.1.2, issues `info_sindex_create` with `index_type="set"` and no bin/type args.
- Added `_do_default` — routes to `_do_create_set` when `set` modifier is present (no bin-type); otherwise raises a targeted error:
  - `in <type>` without bin-type → hints the correct bin-based syntax
  - no `set`, no bin-type → shows both valid forms
- Overrides `_format_sub_commands_help` to suppress `do_*` autocomplete hooks from the help output.
- Updated `@CommandHelp`: usage string shows both forms side-by-side; `bin-type` and `in` modifier descriptions updated accordingly.

### `lib/live_cluster/client/node.py`
- `bin_name` / `bin_type` params changed from `str` to `Optional[str]` to support set-based calls.
- `info_sindex_create` now uses `elif bin_name:` so the `bin=`/`type=`/`indexdata=` fields are omitted from the wire command when both are `None`.
- Changed `feature_support["expression_indexing"]` to `feature_support.get("expression_indexing")` to avoid `KeyError` for callers that omit that key.

### `lib/utils/constants.py`
- Added `SERVER_SINDEX_SET_INDEX_FIRST_VERSION = "8.1.2"`.

## Tests
| File | Tests added |
|---|---|
| `test_manage_controller.py` | 12 tests covering: success, version gate, warn/prompt, asinfo error, `in`-only error hint, set+`in` error hint, 4 individual unsupported modifiers, combined unsupported modifiers, no-set/no-bin-type error |
| `test_node.py` | `test_info_sindex_create_set_based` — asserts the wire command contains no `bin=`/`type=`/`indexdata=` fields |
| `test_view.py` | 3 tests for `show_sindex`: basic render with set-based data, `like` filtering, empty-data early return |

---

### `help manage sindex create`

```
Admin> help manage sindex create

Create a new secondary index.

Usage:  manage sindex create COMMAND
or
Usage:  manage sindex create <bin-type> <index-name> ns <ns> [set <set>] [bin <bin-name>] [in <index-type>] [ctx <ctx-item> [. . .]] [ctx_base64 <context>] [exp_base64 <expression>]
        manage sindex create <index-name> ns <ns> set <set>

        bin-type       - The bin type of the provided <bin-name>. Must be one
                         of: numeric, string, geo2dsphere, or blob. Omit
                         bin-type entirely and provide 'set <set>' to create a
                         set-based index (requires server >= 8.1.2).
        index-name     - Name of the secondary index to be created. Should be 20
                         characters or less and not contain ":" or ";".
        ns             - Name of namespace to create the secondary index on.
        set            - Name of set to create the secondary index on.
        bin            - Name of bin to create secondary index on.
        in             - Specifies how the secondary index is to collect keys.
                         list: Use the elements of a list as keys. mapkeys: Use
                         the keys of a map as keys. mapvalues: Use the values of
                         a map as keys. [default: Use the contents of a bin as
                         keys.]
        ctx            - A list of context items describing how to index into a
                         CDT. Possible values include: list_index(<int>)
                         list_rank(<int>) list_value(<value>), map_index(<int>),
                         map_rank(<int>) map_key(<value>), and
                         map_value(<value>). Where <value> i <string>,
                         int(<int>), bool(<bool>), or bytes(<base64>) a base64
                         encoded byte array (no quotes).
        ctx_base64     - The base64 encoding of the context. Cannot be used with
                         'ctx' modifier.
        exp_base64     - The base64 encoding of the expression. ctx, ctx_base64
                         and bin will not be allowed when exp_base64 is
                         specified.

Run 'help manage sindex create COMMAND' for more information on a command.
```

---

### Create a set-based index
```
Admin+> manage sindex create mysetidx ns test set myset
Use 'show sindex' to confirm mysetidx was created successfully.
```

### `show sindex`
```
~~~~~~~Secondary Indexes~~~~~~~
   Index|Namespace|  Set|  Bin|    Bin|  Index|State
    Name|         |     |     |   Type|   Type|
index1  |test     |demo |    k|numeric|default|RW
mysetidx|test     |myset|   --|--     |set    |--
```

---

### Error cases

No bin-type and no set:
```
Admin+> manage sindex create myidx ns test
ERROR: bin-type is required. Must be one of: numeric, string, geo2dsphere, blob.
       To create a set-based index, use: <index-name> ns <ns> set <set>
```

`in` modifier without bin-type:
```
Admin+> manage sindex create myidx ns test in list
ERROR: bin-type is required when using 'in list'.
       Use: <bin-type> <index-name> ns <ns> bin <bin-name> in list
```

Set index with `bin` modifier:
```
Admin+> manage sindex create myidx ns test set myset bin mybin
ERROR: Set type secondary index does not support: bin. Only 'ns' and 'set' modifiers are valid.
```

Set index with `in` modifier:
```
Admin+> manage sindex create myidx ns test set myset in list
ERROR: Set-based indexes do not support the 'in' modifier.
       To index into a CDT with 'in list', specify a bin-type: <bin-type> <index-name> ns <ns> bin <bin-name> in list
```


dwelch-spike (changes requested):
This is looking good. Just want to run a few questions by you to double check things before it is merged.

dwelch-spike (approved):
Nice, assuming tests pass these changes LGTM.

PR #467, opened 2026-04-13
feat: TOOLS-3429 Bifurcate index memory reporting in summary; align stop-writes with server

a-spiker:
**Description:**

Refactors the `summary` command's memory accounting so each component (primary index, secondary index, set index, data) is reported on its own line, and aligns the index-memory stop-writes calculation with the server's `eval_stop_writes()` exactly. Adds a `System Memory` cluster line, a combined `Indexes Memory` budget line, and renames data lines to make the data/index distinction explicit.

**Summary Output Changes:**

- **Bifurcated index reporting** — Secondary index, set index, and primary index now each get their own summary lines (`Pmem/Flash/Shmem Sindex`, `Set Index`, `Pmem/Flash/Shmem Index`). Previously sindex and set-index were folded into the primary-index line.

- **Renamed data lines** to make "data, not indexes" explicit:
  - `Memory` → `Data Memory`
  - `Device` → `Data Device`
  - `Pmem` → `Data Pmem`
  - Underlying dict keys renamed to match (`memory` → `data_memory`, etc.).

- **New `System Memory` cluster line** showing used%/avail% derived from `system_free_mem_pct` — a quick "is this machine running out of RAM?" indicator.

- **New `Indexes Memory` line** (cluster + namespace) that mirrors the server's `eval_stop_writes()` `ixs_sz`: combines non-persisted PI + non-persisted SI + set_index against `indexes-memory-budget`. Emitted only when the budget is configured (>0).

**Stop-Writes Changes:**

- The synthesized `indexes_memory_used` metric now mirrors the server's combined `ixs_sz` exactly (PI/SI excluded when persisted) and is checked against `indexes-memory-budget`.
- Added `Stop%` column for `stop-writes-sys-memory-pct` in the System Memory subgroup of `info namespace`.

**Health Check Changes:**

- Index memory check now uses the server's pre-computed `indexes_memory_used_pct` directly rather than approximating client-side.
- Added a sindex mounts-budget check (`sindex_used_bytes` vs `sindex-type.mounts-budget`).

**Backward Compatibility:**

- Pre-7.0 metric names handled as fallbacks: `memory_used_index_bytes`, `memory_used_sindex_bytes`, `memory_used_set_index_bytes`, `sindex_pmem_used_bytes` / `sindex_flash_used_bytes`, plus pre-7.0 `mounts-size-limit` budgets. Older servers continue to render correctly with the new bifurcation.
- The `Indexes Memory` line is gated on `indexes-memory-budget` (7.1+ config); older servers simply don't show it.

### DIFF
--- a/lib/health/query.py
+++ b/lib/health/query.py
@@ -331,17 +331,12 @@ ASSERT(warn, True, "Low namespace disk available pct.", "OPERATIONS", WARNING,
 				"Namespace disk available pct check.");
     
 SET CONSTRAINT VERSION >= 7.1.0;
-index_bytes = select "index_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
-sindex_bytes = select "sindex_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
-set_index_bytes = select "set_index_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
-ixs_memory_used = do index_bytes + sindex_bytes;
-ixs_memory_used = do ixs_memory_used + set_index_bytes;
-stop_used_bytes = select "indexes-memory-budget" as "stats" from NAMESPACE.CONFIG save;
-budget_configured = do stop_used_bytes > 0;
-critical = do ixs_memory_used <= stop_used_bytes;
-ASSERT(critical, True, "High namespace indexes memory usage (stop-write enabled).", "OPERATIONS", CRITICAL,
-				"Listed namespace[s] have higher than normal combined memory usage for primary index, secondary index, and set index. Probable cause - namespace size misconfiguration.",
-				"Critical Namespace indexes memory used check.", budget_configured);
+ixs_used_pct = select "indexes_memory_used_pct" as "stats" from NAMESPACE.STATISTICS save;
+ixs_has_budget = do ixs_used_pct > 0;
+ixs_critical = do ixs_used_pct <= 100;
+ASSERT(ixs_critical, True, "High namespace indexes memory usage (stop-write enabled).", "OPERATIONS", CRITICAL,
+				"Listed namespace[s] have higher than normal combined memory usage for indexes. Probable cause - namespace size misconfiguration.",
+				"Critical Namespace indexes memory used check.", ixs_has_budget);
 
 SET CONSTRAINT VERSION >= 7.0.0;
 sindex_used = select "sindex_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
--- a/lib/health/query.py
+++ b/lib/health/query.py
@@ -331,13 +331,26 @@ ASSERT(warn, True, "Low namespace disk available pct.", "OPERATIONS", WARNING,
 				"Namespace disk available pct check.");
     
 SET CONSTRAINT VERSION >= 7.1.0;
-used_bytes = select "index_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
+index_bytes = select "index_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
+sindex_bytes = select "sindex_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
+set_index_bytes = select "set_index_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
+ixs_memory_used = do index_bytes + sindex_bytes;
+ixs_memory_used = do ixs_memory_used + set_index_bytes;
 stop_used_bytes = select "indexes-memory-budget" as "stats" from NAMESPACE.CONFIG save;
 budget_configured = do stop_used_bytes > 0;
-critical = do used_bytes <= stop_used_bytes;
-ASSERT(critical, True, "High namespace index memory used pct (stop-write enabled).", "OPERATIONS", CRITICAL,
-				"Listed namespace[s] have higher than normal memory usage for indexes. Probable cause - namespace size misconfiguration.",
-				"Critical Namespace index memory used pct check.", budget_configured);
+critical = do ixs_memory_used <= stop_used_bytes;
+ASSERT(critical, True, "High namespace indexes memory usage (stop-write enabled).", "OPERATIONS", CRITICAL,
+				"Listed namespace[s] have higher than normal combined memory usage for primary index, secondary index, and set index. Probable cause - namespace size misconfiguration.",
+				"Critical Namespace indexes memory used check.", budget_configured);
+
+SET CONSTRAINT VERSION >= 7.0.0;
+sindex_used = select "sindex_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
+sindex_budget = select "sindex-type.mounts-budget" as "stats" from NAMESPACE.CONFIG save;
+sindex_budget_configured = do sindex_budget > 0;
+sindex_critical = do sindex_used <= sindex_budget;
+ASSERT(sindex_critical, True, "High namespace sindex mounts usage.", "OPERATIONS", CRITICAL,
+				"Listed namespace[s] have higher than normal sindex mounts usage. Probable cause - namespace size misconfiguration.",
+				"Critical Namespace sindex mounts used check.", sindex_budget_configured);
 
 SET CONSTRAINT VERSION >= 7.0.0;
 used = select "data_used_pct" as "stats" from NAMESPACE.STATISTICS save;
--- a/lib/health/query.py
+++ b/lib/health/query.py
@@ -333,8 +333,8 @@ ASSERT(warn, True, "Low namespace disk available pct.", "OPERATIONS", WARNING,
 SET CONSTRAINT VERSION >= 7.1.0;
 ixs_used_pct = select "indexes_memory_used_pct" as "stats" from NAMESPACE.STATISTICS save;
 ixs_has_budget = do ixs_used_pct > 0;
-ixs_critical = do ixs_used_pct <= 100;
-ASSERT(ixs_critical, True, "High namespace indexes memory usage (stop-write enabled).", "OPERATIONS", CRITICAL,
+ixs_nominal = do ixs_used_pct <= 100;
+ASSERT(ixs_nominal, True, "High namespace indexes memory usage (stop-write enabled).", "OPERATIONS", CRITICAL,
 				"Listed namespace[s] have higher than normal combined memory usage for indexes. Probable cause - namespace size misconfiguration.",
 				"Critical Namespace indexes memory used check.", ixs_has_budget);
 
@@ -342,8 +342,8 @@ SET CONSTRAINT VERSION >= 7.0.0;
 sindex_used = select "sindex_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
 sindex_budget = select "sindex-type.mounts-budget" as "stats" from NAMESPACE.CONFIG save;
 sindex_budget_configured = do sindex_budget > 0;
-sindex_critical = do sindex_used <= sindex_budget;
-ASSERT(sindex_critical, True, "High namespace sindex mounts usage.", "OPERATIONS", CRITICAL,
+sindex_nominal = do sindex_used <= sindex_budget;
+ASSERT(sindex_nominal, True, "High namespace sindex mounts usage.", "OPERATIONS", CRITICAL,
 				"Listed namespace[s] have higher than normal sindex mounts usage. Probable cause - namespace size misconfiguration.",
 				"Critical Namespace sindex mounts used check.", sindex_budget_configured);
 
--- a/lib/utils/common.py
+++ b/lib/utils/common.py
@@ -1189,11 +1189,15 @@ def create_summary(
                 cluster_pmem_index_used += sindex_used
                 ns_pi = summary_dict["NAMESPACES"][ns].get("pmem_index")
                 if ns_pi:
-                    ns_pi["total"] = ns_pi.get("total", 0) + sindex_size
                     ns_pi["used"] = ns_pi.get("used", 0) + sindex_used
-                    if ns_pi.get("total", 0) > 0:
+                    if "total" in ns_pi:
+                        ns_pi["total"] += sindex_size
                         ns_pi["avail"] = ns_pi["total"] - ns_pi["used"]
-                        ns_pi["used_pct"] = (ns_pi["used"] / ns_pi["total"]) * 100.0
+                        ns_pi["used_pct"] = (
+                            (ns_pi["used"] / ns_pi["total"]) * 100.0
+                            if ns_pi["total"]
+                            else 0
+                        )
                         ns_pi["avail_pct"] = 100.0 - ns_pi["used_pct"]
                 elif sindex_size > 0:
                     sindex_avail = sindex_size - sindex_used
@@ -1214,11 +1218,15 @@ def create_summary(
                 cluster_flash_index_used += sindex_used
                 ns_fi = summary_dict["NAMESPACES"][ns].get("flash_index")
                 if ns_fi:
-                    ns_fi["total"] = ns_fi.get("total", 0) + sindex_size
                     ns_fi["used"] = ns_fi.get("used", 0) + sindex_used
-                    if ns_fi.get("total", 0) > 0:
+                    if "total" in ns_fi:
+                        ns_fi["total"] += sindex_size
                         ns_fi["avail"] = ns_fi["total"] - ns_fi["used"]
-                        ns_fi["used_pct"] = (ns_fi["used"] / ns_fi["total"]) * 100.0
+                        ns_fi["used_pct"] = (
+                            (ns_fi["used"] / ns_fi["total"]) * 100.0
+                            if ns_fi["total"]
+                            else 0
+                        )
                         ns_fi["avail_pct"] = 100.0 - ns_fi["used_pct"]
                 elif sindex_size > 0:
                     sindex_avail = sindex_size - sindex_used
--- a/lib/utils/common.py
+++ b/lib/utils/common.py
@@ -1099,7 +1099,7 @@ def create_summary(
             ).values()
         )[0]
 
-        # Index
+        # Primary Index
         index_size = sum(
             util.get_value_from_second_level_of_dict(
                 ns_configs[ns],
@@ -1139,13 +1139,11 @@ def create_summary(
                     "used_pct": index_used_pct,
                 }
             else:
-                # shmem does not require you to configure mounts-budget
                 ns_index_usage = {
                     "used": index_used,
                 }
 
             if index_type == "pmem":
-                # TODO handle the cluster level aggregate
                 cluster_pmem_index_total += index_size
                 cluster_pmem_index_used += index_used
                 summary_dict["NAMESPACES"][ns]["pmem_index"] = ns_index_usage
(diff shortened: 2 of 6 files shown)


lib/utils/common.py
```diff
@@ -1160,6 +1158,110 @@ def create_summary(
...
+                cluster_flash_index_total += sindex_size
+                cluster_flash_index_used += sindex_used
+                ns_fi = summary_dict["NAMESPACES"][ns].get("flash_index")
+                if ns_fi:
+                    ns_fi["total"] = ns_fi.get("total", 0) + sindex_size
+                    ns_fi["used"] = ns_fi.get("used", 0) + sindex_used
+                    if ns_fi.get("total", 0) > 0:
+                        ns_fi["avail"] = ns_fi["total"] - ns_fi["used"]
+                        ns_fi["used_pct"] = (ns_fi["used"] / ns_fi["total"]) * 100.0
+                        ns_fi["avail_pct"] = 100.0 - ns_fi["used_pct"]
```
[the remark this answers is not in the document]
a-spiker: This is not possible as total will always be there, This is a good defensive mechanism, I'll keep it.

lib/utils/common.py
```diff
@@ -1160,6 +1158,110 @@ def create_summary(
...
+                cluster_pmem_index_total += sindex_size
+                cluster_pmem_index_used += sindex_used
+                ns_pi = summary_dict["NAMESPACES"][ns].get("pmem_index")
+                if ns_pi:
+                    ns_pi["total"] = ns_pi.get("total", 0) + sindex_size
+                    ns_pi["used"] = ns_pi.get("used", 0) + sindex_used
+                    if ns_pi.get("total", 0) > 0:
+                        ns_pi["avail"] = ns_pi["total"] - ns_pi["used"]
+                        ns_pi["used_pct"] = (ns_pi["used"] / ns_pi["total"]) * 100.0
+                        ns_pi["avail_pct"] = 100.0 - ns_pi["used_pct"]
```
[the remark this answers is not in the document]
a-spiker: Same as comment 1.

lib/health/query.py
```diff
@@ -331,13 +331,26 @@
...
 stop_used_bytes = select "indexes-memory-budget" as "stats" from NAMESPACE.CONFIG save;
 budget_configured = do stop_used_bytes > 0;
-critical = do used_bytes <= stop_used_bytes;
-ASSERT(critical, True, "High namespace index memory used pct (stop-write enabled).", "OPERATIONS", CRITICAL,
-				"Listed namespace[s] have higher than normal memory usage for indexes. Probable cause - namespace size misconfiguration.",
-				"Critical Namespace index memory used pct check.", budget_configured);
+critical = do ixs_memory_used <= stop_used_bytes;
+ASSERT(critical, True, "High namespace indexes memory usage (stop-write enabled).", "OPERATIONS", CRITICAL,
+				"Listed namespace[s] have higher than normal combined memory usage for primary index, secondary index, and set index. Probable cause - namespace size misconfiguration.",
+				"Critical Namespace indexes memory used check.", budget_configured);
```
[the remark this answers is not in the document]
a-spiker: Changing this to indexes_memory_used_pct for better approach

lib/health/query.py
```diff
@@ -331,13 +331,21 @@
...
-used_bytes = select "index_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
-stop_used_bytes = select "indexes-memory-budget" as "stats" from NAMESPACE.CONFIG save;
-budget_configured = do stop_used_bytes > 0;
-critical = do used_bytes <= stop_used_bytes;
-ASSERT(critical, True, "High namespace index memory used pct (stop-write enabled).", "OPERATIONS", CRITICAL,
-				"Listed namespace[s] have higher than normal memory usage for indexes. Probable cause - namespace size misconfiguration.",
-				"Critical Namespace index memory used pct check.", budget_configured);
+ixs_used_pct = select "indexes_memory_used_pct" as "stats" from NAMESPACE.STATISTICS save;
+ixs_has_budget = do ixs_used_pct > 0;
+ixs_critical = do ixs_used_pct <= 100;
```
dwelch-spike: This is actually true when index usage is normal right? I think it should be called something like "idxs_nominal"

lib/health/query.py
```diff
@@ -331,13 +331,21 @@
...
+ASSERT(ixs_critical, True, "High namespace indexes memory usage (stop-write enabled).", "OPERATIONS", CRITICAL,
+				"Listed namespace[s] have higher than normal combined memory usage for indexes. Probable cause - namespace size misconfiguration.",
+				"Critical Namespace indexes memory used check.", ixs_has_budget);
+
+SET CONSTRAINT VERSION >= 7.0.0;
+sindex_used = select "sindex_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
+sindex_budget = select "sindex-type.mounts-budget" as "stats" from NAMESPACE.CONFIG save;
+sindex_budget_configured = do sindex_budget > 0;
+sindex_critical = do sindex_used <= sindex_budget;
```
dwelch-spike: This is true when sindex usage is normal right? I think it should be called something like "sindex_nominal".

lib/utils/common.py
```diff
@@ -1160,6 +1158,118 @@ def create_summary(
                 summary_dict["NAMESPACES"][ns]["shmem_index"] = ns_index_usage
                 summary_dict["NAMESPACES"][ns]["index_type"] = index_type
 
+        # Secondary Index — fold into the matching primary index accumulator
+        # so it appears in the same summary line (mirrors server ixs_memory_used).
+        sindex_type = list(
+            util.get_value_from_second_level_of_dict(
+                ns_stats, ("sindex-type",), default_value="", return_type=str
+            ).values()
+        )[0]
```
dwelch-spike: Why turn this into a list only to grab the first element?

lib/utils/common.py
```diff
@@ -1647,21 +1757,40 @@ def _format_ns_stop_writes_metrics(
...
                 stats, config, default_value=None, return_type=int
             )
 
-            if usage is not None and threshold is not None:
-                sw = _is_stop_writes_cause(usage, threshold, stop_writes)
+            if threshold is not None and threshold > 0:
+                pi_persisted = stats.get("index-type", "") in ("flash", "pmem")
+                si_persisted = stats.get("sindex-type", "") in ("flash", "pmem")
+
+                index_mem_sz = (
```
dwelch-spike: How about we call this `index_mem_used` to be consistent with the previous calculations?

lib/utils/common.py
```diff
@@ -1647,21 +1757,40 @@ def _format_ns_stop_writes_metrics(
...
+                    0
+                    if pi_persisted
+                    else util.get_value_from_dict(
+                        stats, "index_used_bytes", default_value=0, return_type=int
+                    )
+                )
+                set_index_sz = util.get_value_from_dict(
+                    stats, "set_index_used_bytes", default_value=0, return_type=int
+                )
+                sindex_mem_sz = (
```
dwelch-spike: How about we call this `sindex_mem_used` to be consistent with the previous calculations?

lib/utils/common.py
```diff
@@ -1647,21 +1757,40 @@ def _format_ns_stop_writes_metrics(
...
+                si_persisted = stats.get("sindex-type", "") in ("flash", "pmem")
+
+                index_mem_sz = (
+                    0
+                    if pi_persisted
+                    else util.get_value_from_dict(
+                        stats, "index_used_bytes", default_value=0, return_type=int
+                    )
+                )
+                set_index_sz = util.get_value_from_dict(
```
dwelch-spike: How about calling this `set_index_used`?

lib/utils/common.py
```diff
@@ -1647,21 +1757,40 @@ def _format_ns_stop_writes_metrics(
...
+                    stats, "set_index_used_bytes", default_value=0, return_type=int
+                )
+                sindex_mem_sz = (
+                    0
+                    if si_persisted
+                    else util.get_value_from_dict(
+                        stats, "sindex_used_bytes", default_value=0, return_type=int
+                    )
+                )
+                ixs_sz = index_mem_sz + set_index_sz + sindex_mem_sz
```
dwelch-spike: nit: I think "usage" is more clear, previously size was used to name index budgets.

lib/utils/common.py
```diff
@@ -1647,21 +1757,40 @@ def _format_ns_stop_writes_metrics(
...
+
+                metric = "indexes_memory_used"
```
dwelch-spike, line 1916: Where does this name come from? I don't see a `indexes_memory_used` in the metric reference for the server. Is it just the name of something in the summary output?

dwelch-spike (changes requested):
Double checking some things and I'm seeing some confusing results locally. I'm seeing the sindex usage reflected in the shmem index line but not in the total memory usage. Have a suggestion for another approach.

```
Admin+> summary -l
Cluster
=======

   1.   Cluster Name             :  cakery
   2.   Server Version           :  E-8.1.2.0
   3.   OS Version               :  Ubuntu 24.04.4 LTS (6.17.0-20-generic)
   4.   Cluster Size             :  1
   5.   Devices                  :  Total 0, per-node 0
   6.   Shmem Index              :  284.288 MB used
   7.   Memory                   :  Total 8.000 GB, 0.74% used (60.380 MB), 95.50% available contiguous space (7.640 GB)
   8.   License Usage            :  Latest: (60.380 MB) ?
   9.   Active Namespaces        :  1 of 2
   10.  Active Features          :  Index-on-shmem, KVS, PIndex Query


Namespaces
==========

   test
   ====
   1.   Devices                  :  Total 0, per-node 0
   2.   Shmem Index              :  284.288 MB used
   3.   Memory                   :  Total 4.000 GB, 1.47% used (60.380 MB), 93.00% available contiguous space (3.720 GB)
   4.   License Usage            :  Latest: (60.380 MB) ?
   5.   Replication Factor       :  2
   6.   Rack-aware               :  False
   7.   Master Objects           :  987.377 K
   8.   Compression-ratio        :  1.0
```

I think folding everything into the index usage line is confusing because "index" makes me think primary index. I propose splitting these into separate lines, clarifying that the "Memory" line only tracks data, and adding a system free mem pct line to make checking total memory usage easy.

My suggestion: 

1. Split sindex into separate lines (instead of folding into primary index)
Add shmem_sindex, pmem_sindex, flash_sindex as distinct summary dict keys
Each gets its own line: "Shmem Sindex", "Pmem Sindex", "Flash Sindex"
pmem/flash variants show total/used/avail/pct (from sindex-type.mounts-budget)
shmem variant shows used-only (no configured budget)

2. Separate line for set index
Add set_index as its own summary dict key
Displayed as "Set Index" -- always RAM, used-only

3. Rename data storage lines for clarity. These only count storage used for data, not indexes, so we should clarify that to avoid confusion in the future.

"Memory" -> "Data Memory"
"Device" -> "Data Device"
"Pmem" -> "Data Pmem"

4. Add system memory line at cluster level
New "System Memory" line using service-level stats (system_free_mem_pct)
Gives a quick "is this machine running out of RAM?" indicator
Example output (shmem namespace, storage-engine=memory):

```
Cluster
=======
   1.   System Memory            :  85.00% free
   2.   Shmem Index              :  200.000 MB used
   3.   Shmem Sindex             :  64.000 MB used
   4.   Set Index                :  20.288 MB used
   5.   Data Memory              :  Total 8.000 GB, 0.74% used (60.380 MB), 95.50% available contiguous space (7.640 GB)
   ...
```
Example output (flash index namespace, storage-engine=device):

```
   1.   Flash Index              :  Total 50.000 GB, 25.00% used (12.500 GB), ...
   2.   Shmem Sindex             :  64.000 MB used
   3.   Set Index                :  20.288 MB used
   4.   Data Device              :  Total 100.000 GB, 45.00% used (45.000 GB), ...
   ...
```
Files to change:

lib/utils/common.py -- separate accumulators for sindex/set_index, add system memory to cluster summary
lib/view/view.py -- add new lines to list view, rename existing labels
lib/view/templates.py -- add new fields to table view, rename subgroups/display names
lib/health/query.py -- no changes needed (already handles sindex separately)
test/unit/utils/test_common.py -- update expected dicts for new keys

lib/health/query.py
a-spiker: Done — renamed to `ixs_nominal` (true when normal/under budget).

lib/utils/common.py
```diff
@@ -1160,6 +1196,116 @@ def create_summary(
...
+                "used": ixs_used,
+                "used_pct": ixs_used_pct,
+                "avail": ixs_avail,
+                "avail_pct": 100.0 - ixs_used_pct,
+            }
+            summary_dict["NAMESPACES"][ns]["indexes_memory"] = ns_indexes_memory
+            cluster_indexes_memory_total += ixs_budget
+            cluster_indexes_memory_used += ixs_used
+
         storage_engine_type = list(
```
dwelch-spike: From Architecture agent: **Important:** This site still uses `list(...)[0]` while the same anti-pattern was rewritten to `next(iter(..., return_type=str).values()), "")` two hunks above for `index_type` (line 1130) and `sindex_type` (line 1200) in this same PR. Three sibling reads of "single value from a second-level dict" should behave identically. If `storage-engine` is absent or no node reports it, the `list(...)[0]` form raises `IndexError` and aborts summary generation for every following namespace; the index/sindex sites now degrade gracefully. Fix: switch to the same `next(iter(...), "")` form here.

lib/utils/common.py
```diff
@@ -1160,6 +1196,116 @@ def create_summary(
...
+
+            if sindex_type == "pmem":
+                cluster_pmem_sindex_total += sindex_size
+                cluster_pmem_sindex_used += sindex_used
+                summary_dict["NAMESPACES"][ns]["pmem_sindex"] = ns_sindex_usage
+            elif sindex_type == "flash":
+                cluster_flash_sindex_total += sindex_size
+                cluster_flash_sindex_used += sindex_used
+                summary_dict["NAMESPACES"][ns]["flash_sindex"] = ns_sindex_usage
+            else:
```
dwelch-spike, line 1260: From Logic agent: **Important:** Asymmetric handling of unknown `index-type` vs `sindex-type`. The PI branch above (line 1184: `if pmem / elif flash / elif shmem`) is closed — an unknown or empty `index_type` silently drops `index_used` from both the cluster shmem aggregate and the per-namespace breakdown. The SI branch here uses `else` and treats unknown as shmem. Two policies for the same shape of data in the same function. The downstream effect: if `index_type` is ever unrecognized (e.g. a future server adds a new storage type), the bytes still flow into `ixs_used` at line 1293 (unconditionally, since `pi_persisted = index_type in ("flash","pmem")` is False for unknowns), so the `Indexes Memory` aggregate counts them while the PI breakdown silently doesn't render. Pick one. The SI form (open `else` defaulting to shmem) is safer — converting PI to `if pmem / elif flash / else shmem` makes both branches mean the same thing.

test/unit/view/test_view.py
```diff
@@ -821,9 +821,9 @@ def test_summary_cluster_list_view(self):
    7.   Flash Index              :  Total 1.000 B, 4.00% used (2.000 B), 5.00% available (3.000 B)
    8.   Shmem Index              :  Total 1.000 B, 4.00% used (2.000 B), 5.00% available (3.000 B)
    9.   Memory                   :  Total 1.000 B, 4.00% used (2.000 B), 5.00% available (3.000 B) includes data, pindex, and sindex
-   10.  Memory                   :  Total 1.000 B, 4.00% used (2.000 B), 5.00% available contiguous space (3.000 B)
-   11.  Device                   :  Total 1.000 B, 4.00% used (2.000 B), 5.00% available contiguous space (3.000 B)
-   12.  Pmem                     :  Total 1.000 B, 4.00% used (2.000 B), 5.00% available contiguous space (3.000 B)
+   10.  Data Memory              :  Total 1.000 B, 4.00% used (2.000 B), 5.00% available contiguous space (3.000 B)
```
dwelch-spike: From Tests agent: **Important:** `test_summary_cluster_list_view` and `test_summary_namespace_list_view` only got mechanical renames (memory→data_memory etc.) — they never inject any of the new summary keys this PR adds: `system_memory`, `shmem_sindex`, `pmem_sindex`, `flash_sindex`, `set_index`, or `indexes_memory`. If `_summary_cluster_list_view` forgot to wire one of those branches, or formatted `System Memory` with wrong units, no test fails — I verified live against an 8.1.2 server that the lines render, but the rendering paths in lib/view/view.py (lines 2352-2422 cluster, 2547-2572 namespace) have no unit-test coverage at all. Also: the new `Stop%` column for `stop-writes-sys-memory-pct` in lib/view/templates.py is not referenced anywhere in test_view.py (`grep -n 'stop-writes-sys-memory-pct\|Stop%' test/unit/view/test_view.py` returns nothing). Add a case per new key to the expected-output strings here, and an actual-output test for the Stop% column.

dwelch-spike:
Pulled the branch and ran `asadm.py` against a local 8.1.2 server. The new output matches what I asked for in the prior review:

**Default summary (no `indexes-memory-budget` configured)** — sindex now reads as its own line, data lines are explicit, and System Memory gives the quick at-a-glance check I wanted:
```
Cluster
=======
   1.   Cluster Name             :  cakery
   2.   Server Version           :  E-8.1.2.0-start-48-g4a81ee959
   ...
   6.   System Memory            :  62.00% free
   7.   Shmem Index              :  54.004 MB used
   8.   Shmem Sindex             :  224.000 MB used
   9.   Data Memory              :  Total 8.000 GB, 0.66% used (54.004 MB), 95.50% available contiguous space (7.640 GB)
```

**After setting `indexes-memory-budget=1GiB`** — the new `Indexes Memory` aggregate appears, and `show stop-writes` gains the matching `indexes-memory-budget` row:
```
   9.   Indexes Memory           :  Total 1.000 GB, 27.15% used (278.004 MB), 72.85% available (745.996 MB)
```
```
indexes-memory-budget   |test  |...|False  |indexes_memory_used  |27.15 %|       291.508 M|          1.074 G
stop-writes-sys-memory-pct|test|...|False  |system_free_mem_pct  |43.33 %|          39.0 %|           90.0 %
```
(`278.004 MB` in summary = `291.508 M` in stop-writes — same bytes, different MiB-vs-M unit conventions in those two views; pre-existing display inconsistency unrelated to this PR.)

**How this addresses the prior review:**

1. ✅ **Split sindex into separate lines** — `Shmem Sindex`/`Pmem Sindex`/`Flash Sindex` are now distinct lines (verified `Shmem Sindex` rendering live; the other two are dead branches without a non-shmem sindex test cluster, but the code path is symmetric).
2. ✅ **Separate `Set Index` line** — code path present at view.py:2409-2415 and 2562-2565; my test server has no set indexes so the line is correctly suppressed.
3. ✅ **Renamed data lines** — `Memory`→`Data Memory`, `Device`→`Data Device`, `Pmem`→`Data Pmem`. Underlying dict keys renamed to match.
4. ✅ **System Memory cluster line** — shows `62.00% free` derived from `system_free_mem_pct`. Clear at-a-glance RAM indicator.
5. ➕ **Bonus `Indexes Memory` aggregate line** — not in my original proposal but useful: gated on `indexes-memory-budget > 0` so older servers don't show a spurious line, and mirrors the server's `eval_stop_writes()` `ixs_sz` so the threshold/usage in `show stop-writes` matches the summary line.

All 150 unit tests pass locally.

Inline comments above cover the remaining concerns (one residual `list(...)[0]` site, PI/SI asymmetric handling, missing view-layer test coverage for the new keys + Stop% column).

lib/utils/common.py
a-spiker: Done

lib/utils/common.py
a-spiker: Done

test/unit/view/test_view.py
a-spiker: Done

dwelch-spike (approved):
LGTM

PR #469, opened 2026-04-14
fix: TOOLS-3443 Include index-type prefix in namespace set-config commands

a-spiker:
## Summary

- Removed `index-type` from the subcontext exclusion set in `info_set_config_namespace`, so that `index-type` sub-parameters (like `evict-mounts-pct` and `mounts-budget`) correctly get the `index-type.` prefix in the generated `set-config` info command.
- The Aerospike server expects `index-type.evict-mounts-pct` and `index-type.mounts-budget` as exact parameter names (confirmed via server source in `cfg_set_namespace`). The previous blanket exclusion (added in TOOLS-2768) was dropping this prefix, causing the server to reject these parameters with `ERROR:4:invalid set-config parameter`.
- `storage-engine` remains correctly excluded since the server expects bare names (e.g. `defrag-lwm-pct`, not `storage-engine.defrag-lwm-pct`) for storage-engine parameters.
- Updated unit tests to use actual dynamic `index-type` parameters (`evict-mounts-pct`, `mounts-budget`) instead of the non-dynamic `mounts-size-limit` used previously.

## Jira

[TOOLS-3443](https://aerospike.atlassian.net/browse/TOOLS-3443)

## Test plan

- [x] Unit tests pass (`test_info_set_config_namespace_success`, `test_info_set_config_namespace_fail`, `test_info_set_config_namespace_with_namespaces_error`)
- [x] Manual test: `manage config namespace <ns> index-type param evict-mounts-pct to <value>` on a namespace with pmem/flash index-type produces correct `set-config:context=namespace;namespace=<ns>;index-type.evict-mounts-pct=<value>`
- [x] Manual test: `manage config namespace <ns> storage-engine param defrag-lwm-pct to <value>` still produces bare `defrag-lwm-pct` (no `storage-engine.` prefix)
- [x] Manual test: `manage config namespace <ns> geo2dsphere-within param max-cells to <value>` still produces `geo2dsphere-within-max-cells` (hyphen delimiter)

[TOOLS-3443]: https://aerospike.atlassian.net/browse/TOOLS-3443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/live_cluster/client/node.py
```diff
@@ -2089,7 +2089,7 @@ async def info_set_config_namespace(
         self, param, value, namespace, set_=None, subcontext=None
     ):
         new_param = param
-        if subcontext and subcontext not in {"storage-engine", "index-type"}:
+        if subcontext and subcontext != "storage-engine":
```
dwelch-spike, line 2098: The ticket says "This was done to fix the case where mounts-size-limit under index-type should be sent without the prefix". Are you sure that changing this to always include .index-type is ok? What about for mounts-size-limit on old servers?

dwelch-spike: I had a look at the server code and it looks like the last time `mounts-size-limit` without the `index-type.` prefix was required was in server 6.2. So I'm not concerned about backwards compatibility here since that version is long out of support. Also, it looks like in modern server versions, anything under index-type requires the `index-type.` prefix so I think this change is OK.

dwelch-spike (dismissed):
LGTM

dwelch-spike (approved):
LGTM

PR #475, opened 2026-05-10
fix: TOOLS-3445 Validate file/directory types for TLS path options

a-spiker:
## Summary
- `--tls-capath` silently accepted a file path instead of a directory, causing TLS connections to fail without any diagnostic message; added an explicit check that raises a clear error
- Extended the same pattern to `--tls-cafile`, `--tls-certfile`, and `--tls-keyfile`, which previously produced confusing errors (e.g. "Unable to locate key file" when a directory was passed)
- Fixed a latent variable-shadowing bug: the local variable `path = ""` in the `except` block was shadowing the `os.path` module import, which would have caused `UnboundLocalError` on any future use of `path.exists()` in the same function scope

## Test plan
- [x] `test_capath_as_file_raises` — asserts clear error when `--tls-capath` receives a file
- [x] `test_cafile_as_directory_raises` — asserts clear error when `--tls-cafile` receives a directory
- [x] `test_certfile_as_directory_raises` — asserts clear error when `--tls-certfile` receives a directory
- [x] `test_keyfile_as_directory_raises` — asserts clear error when `--tls-keyfile` receives a directory
- [x] Full `test_ssl_context.py` suite passes (111 tests)


dwelch-spike (approved):
LGTM

PR #476, opened 2026-05-10
fix: TOOLS-3456 Deprecate health command and fix stop-writes index memory checks for flash/pmem

a-spiker:
## Summary

Fixes TOOLS-3456. Two parts:

1. **Deprecates the `health` command** in both live-cluster and collectinfo modes (help text plus a runtime warning), as agreed with product.
2. **Replaces the incorrect stop-writes attribution for persisted (flash/pmem) indexes with health checks and view highlighting that match actual server behavior**, on all supported server versions.

## Background

The original bug: asadm compared `index_used_bytes` against `indexes-memory-budget` regardless of `index-type`, producing false CRITICAL alerts for All-Flash namespaces (e.g. ~956 GB flash index vs ~7.5 GB RAM budget). The core comparison was fixed on master by TOOLS-3429 (#467), which aligned `show stop-writes` and the 7.1+ health check with the server (`indexes_memory_used_pct` is server-computed and excludes persisted indexes).

An earlier revision of this PR then surfaced `index_used_bytes` vs `index-type.mounts-budget` as a *stop-writes* metric. Review (see inline comments) established that the server never triggers stop-writes on mounts usage: `eval_stop_writes()` (nsup.c) zeroes out the persisted index contribution and never consults `pi_mounts_budget`. Mounts budget instead drives **eviction** (`evict-mounts-pct`, `mounts-high-water-pct` pre 7.0) via `eval_hwm_breached()`, The budget itself is not enforced in the allocation path (it is a startup sanity check and the eviction/statistics denominator): index allocations fail with per-record write errors when the mounts themselves fill, which coincides with the budget when the budget reflects actual mount capacity, and `stop_writes` stays false throughout. Verified against server tags 4.3.0.10 through 8.1.1.2.

This PR now implements the redesign agreed in the review thread instead.

### Deprecate `health` command
- `lib/live_cluster/health_check_controller.py`, `lib/collectinfo_analyzer/health_check_controller.py`: deprecation note in `CommandHelp`, `(deprecated)` tag in the short help, and a `logger.warning` at the start of `_do_default`. The stale "still in beta" sentence is dropped from the live-cluster long help. The live-cluster controller keeps its pre-existing `hide=True`, so the short-help tag only renders in collectinfo mode; live-cluster users see the long-help note and the runtime warning.

### Health checks for persisted index mounts usage (`lib/health/query.py`)
New cross-version checks that describe what the server actually does. Stat and config names are alias-merged so one check covers all versions:

| Check | Level | 7.0+ | pre 7.0 |
|---|---|---|---|
| PI mounts usage >= evict threshold (skipped when 0/unset) | WARNING | `index_mounts_used_pct` vs `index-type.evict-mounts-pct` | `index_flash_used_pct` / `index_pmem_used_pct` vs `index-type.mounts-high-water-pct` (4.3+) |
| PI mounts usage >= 95% of budget | CRITICAL | `index-type.mounts-budget` | `index-type.mounts-size-limit` (4.3+) |
| PI flash arena allocation >= 95% of budget (alloc runs ahead of usage: stages are never returned) | WARNING | `index_flash_alloc_pct` | `index_flash_alloc_pct` (5.6+, the stat does not exist earlier) |
| Sindex mounts usage >= evict threshold | WARNING | `sindex_mounts_used_pct` vs `sindex-type.evict-mounts-pct` | `sindex_pmem_used_pct` (6.3+) / `sindex_flash_used_pct` (6.4+) vs `sindex-type.mounts-high-water-pct` |
| Sindex mounts usage >= 95% of budget | CRITICAL | `sindex-type.mounts-budget` | `sindex-type.mounts-size-limit` (6.3+) |
| Combined indexes memory >= `evict-indexes-memory-pct` (7.1+, complements the existing stop-writes CRITICAL) | WARNING | `indexes_memory_used_pct` | n/a |

- Version gates match stat introduction: PI mounts checks run from 4.3, the flash alloc check from 5.6 (`SET CONSTRAINT VERSION >= 5.6.0`), and the sindex mounts checks from 6.3 (`SET CONSTRAINT VERSION >= 6.3.0`; persisted sindex does not exist earlier). Verified by grepping server source at tags 4.3.0.10 through 8.1.1.2.
- Cross-source (stats vs config) comparisons use `on common`, so a namespace missing the evict config key is dropped individually instead of the key-set mismatch disabling the WARNING for every namespace.
- The CRITICAL messages are worded around the real failure mode (allocation failures / write errors as the mounts fill), not stop-writes and not a hard budget cap.
- Removed the old `sindex_used_bytes > sindex-type.mounts-budget` CRITICAL: it only fired above 100% of the budget (after writes were already failing) and is strictly subsumed by the new 95% check, which also extends coverage to pre-7.0 servers.
- The mounts pct statistics are only emitted for flash/pmem namespaces, so in-memory index namespaces are naturally excluded.

### `info namespace usage` highlighting (`lib/view/templates.py`)
Primary and Secondary Index `Used%` now have a second alert tier:

| Condition | Color | Meaning |
|---|---|---|
| Used% < Evict% | none | healthy |
| Used% >= Evict% (and Evict% configured) | yellow (existing) | eviction running |
| Used% >= 90% | red (new) | near the mounts/memory budget, allocations may start failing |

### `show stop-writes`
No change needed. Master's version (TOOLS-3429) already mirrors `eval_stop_writes()` exactly; the misleading mounts-budget entry from the earlier revision of this PR is gone.

## Tests

- `test/unit/health/test_healthchecker.py`: new `PersistedIndexHealthCheckTests` (15 tests) running the real `HealthChecker` against synthetic single-node inputs, in the individual-per-check style the file's TODO asks for. Covers: the exact ticket scenario (healthy All-Flash, no alerts), eviction-range WARNING, 95% CRITICAL, pre-7.0 stat/config names for flash and pmem PI on a 6.4 server, pre-7.0 sindex names (flash on 6.4, pmem on 6.3, pinning the 6.3 gate), a partial-config scenario proving `on common` keeps the WARNING for configured namespaces, a positive control proving checks after the mounts block still execute on memory-only input, `evict-mounts-pct 0` suppressing the WARNING while keeping the CRITICAL, flash arena alloc WARNING, sindex tiers, and the 7.1+ combined-memory checks. These also guard against DSL syntax errors in the new queries (the health query language terminates statements on `;` even inside strings, which these tests caught during development).
- `test/unit/view/test_templates.py`: parameterized tests for the `Used%` formatter tiers (red wins over yellow, red applies even with eviction disabled, healthy values unformatted) for both index subgroups, rendered through the real `sheet.render` pipeline so formatter precedence is exercised, not mirrored.
- `test/unit/live_cluster/test_health_controller.py`, `test/unit/collectinfo_analyzer/test_health_controller.py`: new tests asserting the deprecation warning is logged and `_do_default` still runs to completion in both modes.
- Full unit suite: 1490 passed. `black --check` clean.

## Test plan

- [x] Unit suite passes (`pytest test/unit`)
- [x] Health checker executes the full query file with zero syntax exceptions
- [x] Manual: `health` against a live cluster shows the deprecation warning
- [x] Manual: `info namespace usage` on a flash namespace shows yellow at Evict% and red at >= 90%


dwelch-spike:
@a-spiker does @rbotzer know we are deprecating the health command? Is there a plan in place for a replacement?

a-spiker:
> @a-spiker does @rbotzer know we are deprecating the health command? Is there a plan in place for a replacement?

Yes,  he agrees to this. And he agreed to the health text.

lib/health/query.py
```diff
@@ -332,12 +332,27 @@
...
+ASSERT(critical, True, "High namespace index memory used bytes (stop-write enabled).", "OPERATIONS", CRITICAL,
 				"Listed namespace[s] have higher than normal memory usage for indexes. Probable cause - namespace size misconfiguration.",
-				"Critical Namespace index memory used pct check.", budget_configured);
+				"Critical Namespace index memory used bytes check.", memory_check);
+
+mounts_budget = select "index-type.mounts-budget" as "stats" from NAMESPACE.CONFIG save;
+mounts_budget_configured = do mounts_budget > 0;
+persisted_check = do mounts_budget_configured && is_persisted;
+mounts_critical = do used_bytes <= mounts_budget;
+ASSERT(mounts_critical, True, "High namespace index mounts used bytes.", "OPERATIONS", CRITICAL,
```
dwelch-spike: Per `eval_stop_writes` (nsup.c:942-1021), `index-type.mounts-budget` is not on the stop-writes path. The server zeros out the persisted PI contribution to `ixs_sz` (nsup.c:947-950) and never compares `index_used_bytes` against `pi_mounts_budget` for stop-writes purposes. `pi_mounts_budget` drives eviction (via `pi_evict_mounts_pct`) and is an arena allocation cap — when exhausted, the arena logs a ticker warning and individual allocations fail, but `ns->stop_writes` stays `false`.

So a CRITICAL ASSERT here will only ever fire if stop-writes was already set for another reason (sys memory / data-avail / data-used / clock skew) while mounts are coincidentally near budget — misattributing the cause.

Suggestion: downgrade to WARNING and reword the messages to reflect actual server behavior, e.g.:
- short: "High namespace index mounts usage."
- description: "Index mounts approaching `index-type.mounts-budget`. Eviction triggers at `index-type.evict-mounts-pct`. Consider raising `index-type.mounts-budget` or evicting more aggressively."

That way the alert is correctly framed as a capacity/eviction concern rather than a stop-writes trigger.

lib/health/query.py
```diff
@@ -332,12 +332,27 @@
...
 used_bytes = select "index_used_bytes" as "stats" from NAMESPACE.STATISTICS save;
+index_type = select "index-type" as "stats" from NAMESPACE.STATISTICS;
+is_flash = do index_type == "flash";
+is_pmem = do index_type == "pmem";
+is_persisted = do is_flash || is_pmem;
+is_not_persisted = do is_persisted == False;
+
 stop_used_bytes = select "indexes-memory-budget" as "stats" from NAMESPACE.CONFIG save;
 budget_configured = do stop_used_bytes > 0;
+memory_check = do budget_configured && is_not_persisted;
```
dwelch-spike: Gating the `indexes-memory-budget` check to non-persisted indexes drops a check that *is* still relevant for flash/pmem namespaces. Per nsup.c:951-955, `ixs_sz = index_mem_sz + set_index_sz + sindex_mem_sz`, where:
- `index_mem_sz = 0` for persisted PI (correct gate)
- `set_index_sz` is **always** counted regardless of persistence
- `sindex_mem_sz = 0` only when sindex is persisted

So a flash/pmem PI namespace with a memory-resident set-index or non-persisted sindex can still trip `indexes-memory-budget` on the server, but this ASSERT will no longer fire for those namespaces.

This isn't a regression introduced by the PR — the old check was using `index_used_bytes` (the PI's bytes), so it wasn't really catching the set-index/sindex case either. But the explicit `is_not_persisted` gate now hides that completely. Could be worth a follow-up assert that compares `set_index_used_bytes + sindex_used_bytes (when sindex is memory)` to `indexes-memory-budget` for persisted-PI namespaces.

lib/utils/common.py
```diff
@@ -1647,11 +1647,21 @@ def _format_ns_stop_writes_metrics(
...
+            )
+            is_persisted = index_type in ("flash", "pmem")
+
             metric = "index_used_bytes"
-            config = "indexes-memory-budget"
             usage = util.get_value_from_dict(
                 stats, metric, default_value=None, return_type=int
             )
+
+            if is_persisted:
```
dwelch-spike: Same concern as the inline note on `lib/health/query.py` for the new mounts ASSERT: the only place `ns->stop_writes` is set is `eval_stop_writes` (nsup.c:1009), and that function never consults `pi_mounts_budget`. Attributing a `show stop-writes` entry to `index-type.mounts-budget` will be misleading — the field will appear under stop-writes only because the test fixture (or a coincidentally-true ns-level `stop_writes` from another cause) lets `_is_stop_writes_cause` return True.

Two options:
1. Keep the entry but suppress `stop_writes: True` for the persisted case (always report `stop_writes: False`, so the row is informational only).
2. Move this metric out of the stop-writes summary into a separate "high index mounts usage" surface, since it's an eviction/capacity concern rather than a stop-writes one.

Either way, the current tests assert behavior the server can't actually produce on its own (e.g. `flash, index_used_bytes=90, mounts-budget=90 → stop_writes=True`). The PASS only happens because the fixture explicitly sets `"stop_writes": "true"` on the ns — which on a real cluster would have to be caused by sys-memory / data-avail-pct / data-used-pct / clock-skew.

dwelch-spike (commented):
## Review summary

Cross-checked against the server source under `server/aerospike-server/` — `as/src/base/nsup.c` (stop-writes eval), `as/src/base/thr_info.c` (stat emissions), `as/src/base/cfg_info.c` (config emissions), `as/src/storage/drv_ssd.c` (storage stats).

**Direction is right.** The old code compared `index_used_bytes` to `indexes-memory-budget` for all namespaces — meaningless for persisted (flash/pmem) PI. The PR addresses that.

**Main concern — `index-type.mounts-budget` is not a stop-writes trigger in the server.** Per `eval_stop_writes` (nsup.c:942-1021), `ns->stop_writes = true` is set only for:
- `stop-writes-sys-memory-pct` (sys memory)
- `indexes-memory-budget` (where `ixs_sz` explicitly zeros out persisted PI and persisted sindex — see nsup.c:947-955)
- `storage-stop-writes-avail-pct` (data device avail %)
- `storage-stop-writes-used-pct` (data device used %)
- `clock_skew_stop_writes`

`pi_mounts_budget` is consulted only for: HWM eviction via `pi_evict_mounts_pct` (nsup.c:720-724), arena allocation cap (allocation failures log a ticker warning — `arenax_ee.c:444-447` — but don't set `stop_writes`), and reporting (`index_mounts_used_pct`, `index_flash_alloc_pct`). So a flash/pmem namespace at 100% of mounts-budget will not flip the namespace `stop_writes` stat on its own. The new CRITICAL ASSERT (and the new `show stop-writes` entry) will only fire when stop-writes was already triggered for a different reason and mounts happen to be near budget — i.e. misattribution.

Inline comments below on the specific lines.

**Smaller items:**
- The new tests place `"index-type"` under `ns_stats`. `index-type` is a config field (cfg_info.c:545); the test only works because `create_stop_writes_summary` deep-merges stats and config (common.py:1744). Logically belongs under `ns_config`.
- `_is_stop_writes_cause` uses `usage >= threshold`, while the server uses `>` (`ixs_sz > ixs_memory_budget`, nsup.c:982). Pre-existing inconsistency, not introduced here.
- The deprecation messages on `health` don't point users at alternatives (e.g. `show stop-writes`, `show config diff`, `show best-practices`). Worth adding so users have a migration path.
- The new `query.py` mounts ASSERT reuses "namespace size misconfiguration" as the probable cause. For mounts overrun, "index mounts under-provisioned — consider raising `index-type.mounts-budget` or lowering `index-type.evict-mounts-pct`" would be more actionable.

lib/utils/common.py
dwelch-spike: This seems worth addressing but might need some product input about where to put this. @a-spiker if you think moving this somewhere else is out of scope I'm fine with you taking option 1, then making a ticket for a separate view. Let me know what you think.

dwelch-spike: Follow-up design thoughts on where this should live instead, after looking at adjacent asadm surfaces:

**There's already a near-perfect home: `info namespace usage`** (`lib/view/templates.py:253` → `info_namespace_usage_sheet`). It has a "Primary Index" subgroup with Type / Used / Used% / Evict% per node per namespace, and a `yellow_alert` formatter that highlights Used% when it crosses Evict% (templates.py:336-342). The data, granularity, and visual treatment for "eviction in progress" are already there — it's just missing a stronger signal for "near full / critical".

Other places that touch persisted index usage:
- **`summary`** — cluster-level `flash_index` / `pmem_index` totals (templates.py:1490-1503). Good for capacity planning, not runtime alerts.
- **`show statistics namespace`** — raw `index_mounts_used_pct`, `index_flash_alloc_pct`. Available but buried.
- **`show stop-writes`** — where this PR puts it. Wrong place, per the comment above.

### Recommendation

**No new command.** Three surfaces is already on the edge; adding a fourth splits related signals.

1. **Strengthen `info namespace usage` Primary Index Used%** — add a second formatter alongside the existing `yellow_alert`:

   | Condition | Color | Meaning |
   |---|---|---|
   | `Used% < Evict%` | none | Healthy |
   | `Used% >= Evict%` | yellow (existing) | Eviction running |
   | `Used% >= 90%` (or 95%) | **red** (new) | Near `mounts-budget` — arena allocations may start failing |

   Small change at templates.py:336-342: add `Formatters.red_alert(lambda edata: edata.value >= 0.9)`. Same shape for Secondary Index.

2. **Rework the new `lib/health/query.py` ASSERT** into two checks that match server semantics:

   ```
   evicting  = do mounts_pct >= evict_pct;
   ASSERT(evicting, False, "Index mounts in eviction range.", "OPERATIONS", WARNING,
       "Eviction is reducing capacity pressure; verify it can keep pace with insert rate.",
       "Index mounts eviction range check.", is_persisted);

   near_full = do mounts_pct >= 95;
   ASSERT(near_full, False, "Index mounts near capacity.", "OPERATIONS", CRITICAL,
       "Arena allocation failures will begin once mounts-budget is exhausted, causing per-record write errors. Raise index-type.mounts-budget or lower index-type.evict-mounts-pct.",
       "Index mounts near capacity check.", is_persisted);
   ```

   The CRITICAL one fires near the actual failure mode (arena allocation failures), not a non-existent stop-writes trigger. WARNING covers the "eviction kicked in" case that currently has no surface.

3. **Drop the persisted branch in `_format_ns_stop_writes_metrics`** (this file). `show stop-writes` should only show metrics the server can actually use to set `ns->stop_writes`. Gating `indexes-memory-budget` away from persisted-PI namespaces was right; replacing it with `index-type.mounts-budget` wasn't.

### Why this is better

- Honest about server semantics — nothing claims mounts-budget triggers stop-writes when it doesn't.
- Two thresholds match two distinct operational events: WARNING = eviction is doing its job; CRITICAL = eviction is losing the race, write failures imminent. The single CRITICAL in the PR collapses these.
- Uses existing UI — no new command to discover; the existing Used% column already has `mounts-budget` as the denominator.
- Symmetric with how data-on-device is already presented: `info namespace usage` already shows Disk Used%/Avail% with thresholds on the data side.

Happy to sketch the templates.py and query.py edits as a follow-up PR if useful.

dwelch-spike: ^ Claude's recommendation. Lets go over this next week and deliberate.

lib/health/query.py
```diff
@@ -338,14 +338,56 @@
...
+persisted indexes. Instead, eviction starts at index-type.evict-mounts-pct
+(mounts-high-water-pct pre 7.0), and index allocations fail with per-record write
+errors once usage reaches index-type.mounts-budget (mounts-size-limit pre 7.0).
+The mounts usage pct statistics are only reported for flash/pmem index types, so
+these checks are skipped for in-memory index namespaces.
+*/
+SET CONSTRAINT VERSION >= 4.3.0;
+pi_mounts_used_pct = select "index_mounts_used_pct" as "stats", "index_flash_used_pct" as "stats", "index_pmem_used_pct" as "stats" from NAMESPACE.STATISTICS save;
+pi_evict_mounts_pct = select "index-type.evict-mounts-pct" as "stats", "index-type.mounts-high-water-pct" as "stats" from NAMESPACE.CONFIG save;
+pi_evict_configured = do pi_evict_mounts_pct > 0;
```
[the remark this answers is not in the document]
a-spiker: Verified this against the engine. The premise is right: `select` raises when none of the keys exist anywhere (`commands.py:132-135`). But the impact is different from what's described here:

- The exception is per-statement and caught. The dependent statements and the ASSERT are skipped, which is exactly the intended outcome on a cluster with no persisted indexes (no false alerts). Nothing surfaces in default or `-v` output; these only print under `health -d` (`view.py:2144-2145`).
- This is the established pattern of the whole query file, not something new here. On a memory-only cluster, master's query file already produces ~830 internal exceptions of this class (sprig checks select `index-type.mounts-size-limit`/`mounts-budget` unconditionally, the previous sindex check selected `sindex-type.mounts-budget` which is only emitted for persisted sindex, XDR/DC checks on clusters without XDR, etc.). This block adds a net +13 of the same class, and removes 4 by replacing the old sindex check.
- The suggested rework doesn't avoid the exception: `index-type.mounts-budget` and `evict-mounts-pct` are also emitted only for persisted namespaces (`cfg_info.c` gates them on `as_namespace_index_persisted`), so a select on them raises identically on memory-only clusters. The DSL has no construct for a no-op select on an absent-everywhere key; the ASSERT filter arg only gates assert application, not selects. It would also mean recomputing a pct the server already computes for us.

Added `test_memory_only_cluster_skips_mounts_checks` and `test_mixed_cluster_alerts_flash_namespace_only` (604e966) to pin down the skip/isolation behavior: memory-only clusters fire no mounts alerts and execution completes with zero syntax errors, and mixed clusters alert only on the flash namespace.

lib/health/query.py
```diff
@@ -338,14 +338,56 @@
...
+				"Listed namespace[s] have primary index mounts usage at or above 95% of index-type.mounts-budget (mounts-size-limit pre 7.0). Index allocations fail with write errors once the budget is exhausted. Increase the budget or reduce index usage.",
+				"Critical Namespace index mounts used check.");
+
+// Flash arena allocation can run ahead of usage (stages are pre-allocated and
+// freed elements are reused but not returned), so also watch alloc pct.
+pi_flash_alloc_pct = select "index_flash_alloc_pct" as "stats" from NAMESPACE.STATISTICS save;
+pi_alloc_nominal = do pi_flash_alloc_pct < 95;
+ASSERT(pi_alloc_nominal, True, "High namespace index mounts allocation (flash).", "OPERATIONS", WARNING,
+				"Listed namespace[s] have flash primary index arena allocation at or above 95% of index-type.mounts-budget (mounts-size-limit pre 7.0). New arena stage allocations fail once the budget is exhausted and writes then depend on reusing freed index elements. Increase the budget or reduce index usage.",
+				"Namespace index mounts allocation check.");
```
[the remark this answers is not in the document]
a-spiker: Same engine behavior as the thread above: on clusters with no flash-index namespaces this select raises internally, the check is skipped, and nothing appears in default or `-v` output (exceptions print only with `health -d`). That is the intended outcome, and it matches how the rest of the query file handles optional stats.

Extra gating can't help inside the DSL: any gate would itself need a select on flash-only stats or configs, which has the same no-match property. Moving it to Python would lose the health framework's version handling, grouping, and reporting for no functional gain.

Covered by `test_memory_only_cluster_skips_mounts_checks` (604e966), which asserts a cluster without flash indexes fires no mounts or alloc alerts and executes cleanly.

dwelch-spike:
Needs conflicts resolved.

dwelch-spike:
**PR Review (version 2026.07.06)** — re-reviewed at `0b16f99`

**Promises:**
1. `health` is deprecated in live-cluster and collectinfo modes: help text + runtime warning; command still works otherwise.
2. Healthy All-Flash/pmem namespaces produce no false index-related CRITICAL on any supported server version.
3. New health checks match actual server behavior with correct per-version stat/config names and version gates (PI mounts 4.3+, flash alloc 5.6+, sindex mounts 6.3+, combined memory 7.1+).
4. Removing the old 7.0+ sindex mounts-budget CRITICAL loses no coverage.
5. `info namespace usage` PI/SI Used% gets red at >=90%, red wins over yellow, yellow unchanged.
6. `show stop-writes` behavior unchanged.
7. New tests run the real HealthChecker/templates and would catch regressions.

| Gate | Status |
|------|--------|
| Contract | PASS — version gates corrected and boundary-probed (alloc skipped at 5.5/fires at 5.6; sindex skipped at 6.2/fires at 6.3); mechanism wording now matches server source; PR description table corrected |
| Impacted surface | PASS |
| Failure paths | PASS — `on common` on all three cross-source comparisons; original repro (partial evict config) probed at old head (bug reproduces) and new head (WARNING fires for exactly the configured namespace, zero key-mismatch exceptions) |
| Evidence | PASS — triple stat-name mutation now fails exactly the 3 new tests; sindex config-name mutation fails 2; post-mounts positive control pinned to the 7.0+ block by message mutation; yellow-before-red swap caught by the rewritten render-pipeline test; deprecation-warning removal caught; full suite 1489 passed / 1 pre-existing environmental failure (test_ssh, fails identically on master) |
| Quality | PASS |

**Prior findings:**
| Finding | Status |
|---------|--------|
| Data Flow: missing `on common` on cross-source comparisons (query.py) | Addressed in 0b16f99 — probe-verified |
| Architecture: `index_flash_alloc_pct` gated at 4.3 but exists from 5.6 (query.py) | Addressed — `>= 5.6.0`, boundary-probed |
| Architecture: sindex mounts checks gated at 4.3 but persisted sindex is 6.3+ (query.py) | Addressed — `>= 6.3.0`, boundary-probed |
| Architecture: "allocations fail at mounts-budget" misstates server behavior (query.py) | Addressed — reworded; parse-verified (`;` in block comment strips cleanly, zero syntax exceptions) |
| Style: "beta" + "deprecated" contradiction (live controller) | Addressed |
| Control Flow: "(deprecated)" short_msg unreachable under `hide=True` (live controller) | Dismissed — author kept `hide=True` and documented the asymmetry in the PR description (the accept option offered in the finding); description/behavior mismatch resolved |
| Style: red-tier comment wrong for shmem path (templates.py) | Addressed |
| Tests: deprecation had zero test evidence (controllers) | Addressed — 2 new test modules; warning-removal mutation caught |
| Tests: pmem/pre-7.0-sindex name matrix untested (**important**) | Addressed — 3 new tests; triple-name mutation now fails exactly those 3 |
| Tests: memory-only test asserted only absences | Addressed — positive control proven to pin the 7.0+ post-mounts check specifically |
| Complexity: formatter test mirrored the render loop | Addressed — real `sheet.render`; order-swap mutation caught |

**1 new finding** posted inline (minor): the `warnings.filterwarnings("error", ...)` in the new live-cluster controller test installs a process-global filter with no cleanup — safe under pytest (CI), leaks under plain `unittest`.

lib/health/query.py
```diff
@@ -338,14 +338,56 @@
...
+(mounts-high-water-pct pre 7.0), and index allocations fail with per-record write
+errors once usage reaches index-type.mounts-budget (mounts-size-limit pre 7.0).
+The mounts usage pct statistics are only reported for flash/pmem index types, so
+these checks are skipped for in-memory index namespaces.
+*/
+SET CONSTRAINT VERSION >= 4.3.0;
+pi_mounts_used_pct = select "index_mounts_used_pct" as "stats", "index_flash_used_pct" as "stats", "index_pmem_used_pct" as "stats" from NAMESPACE.STATISTICS save;
+pi_evict_mounts_pct = select "index-type.evict-mounts-pct" as "stats", "index-type.mounts-high-water-pct" as "stats" from NAMESPACE.CONFIG save;
+pi_evict_configured = do pi_evict_mounts_pct > 0;
+pi_below_evict = do pi_mounts_used_pct < pi_evict_mounts_pct;
```
dwelch-spike: From Data Flow agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

The three cross-source comparisons — `pi_below_evict` (this line), `sindex_below_evict`, and `ixs_below_evict` — compare a STATISTICS-derived dict against a CONFIG-derived dict without `on common`. If the evict config key is missing from even one namespace while the stat is present (failed info call on one node, partial collectinfo — exactly the degraded-cluster scenarios recent branch work targets), `BinaryOperation._operate_dicts` (lib/health/operation.py:484-493) raises `HealthException("Wrong operands with non-matching keys")`, the variable becomes `None`, and the eviction WARNING is silently skipped for **every** namespace — not just the mismatched one. Reproduced against PR head: two flash namespaces, one in eviction range with evict-mounts-pct configured, the other missing the config key → zero failed asserts; the configured namespace's WARNING is lost (visible only under `health -d`). The promised "skipped when 0/unset" semantics hold only when the key is present-with-0 or absent uniformly. The 95% CRITICAL checks are immune (dict-vs-scalar comparison never key-mismatches).

## Reasoning Chain

1. `select_keys_from_dict` (operation.py:958-1063) includes a namespace in the result only if a requested key matched there, so the stats select and config select can produce different namespace key sets.
2. `_operate_dicts` with `on_common_only=False` raises on any key-set asymmetry instead of operating on the intersection.
3. `p_statement` assigns `None` and records the exception (parser.py:308-330, 527-547); the loop continues (health_checker.py:370-380).
4. The subsequent ASSERT gets `data=None` → "Wrong Input Data for ASSERT operation." (operation.py:834-835) → no assert output at all.
5. Probe on PR head confirmed: the configured, in-eviction-range namespace produced no WARNING.

## Solution

Append `on common` to the three comparisons, matching the file's existing pattern for cross-source comparisons (e.g. `r = do oc == c on common`, query.py:264):

```
pi_below_evict = do pi_mounts_used_pct < pi_evict_mounts_pct on common;
```

and likewise for `sindex_below_evict` and `ixs_below_evict`. With `on common`, namespaces missing the config key are individually dropped (true "unset → skipped for that namespace") while configured namespaces keep their WARNING. In the aligned-keys normal case the intersection equals the full key set — no behavior change.

lib/health/query.py
```diff
@@ -338,14 +338,56 @@
...
+				"Namespace index mounts eviction range check.", pi_evict_configured);
+
+pi_mounts_nominal = do pi_mounts_used_pct < 95;
+ASSERT(pi_mounts_nominal, True, "High namespace index mounts usage (near capacity).", "OPERATIONS", CRITICAL,
+				"Listed namespace[s] have primary index mounts usage at or above 95% of index-type.mounts-budget (mounts-size-limit pre 7.0). Index allocations fail with write errors once the budget is exhausted. Increase the budget or reduce index usage.",
+				"Critical Namespace index mounts used check.");
+
+// Flash arena allocation can run ahead of usage (stages are pre-allocated and
+// freed elements are reused but not returned), so also watch alloc pct.
+pi_flash_alloc_pct = select "index_flash_alloc_pct" as "stats" from NAMESPACE.STATISTICS save;
```
dwelch-spike, line 378: From Architecture agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

`index_flash_alloc_pct` does not exist before server 5.6, but this check sits inside the `SET CONSTRAINT VERSION >= 4.3.0` block and the PR description's table claims the stat applies to all pre-7.0 versions. On 4.3–5.5 all-flash clusters the select matches nothing, the DSL raises `HealthException`, and the alloc check silently vanishes (visible only via `health -d`). No false alert, but the stated compatibility matrix is wrong and the constraint misdocuments where the check actually runs.

## Reasoning Chain

1. `git grep index_flash_alloc_pct` at server tags 4.3.0.10, 4.9.0.36, 5.0.0.11, 5.1.0.10, 5.2.0.10, 5.3.0.6, 5.4.0.5, 5.5.0.9 — absent from `as/src/base/thr_info.c` at every one.
2. First present at 5.6.0.1 (`thr_info.c:5493`), and continuously through 5.7.0.10, 6.4.0.1, 7.0.0.1, 8.1.1.1.
3. In asadm, an unmatched select raises `HealthException` (lib/health/commands.py:132-135); exceptions print only under `-d` (lib/view/view.py:2159-2161), so the check is skipped without trace in normal output.

## Solution

Move the `pi_flash_alloc_pct` select and its ASSERT under `SET CONSTRAINT VERSION >= 5.6`, and reorder or re-set the constraint for the sindex section that follows (see the separate comment there — it needs `>= 6.3` anyway). Also correct the PR description's version table.

lib/health/query.py
```diff
@@ -338,14 +338,56 @@
...
+pi_alloc_nominal = do pi_flash_alloc_pct < 95;
+ASSERT(pi_alloc_nominal, True, "High namespace index mounts allocation (flash).", "OPERATIONS", WARNING,
+				"Listed namespace[s] have flash primary index arena allocation at or above 95% of index-type.mounts-budget (mounts-size-limit pre 7.0). New arena stage allocations fail once the budget is exhausted and writes then depend on reusing freed index elements. Increase the budget or reduce index usage.",
+				"Namespace index mounts allocation check.");
+
+sindex_mounts_used_pct = select "sindex_mounts_used_pct" as "stats", "sindex_flash_used_pct" as "stats", "sindex_pmem_used_pct" as "stats" from NAMESPACE.STATISTICS save;
```
dwelch-spike, line 385: From Architecture agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

Persisted sindex (`sindex-type` flash/pmem) does not exist on any server before 6.3, and flash sindex specifically not before 6.4 — yet the sindex mounts checks sit under `SET CONSTRAINT VERSION >= 4.3.0`, and the PR description claims pre-7.0 coverage from 4.3. On 4.3–6.2 servers the selects match nothing and the checks silently skip (HealthException, debug-only). Behaviorally harmless — no false CRITICAL — but the constraint is two major versions too loose and the description asserts compatibility that doesn't exist.

## Reasoning Chain

1. `git grep "sindex-type"` at server tags 6.1.0.20 and 6.2.0.20 across `as/` — zero hits; the concept does not exist there.
2. 6.3.0.1: `cfg_info.c:458-470` — `sindex-type` values are `mem|shmem|pmem` only (no flash); emits `sindex-type.mounts-high-water-pct` / `sindex-type.mounts-size-limit` when persisted; `thr_info.c:4262` emits `sindex_pmem_used_pct` only (confirmed through 6.3.0.36 — no `sindex_flash_used_pct`).
3. 6.4.0.1: `thr_info.c:4186` adds `sindex_flash_used_pct`; same config names. 7.0.0.1 renames to `sindex_mounts_used_pct` / `sindex-type.evict-mounts-pct` / `sindex-type.mounts-budget` as the PR claims.
4. Under `>= 4.3.0`, 4.3–6.2 nodes produce empty selects → `HealthException` (lib/health/commands.py:132-135), debug-only.

## Solution

Give the sindex mounts checks their own `SET CONSTRAINT VERSION >= 6.3` block (pmem from 6.3, flash from 6.4 — the alias merge already handles both name sets). Fix the PR description's version table accordingly.

lib/health/query.py
```diff
@@ -338,14 +338,56 @@
...
+ASSERT(ixs_below_evict, True, "High namespace indexes memory usage (eviction range).", "OPERATIONS", WARNING,
+				"Listed namespace[s] have combined indexes memory usage at or above evict-indexes-memory-pct, so eviction should be in progress. Verify eviction is keeping pace with the write load, or increase indexes-memory-budget.",
+				"Namespace indexes memory eviction range check.", ixs_evict_configured);
+
+/*
+Persisted (flash/pmem) index mounts usage checks.
+The server never triggers stop-writes on mounts usage - eval_stop_writes() excludes
+persisted indexes. Instead, eviction starts at index-type.evict-mounts-pct
+(mounts-high-water-pct pre 7.0), and index allocations fail with per-record write
+errors once usage reaches index-type.mounts-budget (mounts-size-limit pre 7.0).
```
dwelch-spike: From Architecture agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

The comment block and the CRITICAL/WARNING message strings claim "index allocations fail with per-record write errors once usage reaches index-type.mounts-budget (mounts-size-limit pre 7.0)". On every version traced (6.4, 7.0, 8.1 — CE and EE), the budget is never enforced in the allocation path: it is validated once at startup (must not exceed total mounts filesystem size) and thereafter serves only as the denominator for the pct stats and the nsup eviction threshold. Arena stage allocation fails on actual filesystem exhaustion (`posix_fallocate` ENOSPC) or the max-stage cap — not at the configured budget. When the budget is set below the mounts' real capacity, usage can pass 95% and even 100% of budget with zero write errors while this CRITICAL asserts failures are imminent. The alert threshold itself is sound capacity hygiene; the stated mechanism is not server behavior.

## Reasoning Chain

1. All uses of `pi_mounts_size_limit` at 6.4.0.1 (CE): cfg parse (`cfg.c:3100/3127`), get-config emission (`cfg_info.c:472`), nsup hwm math (`nsup.c:640`), stat/ticker denominators (`thr_info.c:4144-4159`, `ticker.c:518-528`). No allocation-path use.
2. Same for `pi_mounts_budget` at 7.0.0.1 and 8.1.1.1.
3. EE: budget flows only into `cf_xmem_type_cfg_init`; the flash implementation uses it solely for the startup sanity check `if (budget > total) ... return false` (`cf/src/xmem_flash.c:149-153` at 6.4.0.1 and 8.1.1.1), then builds an `ssd_cfg` containing only the mounts — the budget is discarded.
4. Actual failure path: `cf_arenax_add_stage` (`cf/src/arenax_ee.c` at 7.0.0.1 and 8.1.1.1) fails on `CF_ARENAX_MAX_STAGES` or `cf_xmem_create_block` error (`posix_fallocate`, `xmem_flash.c`) — i.e. real device exhaustion.
5. The single check that would settle it empirically: flash namespace with `mounts-budget` at 50% of mount FS capacity, load past 100% of budget, observe writes still succeed.

## Solution

Keep the thresholds; reword the comment block and the three ASSERT strings so the mechanism matches the server: eviction is budget-based (`evict-mounts-pct` of budget, `nsup.c:685-691` at 7.0.0.1); allocation failures occur when the mounts themselves fill, which coincides with the budget only when the budget reflects actual mount capacity (the recommended configuration). E.g.: "usage is at or above 95% of the configured budget; if the budget reflects actual mount capacity, index allocations will begin failing with write errors."

lib/live_cluster/health_check_controller.py
```diff
@@ -34,7 +34,8 @@
 
 @CommandHelp(
     "Displays health summary. If remote server System credentials provided, then it will collect remote system stats and analyze that also. If credentials are not available then it will collect only localhost system statistics. This command is still in beta and its output should not be directly acted upon without further analysis.",
-    short_msg="Displays health summary",
+    "Note: This command is deprecated and is no longer actively maintained. It will be removed in a future release.",
```
dwelch-spike, line 37: From Style agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

The live-cluster `help health` output now tells the user the command "is still in beta and its output should not be directly acted upon without further analysis" AND, in the very next sentence, that it "is deprecated and ... will be removed in a future release." A command cannot coherently be both in beta (implying it will mature) and deprecated (implying it will be removed).

## Reasoning Chain

`CommandHelp.__init__` joins all positional `long_msg` strings with a single space (lib/base_controller.py:67) into one wrapped paragraph, so the rendered text says "beta" then "deprecated" back to back. The collectinfo controller's long message was rewritten in this same PR ("Displays health summary based on collectinfo data.") — the live-cluster one got no equivalent cleanup.

## Solution

Trim the live-cluster first message to match the collectinfo treatment — drop "This command is still in beta and its output should not be directly acted upon without further analysis." (the deprecation note supersedes that caveat), or fold the caution into the deprecation note.

lib/live_cluster/health_check_controller.py
```diff
@@ -34,7 +34,8 @@
...
+    short_msg="Displays health summary (deprecated)",
```
dwelch-spike, line 38: From Control Flow agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

In live-cluster mode the new "(deprecated)" short_msg is dead text — there is no execution path that renders it.

## Reasoning Chain

This controller is decorated with `hide=True` (pre-existing, line 64 at PR head). Both consumers of `short_message` (lib/base_controller.py): (1) the parent command listing at line 608 is gated by `not CommandHelp.is_hidden(method)`, so `health` never appears in the live-cluster `help` listing; (2) the "Default" row at line 601 is inside `if self.commands:` (line 590), and this controller has no `do_*` subcommands, so that row never prints even on explicit `help health`. Only the long-message note is reachable. The collectinfo controller is not hidden, so its "(deprecated)" marker does render there — the asymmetry is live-cluster only. Flagging because the PR description explicitly promises the marker in short help.

## Solution

Either accept (hidden-command short help is inherently invisible; the runtime warning and long-message note still cover deprecation) and note it in the PR description, or remove `hide=True` so the deprecated command and its "(deprecated)" marker are visible in the command list until removal. Solution confidence medium — hiding arguably serves deprecation better; your call which is intended.

lib/view/templates.py
```diff
@@ -369,6 +369,9 @@ def usage_weighted_avg(edatas: list[EntryData]):
                         converter=Converters.ratio_to_pct,
                     ),
                     formatters=(
+                        # Near the mounts/memory budget - index allocations
```
dwelch-spike: From Style agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

The comment justifying the red tier — "index allocations fail once the budget is exhausted" — is not accurate for either path this field can take. For the in-memory `indexes-memory-budget` path (shmem, 7.1+), the consequence of budget exhaustion is stop-writes, not allocation failure. For the flash/pmem mounts path, the budget is not an allocation cap either — allocations fail when the mounts filesystem actually fills, which coincides with the budget only when the budget matches mount capacity (see the corresponding Architecture comment on lib/health/query.py).

## Reasoning Chain

This Used% field is fed by three projectors (templates.py:334-365): `index_mounts_used_pct` (flash/pmem 7.0+), pre-7.0 mounts bytes/size-limit, and `index_used_bytes / indexes-memory-budget` (shmem 7.1+). For the shmem path, the server's `eval_stop_writes()` is what enforces the memory budget — this PR's own `test_indexes_memory_over_budget_is_critical` asserts the "stop-write enabled" CRITICAL for exactly that case. Corroborated independently by the server-source trace in the query.py comment thread. Same text at the Secondary Index copy (~line 460).

## Solution

Reword both copies to cover the real mechanisms, e.g.: "Near the mounts/memory budget — stop-writes for in-memory indexes; for flash/pmem, allocations start failing as the mounts fill." Keep the 90% threshold.

lib/live_cluster/health_check_controller.py
```diff
@@ -127,6 +128,11 @@ async def _get_as_meta_data(self, stanza):
             return await self.cluster.info_health_outliers(nodes=self.nodes)
 
     async def _do_default(self, line):
+        logger.warning(
```
dwelch-spike, line 131: From Tests agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

Contract item "health is deprecated in both modes (help + runtime warning), still functional" has zero test evidence. No test in the PR or the existing suite (grep for "deprecat" in test/ finds only unrelated LDT/sindex cases) asserts that either controller emits the deprecation warning or that the command still functions after the change.

## Reasoning Chain

The warning is emitted at the top of `_do_default` in both controllers, before argument parsing. A future refactor of `_do_default` (these controllers parse half a dozen args inline and are prime refactoring targets) can drop or reorder the warning with no failing test. "Still functional" is likewise asserted only by the PR description.

## Solution

In the existing controller test modules, wrap an invocation of each controller's `_do_default` in `self.assertLogs(logger, level="WARNING")` and assert the message contains "deprecated". The same test doubles as the "still functional" check since `_do_default` must run to completion to emit its normal path.

test/unit/health/test_healthchecker.py
```diff
@@ -15,9 +15,288 @@
...
+                    "index-type.mounts-budget": 100 * 1024**3,
+                    "index-type.evict-mounts-pct": 80,
+                }
+            },
+        )
+
+        self.assertEqual(failed.get(self.PI_EVICTION), AssertLevel.WARNING)
+        self.assertEqual(failed.get(self.PI_CAPACITY), AssertLevel.CRITICAL)
+
+    def test_pre_7_0_flash_index_near_capacity_is_critical(self):
```
dwelch-spike, line 152: From Tests agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

Half of the per-version stat-name matrix has zero test evidence. Mutating the production query file to break `index_pmem_used_pct`, `sindex_flash_used_pct`, and `sindex_pmem_used_pct` simultaneously (`_BROKEN` suffixes) left the entire suite green (11 passed, 1 skipped). The same holds for the pre-7.0 sindex config names `sindex-type.mounts-high-water-pct` / `sindex-type.mounts-size-limit`, which no test selects.

## Reasoning Chain

This test covers only the PI flash pre-7.0 path; `test_flash_sindex_near_capacity_is_critical` runs on 8.1.1.1 and selects `sindex_mounts_used_pct`. A select with a misspelled alternative name does not error — it simply matches nothing, so the check silently never fires. That is exactly the failure mode these tests exist to prevent (a health check that quietly evaluates nothing), and it is the failure mode TOOLS-3456 itself was: wrong stat/config pairing. A future regression in any of these names ships undetected, and pmem-index or pre-7.0 persisted-sindex clusters lose their CRITICAL capacity check with no signal.

## Solution

Add two cases mirroring this test: (1) version "6.4.0.1" with `index_pmem_used_pct: 96` + `index-type.mounts-size-limit` / `mounts-high-water-pct`, asserting PI_EVICTION=WARNING and PI_CAPACITY=CRITICAL; (2) version "6.4.0.1" with `sindex_flash_used_pct: 97` (or the pmem variant) + `sindex-type.mounts-size-limit` / `sindex-type.mounts-high-water-pct`, asserting SI_EVICTION and SI_CAPACITY. Two entries reusing `failed_asserts` close the matrix. (Note: if the sindex block's version constraint is tightened to >= 6.3 per the Architecture comment on query.py, these tests also pin that constraint.)

test/unit/health/test_healthchecker.py
```diff
@@ -15,9 +15,288 @@
...
+                    "index-type.mounts-budget": 100 * 1024**3,
+                    "index-type.evict-mounts-pct": 0,
+                }
+            },
+        )
+
+        self.assertNotIn(self.PI_EVICTION, failed)
+        self.assertEqual(failed.get(self.PI_CAPACITY), AssertLevel.CRITICAL)
+
+    def test_memory_only_cluster_skips_mounts_checks(self):
```
dwelch-spike, line 231: From Tests agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

This test asserts only absences (five `assertNotIn` calls), so it passes vacuously under any regression that silences checks wholesale — including the specific hazard its own docstring describes: the mounts selects raising internally on memory-only input. Nothing proves that checks positioned *after* the new mounts block still execute in the same run.

## Reasoning Chain

The new block sits mid-file (query.py:348-390) and its selects are guaranteed to raise `HealthException` on memory-only input. Today `_execute_queries` (health_checker.py:356-391) catches per query and continues — verified by reading it — but that continuation property is exactly what this scenario depends on, and no assertion pins it. If a future engine change made a `HealthException` abort or taint the remaining queries, every post-block check (the >= 7.0 storage stop-writes CRITICALs, HWM checks) would silently vanish for memory-only clusters and this suite would stay green: this test asserts absences, and the other positive tests assert checks that appear *before* the mounts block (the IXS checks at query.py:333-346).

## Solution

Add a positive control to this test's input that trips a check located after the mounts block — e.g. `"data_used_pct": 75` in ns_stats with the matching stop-writes config in ns_config — then assert that check IS in `failed` alongside the five absence assertions. One fixture line and one assertEqual convert five vacuous-capable assertions into a real proof.

test/unit/view/test_templates.py
```diff
@@ -89,6 +89,51 @@ def test_format_license_latest_with_compression(
...
+        for field in templates.info_namespace_usage_sheet.fields:
+            if isinstance(field, Subgroup) and field.title == subgroup_title:
+                for sub_field in field.fields:
+                    if sub_field.title == "Used%":
+                        return sub_field
+        raise AssertionError(f"Used% field not found in {subgroup_title}")
+
+    @staticmethod
+    def _applied_formatter(field, edata):
+        # Mirrors sheet behavior: first formatter to not return None wins.
```
dwelch-spike: From Complexity agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

`_applied_formatter` re-implements the sheet renderer's formatter-precedence loop (first formatter to return non-None wins) instead of exercising the real one. The contract claim "red wins over yellow" is therefore proven against a 5-line test-local mirror of `BaseRSheet` (lib/view/sheet/render/base_rsheet.py:860-866), not against the code that ships.

## Reasoning Chain

1. The real precedence lives in `base_rsheet.py:860`, which also short-circuits on `edata.is_error` first.
2. The repo already has a pattern for testing formatter tiers through the real pipeline: `test_sheet_formatters` (test/unit/view/test_sheet.py:800) builds a sheet, calls `do_render`, and asserts `record["F"]["format"] == "red-alert"` — precedence included; `ShowPmapSheetTest` in this same file renders a real template end-to-end.
3. If the render loop's semantics ever change (formatter stacking, severity ordering), this mirror silently keeps testing the old assumption while staying green.
4. Mitigating: template-tuple ordering regressions ARE caught here, and machinery-precedence regressions are caught generically by test_sheet_formatters — coverage is compositional rather than absent. Hence minor.

## Solution

Either render `info_namespace_usage_sheet` through `sheet.render` with minimal namespace sources and assert on the emitted `"format"` value (the ShowPmapSheetTest pattern), or replace the hand-rolled loop with a call into the render helper the machinery exposes. If neither is practical, tighten the comment to reference base_rsheet.py:860 so a future change to that loop has a breadcrumb back to this mirror.

lib/health/query.py
dwelch-spike: Resolved at 9de05ca — the current revision implements the recommended redesign: the mounts checks are now an eviction-range WARNING (gated on evict pct being configured) plus a near-capacity CRITICAL at 95%, worded around capacity/allocation rather than stop-writes, and no mounts ASSERT claims a stop-writes trigger anymore. One refinement to the new wording is tracked in a fresh Architecture comment on this file: server source shows the budget is a startup sanity check + eviction denominator rather than a hard allocation cap, so the "allocations fail at budget" phrasing should be softened to "as the mounts fill / if budget reflects mount capacity".

lib/health/query.py
dwelch-spike: Resolved at 9de05ca — the `is_not_persisted` gate is gone. The 7.1+ check on master (TOOLS-3429) now uses the server-computed `indexes_memory_used_pct`, which per nsup.c includes set-index and non-persisted sindex memory for persisted-PI namespaces, so the set-index/memory-sindex case this thread raised is covered for all namespaces. This PR adds a complementary eviction-range WARNING on the same stat under the same `>= 7.1.0` constraint.

lib/utils/common.py
dwelch-spike: Resolved at 9de05ca — the PR no longer touches lib/utils/common.py at all (zero delta vs master), so `show stop-writes` keeps master's TOOLS-3429 behavior with no mounts-budget entry. The capacity signal moved to the surfaces recommended in this thread: the red tier on `info namespace usage` Used% and the reworked health checks (eviction WARNING + 95% CRITICAL).

lib/health/query.py
a-spiker: Fixed in 0b16f99. All three cross-source comparisons (`ixs_below_evict`, `pi_below_evict`, `sindex_below_evict`) now use `on common`, matching the file's existing pattern. Added `test_partial_evict_config_still_warns_configured_namespace` pinning the exact repro from this comment: two flash namespaces, one in eviction range with `evict-mounts-pct` configured, the other missing the config key. The configured namespace keeps its WARNING and the assert keys contain only that namespace.

lib/health/query.py
a-spiker: Fixed in 0b16f99. The `pi_flash_alloc_pct` select and its ASSERT moved under `SET CONSTRAINT VERSION >= 5.6.0`. Re-verified against server tags: `index_flash_alloc_pct` is absent at 4.3.0.10 through 5.5.0.27 and present from 5.6.0.25. PR description's version table corrected as well.

lib/health/query.py
a-spiker: Fixed in 0b16f99. The sindex mounts checks now sit under their own `SET CONSTRAINT VERSION >= 6.3.0` block. Added tests pinning the per-version names: `sindex_flash_used_pct` on 6.4.0.1 and `sindex_pmem_used_pct` on 6.3.0.1, both with the pre-7.0 `sindex-type.mounts-size-limit` / `sindex-type.mounts-high-water-pct` config names. PR description table corrected.

lib/health/query.py
a-spiker: Fixed in 0b16f99. The comment block and all three ASSERT messages are reworded: the budget is described as seeding the eviction threshold and the pct statistics, and allocation failures are tied to the mounts actually filling ("If the budget reflects actual mount capacity, index allocations will begin failing with write errors as the mounts fill"). Confirmed on CE source at 8.1.1.2: `pi_mounts_budget` appears only in cfg parse, the nsup eviction pct calculation, and stat denominators.

lib/live_cluster/health_check_controller.py
a-spiker: Fixed in 0b16f99. Dropped the beta sentence from the live-cluster long message; the deprecation note supersedes it, matching the collectinfo treatment.

lib/live_cluster/health_check_controller.py
a-spiker: Keeping `hide=True`: it is pre-existing master behavior (the command was already hidden from the live-cluster help listing before this PR), and hiding arguably serves deprecation better than re-surfacing the command. The short_msg marker still renders for the collectinfo controller, which is not hidden. Noted in the PR description.

lib/view/templates.py
a-spiker: Fixed in 0b16f99. Both copies now read "Near the mounts/memory budget: stop-writes for in-memory indexes; for flash/pmem, allocations fail as the mounts fill." The 90% threshold is unchanged.

lib/live_cluster/health_check_controller.py
a-spiker: Added in 0b16f99: `test/unit/live_cluster/test_health_controller.py` and `test/unit/collectinfo_analyzer/test_health_controller.py`. Each runs the controller's `_do_default` to completion against mocked cluster/log-handler inside `assertLogs`, asserting the warning contains "deprecated" and that `health_checker.execute` and `view.print_health_output` are still reached (the still-functional half of the contract).

test/unit/health/test_healthchecker.py
a-spiker: Fixed in 0b16f99. Added three tests closing the matrix: pmem PI on 6.4.0.1 (`index_pmem_used_pct` + `index-type.mounts-size-limit`/`mounts-high-water-pct`), flash sindex on 6.4.0.1 (`sindex_flash_used_pct` + the pre-7.0 sindex config names), and pmem sindex on 6.3.0.1 (`sindex_pmem_used_pct`, which also pins the new >= 6.3 constraint from the Architecture comment). Re-ran the triple-name mutation from this comment plus the two sindex config names: 3 tests now fail, so a name regression no longer ships silently.

test/unit/health/test_healthchecker.py
a-spiker: Fixed in 0b16f99. The memory-only test input now includes `data_used_pct: 75` with `storage-engine.stop-writes-used-pct: 70` and asserts that post-mounts-block CRITICAL fires alongside the five absence assertions, proving checks after the mounts block still execute in the same run.

test/unit/view/test_templates.py
a-spiker: Fixed in 0b16f99. Replaced the hand-rolled precedence loop with a real `sheet.render` of `info_namespace_usage_sheet` (the ShowPmapSheetTest pattern): single flash namespace source, JSON style, asserting on the emitted `format` value for both Primary and Secondary Index Used%. Formatter precedence is now exercised through the actual render loop in base_rsheet.py.

test/unit/live_cluster/test_health_controller.py
```diff
@@ -0,0 +1,68 @@
...
+    "info_get_originalconfig",
+    "info",
+    "info_health_outliers",
+    "info_system_statistics",
+)
+
+
+class HealthCheckControllerDeprecationTest(unittest.IsolatedAsyncioTestCase):
+    async def asyncSetUp(self) -> None:
+        warnings.filterwarnings("error", category=RuntimeWarning)
```
dwelch-spike, line 47: From Tests agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

`warnings.filterwarnings("error", category=RuntimeWarning)` in `asyncSetUp` installs a process-global filter with no cleanup (no `addCleanup`, no `catch_warnings` context).

## Reasoning Chain

Verified empirically with a probe test inspecting `warnings.filters` after this module runs: under pytest — including CI's exact `pytest --disable-warnings test/unit` invocation — pytest's warnings plugin wraps each test item in `catch_warnings`, so there is no cross-test effect today (full suite green). But under a plain `python -m unittest` run, the probe fails: `('error', None, RuntimeWarning, None, 0)` remains installed for the rest of the process, so any later test that legitimately emits a RuntimeWarning starts erroring. The suite's correctness silently depends on which runner is used.

## Solution

Scope the filter to the test:

```python
cm = warnings.catch_warnings()
cm.__enter__()
self.addCleanup(cm.__exit__, None, None, None)
warnings.filterwarnings("error", category=RuntimeWarning)
```

or, since the repo is Python 3.12, simply `self.enterContext(warnings.catch_warnings())` before the `filterwarnings` call.

lib/health/query.py
dwelch-spike: Verified at 0b16f99 — all three comparisons carry `on common`, and the original repro was probed against the real HealthChecker at both heads: at 9de05ca the WARNING is lost with the `Wrong operands with non-matching keys` signature; at 0b16f99 it fires with assert keys containing exactly the configured namespace, zero key-mismatch exceptions. The new `test_partial_evict_config_still_warns_configured_namespace` pins it. Also traced the interaction with the trailing condition arg: the intersection key set is a subset of the config filter's key set, so nothing is lost. Resolving.

lib/health/query.py
dwelch-spike: Verified at 0b16f99 — `SET CONSTRAINT VERSION >= 5.6.0` now precedes the alloc select, and the gate was boundary-probed with the real checker: skipped at 5.5.0.1, fires at 5.6.0.1, and the PI mounts CRITICAL still fires at 4.3.0.1. PR description table corrected. Resolving.

lib/health/query.py
dwelch-spike: Verified at 0b16f99 — sindex mounts checks sit under `SET CONSTRAINT VERSION >= 6.3.0`, boundary-probed: skipped at 6.2.0.1, fires at 6.3.0.1; the following `>= 7.0.0` block is intact (storage stop-writes fires at 7.0.0.1, skipped at 6.4.0.1). New per-version name tests pin flash-on-6.4 and pmem-on-6.3. Resolving.

lib/health/query.py
dwelch-spike: Verified at 0b16f99 — comment block and all three ASSERT strings reworded to the budget-seeds-eviction-and-stats framing with allocation failures tied to the mounts actually filling. Parse safety confirmed: the semicolons sit inside the `/* */` block, which parse_commands strips line-by-line before the `;` statement split; executing the full QUERIES at the new head yields zero syntax exceptions and all three asserts fire in probes. Resolving.

lib/live_cluster/health_check_controller.py
dwelch-spike: Verified at 0b16f99 — the beta sentence is gone from the live-cluster long message; rendered `CommandHelp.long_message` reads coherently in both modes. Resolving.

lib/live_cluster/health_check_controller.py
dwelch-spike: Accepting the dismissal — keeping `hide=True` was the first option this finding offered, and the PR description now documents the asymmetry (short-help tag renders only in collectinfo mode; live-cluster users get the long-help note and runtime warning), which fixes the description/behavior mismatch that was the actual defect. Resolving.

lib/view/templates.py
dwelch-spike: Verified at 0b16f99 — both comment copies now name the real mechanisms (stop-writes for in-memory, allocations fail as mounts fill for flash/pmem); the formatter tuples themselves are unchanged (red first, same thresholds). Resolving.

lib/live_cluster/health_check_controller.py
dwelch-spike: Verified at 0b16f99 — both new controller test modules run `_do_default` to completion inside `assertLogs` and assert `execute`/`print_health_output` are reached; every mocked attribute was checked to exist on the real classes, and removing the `logger.warning` call makes the live test fail. One minor follow-up posted on the new live-cluster test file (unscoped `warnings.filterwarnings`). Resolving.

test/unit/health/test_healthchecker.py
dwelch-spike: Verified at 0b16f99 — re-ran the exact triple-name mutation from this comment (`index_pmem_used_pct`, `sindex_flash_used_pct`, `sindex_pmem_used_pct` broken simultaneously): exactly the 3 new tests fail. Breaking `sindex-type.mounts-high-water-pct` fails 2 more. (Note: `sindex-type.mounts-size-limit` is intentionally not selected anywhere in query.py — the capacity check keys off the used-pct stat — so the tests passing it as fixture input is realism, not coverage.) The matrix is closed. Resolving.

test/unit/health/test_healthchecker.py
dwelch-spike: Verified at 0b16f99 — the positive control is genuine: mutating only the 7.0+ block's message string (query.py:402) makes `test_memory_only_cluster_skips_mounts_checks` fail, proving the assertion is satisfied by the post-mounts-block check specifically and not the similarly-worded pre-7.0 block. Resolving.

test/unit/view/test_templates.py
dwelch-spike: Verified at 0b16f99 — the test now renders `info_namespace_usage_sheet` through the real `sheet.render` (JSON style) and asserts the emitted `format`; swapping yellow before red in templates.py makes the (92, 80, red-alert) case fail, so precedence is exercised through the actual render loop. `import json` and the remaining `EntryData` usage were already present (no F401). Resolving.

test/unit/live_cluster/test_health_controller.py
a-spiker: Done

dwelch-spike (approved):
LGTM

PR #477, opened 2026-05-10
feat: TOOLS-3482 projection and value filtering for show jobs

a-spiker:
## Summary

Adds projection, filtering, and transpose support to `show jobs queries|scans|sindex-builder` (live cluster **and** collectinfo), plus a rendering fix for wide transposed tables.

### New modifiers

- **`--flip`** (also accepts `-flip`) — transpose the table: **jobs on the Y axis (rows), fields on the X axis (columns)**. The default view is the inverse (fields as rows, jobs as columns).
- **`like <field-substring>`** — project to fields whose names match.
- **`for <ns> [<set>]`** — filter by namespace, and optionally set.
- **`--where <field>=<regex>`** (also accepts `-where`) — value filter on any job field. Repeatable; multiple clauses **AND** together.

### `--where` semantics

- **`<field>`** accepts either the **raw server key** (`ns`, `set`, `job-type`, `trid`, `status`, …) or the **display column header** (`Namespace`, `Set`, `Type`, `Transaction ID`, `Time Since Done`, …), matched **case-insensitively**.
- The RHS is a **regex** (via `util.compile_likes`): plain text matches as a substring; metacharacters (`^ $ . *` …) are supported.
- An **unknown / misspelled field** raises a `ShellException` listing the known fields — instead of silently returning an empty table (you can tell a typo from a genuine zero-match).
- A **malformed clause** (`--where status`, bare `--where`) or an **invalid regex** (`--where "status=("`) raises a friendly `ShellException` naming the offending clause — no stack trace.

### Implementation notes

- `for` / `--where` filtering is applied inside `GetJobsController`, mirroring the existing `show config namespace` / `get_namespace(for_mods=…)` pattern. New helpers `filter_jobs` and `parse_jobs_mods` live in `lib/live_cluster/get_controller.py` and are shared by the live and collectinfo controllers.
- The display-header → raw-key alias map is keyed off the `show_jobs` render template, so it stays in sync with the columns users actually see.
- Modifier help/usage strings are defined **once** in `get_controller.py` and imported by both `show_controller.py` files (previously duplicated byte-for-byte across the two controllers).
- `--flip` selects the columns sheet style in `view.show_jobs` and threads a new `disable_title_fill_repeat` flag through `sheet.render` → `BaseRSheet`. This fixes a rendering bug where a flipped table wider than the terminal repeated the title bar (`~~Query Jobs (ts)~~Query Jobs (ts)~~`); the title now renders **once** at any terminal width.

### Flag-style consistency

`--flip` and `--where` use the `--long` convention shared by every other long-form flag in the CLI (`--by-dc`, `--no-warn`, `--enable-ssh`, …), while `for` / `like` / `with` / `trid` remain barewords (the shared modifier vocabulary used across all `show`/`info` commands). For leniency, `--where`/`-where` and `--flip`/`-flip` are both accepted.

## Examples

```
# Transpose: jobs become rows, fields become columns
show jobs queries --flip

# Project to a handful of field columns
show jobs queries --flip like "^(Node|Namespace|Set|Type|status|Progress%)$"

# Filter by namespace, optionally set
show jobs queries for test
show jobs queries for test testset

# Keep only active jobs (raw key)
show jobs queries --where status=active

# --where accepts display headers too (case-insensitive)
show jobs queries --where Namespace=test
show jobs queries --where "Transaction ID=1234"

# Regex metacharacters work
show jobs queries --where "status=^done"

# Multiple --where clauses AND together
show jobs queries --where ns=test --where status=active

# Combine modifiers
show jobs queries for test --where status=active --flip

# Friendly errors (no stack trace)
show jobs queries --where status        # ShellException: invalid --where clause: 'status' — expected <field>=<pattern>
show jobs queries --where "status=("    # ShellException: invalid --where regex in 'status=(': ...
show jobs queries --where bogus=foo     # ShellException: unknown --where field 'bogus'. Known fields: [...]
```

## Test plan

- [x] `pytest test/unit` — **1353 passed, 1 skipped** at the base, **1365 passed, 1 skipped (140 subtests)** with the review-feedback changes
- [x] `GetJobsControllerTest` — **15 tests**: `for`/`for`+set/`where` filtering, display-name aliases (`Namespace`, `Transaction ID`), unknown-field `ShellException`, empty-data (no spurious error), regex-anchor enforcement, empty-host preservation, invalid-regex
- [x] `ShowJobsControllerTest` (live_cluster) — **21 tests**: `--flip`/`-flip`, `like`, `for`→getter, `where`→getter, `-where`/`--where`/mixed-order, multiple-`where`, malformed-`where`, trailing-`where`
- [x] `ShowJobsController` (collectinfo) — **6 tests**: missing-module/None case, `where`, `flip`, `for`-ns-only, malformed-`where`, trailing-`where`
- [x] `CliViewTest` — `test_show_jobs`, `test_show_jobs_flip` (flip → `SheetStyle.columns` + `disable_title_fill_repeat`), `test_show_jobs_flip_actual_output_title_once` (real-stdout guard asserting the title renders exactly once under a narrow terminal)
- [x] Manual verification against a live 8.1.x cluster and a collectinfo capture: all four modifiers, display-header aliases, `--flip` title-once on narrow (80-col) and wide terminals, and every error path

## Jira

[TOOLS-3482](https://aerospike.atlassian.net/browse/TOOLS-3482)

[TOOLS-3482]: https://aerospike.atlassian.net/browse/TOOLS-3482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/view/view.py
```diff
@@ -1374,10 +1375,16 @@ def show_jobs(
         node_ids = cluster.get_node_ids(with_)
         sources = dict(data=filtered_data, node_names=node_names, node_ids=node_ids)
         common = dict(principal=cluster.get_expected_principal())
+        style = SheetStyle.columns if flip_output else None
```
dwelch-spike, line 1378: From Logic agent: ~~**Critical:**~~ **Minor (conditional):** Correction to my earlier reproduction — this only triggers when the **terminal width is narrower than the rendered table**. In my 80-column shell environment, `show jobs queries --flip` produces `~~~~Query Jobs (ts)~~~~Query Jobs (ts)~~~~` (title appears twice). With `COLUMNS=300 ... --flip` it appears once. `@dwelch-spike` running in a wide terminal does NOT see the duplication.

Root cause is the sheet renderer's title-fill behavior: when the table content width exceeds the terminal width, it fills the title bar by repeating the title. Pre-existing behavior — `--flip` just makes the table much wider (1 column per field × ~21 fields vs. 1 column per job × N jobs), so it crosses the terminal-width threshold sooner.

So this is more cosmetic than I first claimed: users on narrow terminals (CI logs, small windows, `tmux` panes, ssh sessions) will see double-titles with --flip. Workarounds: widen the terminal, or use `like` to project to fewer columns. A real fix would still be worth doing — pass a flag through `sheet.render` so the flipped path opts out of the auto-repeat — but downgrade from Critical to Minor.

Test gap noted in `test_show_jobs_flip` still stands separately: that test cannot detect this regression class because it stubs `sheet.render`.

lib/live_cluster/show_controller.py
```diff
@@ -315,6 +316,23 @@ async def _do_default(self, line):
     "--flip",
     "Flip output table to show Nodes on Y axis and config on X axis.",
 )
+flip_jobs_modifier_help = ModifierHelp(
+    "--flip",
+    "Flip output table to show fields on Y axis and jobs on X axis.",
```
dwelch-spike: From Logic agent: **Important:** This help text describes the **unflipped** layout, not the flipped one. The default `show jobs queries` already has "fields on Y axis (rows: Node, Namespace, Type, ...) and jobs on X axis (columns)" — that's what users see without --flip. Running `--flip` reverses it so jobs become rows and fields become columns. So the help should read "Flip output table to show **jobs on Y axis and fields on X axis**" (the opposite of what's written). The PR description has the same inverted wording. The same string is duplicated verbatim at lib/collectinfo_analyzer/show_controller.py:94 — fix both. Consider also pulling `flip_jobs_modifier_help` / `where_jobs_modifier_help` / `like_jobs_modifier_help` / `for_jobs_modifier_help` / `jobs_usage_extras` into a single shared module since they're currently byte-identical across the two controllers and will drift under maintenance.

lib/live_cluster/get_controller.py
```diff
@@ -1169,6 +1171,93 @@ async def get_user_agents(self, nodes="all"):
...
+            ):
+                del host_jobs[trid]
+                continue
+            if set_filter and not next(
+                iter(util.filter_list([str(job.get("set", ""))], set_filter)), None
+            ):
+                del host_jobs[trid]
+                continue
+            if any(
+                not pat.search(str(job.get(field, ""))) for field, pat in where_compiled
```
dwelch-spike, line 1290: From Architecture agent: **Important:** `-where` reads raw server keys but users see display column names — and unknown fields silently filter every row to empty. Live test on a real server:

- `-where ns=test` → works (5 rows)
- `-where Namespace=test` → empty (user typed the display column header)
- `-where Node=...` → empty (display header; raw key isn't 'Node')
- `-where nonexistent_field=foo` → empty, no error or warning

Root cause here: `job.get(field, "")` returns "" for any missing key, then `pat.search("")` drops the row. Users can't tell "no matches" from "misspelled field".

Fix options (one or both):
1. Build a case-insensitive alias map (`Namespace`→`ns`, `Set`→`set`, `Node`→handled at host layer, `Module`→`module`, `Type`→`job-type`, `Transaction ID`→`trid`, `Time Since Done`→`time-since-done`, etc.) keyed off the display headers in `templates.show_jobs`. Translate the user's `field` before lookup.
2. After data fetch, compute the set of legal field names from the populated job dicts. If any `-where` field is absent from that set, raise `ShellException(f"unknown -where field '{field}'. Known: ...")`.

Bonus: the help text and PR description say nothing about which field names are valid. Either approach must update the help to point users at the column names they actually saw.

test/unit/live_cluster/test_getcontroller.py
```diff
@@ -903,6 +903,96 @@ async def test_get_all_flip(self):
...
+        self.cluster_mock.info_query_show.return_value = {
+            "1.1.1.1": {"1": {"ns": "test", "status": "active(ok)"}}
+        }
+
+        with self.assertRaises(ShellException) as ctx:
+            await self.controller.get_query(where=["status=("])
+
+        self.assertIn("status=(", str(ctx.exception))
+
+    async def test_get_query_where_nonexistent_field_filters_all(self):
```
dwelch-spike: From Tests agent: **Important:** This test cements the silent-filter-everything behavior as expected. `where=["nonexistent=foo"]` returns `{"1.1.1.1": {}}` with no error — that's the live UX defect (silent empty result on typo'd field), not a feature. As written, this test will block the fix: when someone changes `filter_jobs` to raise `ShellException` or warn on unknown fields, this assertion will start failing for the wrong reason. Either:

1. If the team decides the right behavior is `ShellException` on unknown field, replace this test with `assertRaises(ShellException)` and an `assertIn("unknown -where field", ...)` check.
2. If the team decides to keep silent-filter, rename the test to `test_..._is_silent_on_unknown_field` and add an inline comment saying this is intentional, so future readers don't think it's a bug. Currently the name `_filters_all` reads as accidental.

test/unit/view/test_view.py
```diff
@@ -259,6 +259,34 @@ def test_show_jobs(self):
             sources,
             common=common,
             selectors=["foo"],
+            style=None,
+        )
+
+    def test_show_jobs_flip(self):
```
dwelch-spike, line 270: From Tests agent: **Important:** This test asserts `style=SheetStyle.columns` was passed to a mocked `sheet.render` — it never produces output. The live render bug (title appears twice when --flip is used: `~~~~Query Jobs (ts)~~~~Query Jobs (ts)~~~~`) is invisible to this test because the renderer is stubbed. Add an actual-output test along the lines of `test_info_namespace_usage_division_actual_output` (~line 1366 in this same file): unmock `sheet.render` and `print_result`, capture stdout, and assert `output.count("Query Jobs (") == 1`. Until that exists, --flip can break visually on every release and the unit suite will stay green.

dwelch-spike:
Pulled the branch and ran `asadm.py` against a local 8.1.2 cluster with 5 finished query jobs. The four new modifiers and the existing-reused ones all wire through correctly:

| Modifier | Status |
|---|---|
| `--flip` / `-flip` transpose | ✅ jobs become rows, fields become columns (matches the Jira intent) |
| `like <regex>` projection | ✅ filters columns to matching field names (some base columns remain — Node/Namespace/Type/Progress%/Transaction ID/Time Since Done are always shown) |
| `for <ns>` and `for <ns> <set>` | ✅ namespace filter applies; non-matching ns returns empty |
| `-where <field>=<pattern>` | ✅ when field is the raw key (`status`, `ns`, `module`) |
| Multiple `-where` AND-together | ✅ `-where status=done -where ns=test` returns the intersection |
| Malformed `-where status` (no `=`) | ✅ surfaces `ERROR: invalid -where clause: 'status' — expected <field>=<pattern>` |
| Invalid regex `-where status=(` | ✅ surfaces `ERROR: invalid -where regex in 'status=(': missing ), unterminated subpattern` (resolves Copilot's earlier concern) |
| Reused `with <node>` and `trid <TRID>` | ✅ still work |
| `show jobs scans/sindex-builder` version errors | ✅ surface correctly with new modifiers attached |

**Three user-visible defects I found while testing** (inline comments cover details + fixes):

1. **`--flip` renders the section title twice.** The header line shows `~~~~Query Jobs (ts)~~~~Query Jobs (ts)~~~~` instead of once. Reproducible every time. Without --flip it's a single title. The columns-style sheet's geometry is making the title-render path fill-repeat.

2. **PR description and `--flip` help text describe the unflipped layout.** Both say "fields on Y axis, jobs on X axis" — but that's the default. After --flip you get the inverse. Two byte-identical copies of the broken help string exist (live + collectinfo controllers); worth pulling them into a shared module while fixing.

3. **`-where` uses raw server keys (`ns`, `set`, `job-type`, `trid`) but users see display column names (`Namespace`, `Set`, `Type`, `Transaction ID`).** `-where Namespace=test` returns empty silently. Same for any unknown / misspelled field — including the test `test_get_query_where_nonexistent_field_filters_all` which currently asserts this silent-empty as the correct behavior. Either alias display→raw, or raise `ShellException` on unknown fields, and update the help to point at the legal names.

Test suite (`pytest test/unit/{view,live_cluster,collectinfo_analyzer}`) passes — 188 tests + 4 subtests, no failures. The one CI failure is `sign-artifacts / sign` which is the same release-pipeline infra flake we saw on #467 (PR code touches only Python under lib/ and tests; build matrix and downstream checks have no plausible chain to a signing failure).

lib/live_cluster/get_controller.py
```diff
@@ -1169,6 +1171,93 @@ async def get_user_agents(self, nodes="all"):
...
+        line=line, arg="-flip", default=False, modifiers=modifiers, mods=mods
+    ) or util.check_arg_and_delete_from_mods(
+        line=line, arg="--flip", default=False, modifiers=modifiers, mods=mods
+    )
+
+    where = []
+    while "-where" in line:
+        w = util.get_arg_and_delete_from_mods(
+            line=line,
+            arg="-where",
```
dwelch-spike: From Architecture agent: **Minor:** `-where` is single-dash-only, but `-flip` accepts both `-flip` AND `--flip` (see lines 1232-1236 just above). The result is three different modifier styles in one command: bareword (`for`/`like`/`with`/`trid` — per `Modifiers` class in lib/utils/constants.py:139-143), dual-dash (`--flip`/`-flip`), and single-dash-only (`-where`).

The Jira ticket TOOLS-3482 specifies *"Dashed-flag style to match -flip"*, which justifies the dash itself. But `-flip` actually matches **both** dashed forms, so a user who mirrors `--flip` and types `--where status=active` will get an "unknown modifier" error. Long-form (`--where`) is also the more POSIX-idiomatic choice for a value-bearing flag, so users will reach for it first.

**Cheapest fix:** mirror the dual-form parsing used for flip — accept both `-where` and `--where`. About a 2-line change:

```python
where = []
while "-where" in line or "--where" in line:
    arg_name = "--where" if "--where" in line else "-where"
    w = util.get_arg_and_delete_from_mods(
        line=line, arg=arg_name, return_type=str, default=None,
        modifiers=modifiers, mods=mods,
    )
    ...
```

Update the usage strings (`jobs_usage_extras` and `where_jobs_modifier_help` at show_controller.py:323-325 / collectinfo_analyzer/show_controller.py:96-98) to show `[-where|--where <field>=<regex>]`. Dropping the dash entirely (bareword `where`) would be cleaner conceptually but breaks the repeatability marker and diverges from the Jira spec — symmetry with `-flip` is the lighter-weight option.

dwelch-spike: I think this is more important than the `minor` rating my bot gave it. I found the lack of --where confusing, let me know what you think.

test/unit/view/test_view.py
```diff
@@ -259,6 +259,105 @@ def test_show_jobs(self):
...
+                }
+        self.cluster_mock.get_node_names.return_value = node_names
+        self.cluster_mock.get_node_ids.return_value = node_ids
+        self.cluster_mock.get_expected_principal.return_value = "principal"
+
+        patch.stopall()
+
+        narrow_size = os.terminal_size((80, 24))
+        with patch_(
+            "lib.view.sheet.render.base_rsheet.get_terminal_size",
```
dwelch-spike: **Minor (test robustness):** This patch target is unresolvable on Python 3.10, so this test errors out 100% of the time there with `AttributeError: 'function' object has no attribute 'base_rsheet'`.

`lib/view/sheet/__init__.py` does `from .render import render`, which rebinds the `render` attribute on the `lib.view.sheet` package to the **function**, shadowing the `render` **submodule**. So the dotted path `lib.view.sheet.render.base_rsheet` can't be walked via attribute lookup. Python 3.12's `unittest.mock` importer happens to resolve it anyway (which is why CI — Python 3.12 per `Pipfile` — is green), but on 3.10 (e.g. a local `.venv`) it's a hard failure.

The production fix is correct — I verified the title renders exactly once by patching the real module object directly. Only the test is fragile. Suggested robust target:

```python
import importlib
brs = importlib.import_module("lib.view.sheet.render.base_rsheet")
with patch_.object(brs, "get_terminal_size", return_value=narrow_size):
    ...
```

lib/live_cluster/get_controller.py
```diff
@@ -1169,6 +1172,163 @@ async def get_user_agents(self, nodes="all"):
...
+            for job in host_jobs.values():
+                if isinstance(job, dict):
+                    known_fields.update(job.keys())
+
+    where_compiled = []
+    if where:
+        for clause in where:
+            field, _, pattern = clause.partition("=")
+            resolved = _resolve_where_field(field)
+            if known_fields and resolved not in known_fields:
```
dwelch-spike: **Minor (error-message UX):** The unknown-field check uses `known_fields` (keys actually present in the fetched job dicts), but the "Known fields" list in the message at line 1254 is `known_fields | _WHERE_FIELD_ALIASES.keys()` — a larger set. When the current jobs don't populate an aliased field, the two disagree and the error contradicts itself.

Live repro against an 8.1.1 server with basic primary-index query jobs (no `set` key in the job dicts):

```
$ asadm -e "show jobs queries --where set=foo"
ERROR: unknown -where field 'set'. Known fields: [... 'set' ...]
```

It rejects `set` as unknown while listing `set` as known in the same breath. Same for `progress%`, `sindex-name`, etc. whenever the current jobs lack that key. A user filtering a scan/query that *does* target a set would reach for `--where set=...`; here they can't tell a typo from a field that's simply empty for these jobs — the exact "typo vs zero-match" confusion this validation was meant to remove.

Suggested fix — also accept any name that's a legitimate alias, so it returns no matches (instead of erroring) while real typos like `bogusfield` still error:

```python
if (
    known_fields
    and resolved not in known_fields
    and field.strip().lower() not in _WHERE_FIELD_ALIASES
):
    legal = sorted(known_fields | set(_WHERE_FIELD_ALIASES.keys()))
    raise ShellException(f"unknown -where field '{field}'. Known fields: {legal}")
```

dwelch-spike (approved):
LGTM!

PR #490, opened 2026-05-27
fix: pass 'asadm' explicitly in Docker smoke test version check

a-spiker:
## Summary
- `docker run IMAGE --version` fails with `exec: "--version": executable file not found in $PATH` because the Dockerfile uses `CMD ["asadm"]` (not `ENTRYPOINT`), so Docker treats `--version` as the command override rather than an argument to `asadm`
- Fix: pass `asadm` explicitly — `docker run IMAGE asadm --version` — consistent with how the bats shims already work (lines 624/628)

## Root cause
Seen in CI run [26542466330](https://github.com/aerospike/aerospike-admin/actions/runs/26542466330/job/78188376381) during the 5.0.1-rc1 build.

## Test plan
- [ ] Re-run the `Docker build, test & push` job and confirm the smoke test passes


pvinh-spike (approved):
LGTM

PR #493, opened 2026-06-03
feat: TOOLS-2537 add command aliases for improved usability in controllers

a-spiker:
## Summary

Adds a small, generic **command alias** mechanism so alternate spellings resolve to the same command, and uses it to make `show stats` behave exactly like `show statistics` in both live-cluster and collectinfo modes.

### Why not just add `"stats"` to the controller map?
`show`'s sub-commands are resolved via a `PrefixDict` (prefix matching). Registering `"stats"` as a second command key would make the shared prefix `stat`/`sta` match **two** keys → an "Ambiguous command" error, breaking the existing `show stat` shorthand.

### Approach
- `lib/base_controller.py`: added an `aliases` dict (alias → canonical command), initialized via `_init_aliases()` (defaults to empty, mirroring the other `_init_*` helpers). The command token is translated **before** lookup in both `_find_method` and `complete`. Aliases resolve ahead of prefix matching and are **not** added to the `PrefixDict`, so they introduce no ambiguity.
- `lib/live_cluster/show_controller.py` and `lib/collectinfo_analyzer/show_controller.py`: registered `aliases = {"stats": "statistics"}` on both `ShowController`s.

### Behavior
- `show stats`, `show stats namespace`, tab-completion → identical to `show statistics`.
- `show stat` / `show statistics` → unchanged (no regression).

## Testing
- New unit tests:
  - `test/unit/test_base_controller.py` — core mechanism: alias resolves via `_find_method`/`execute`/`complete`, and the alias is **not** registered as a command key.
  - `test/unit/live_cluster/test_show_controller.py` and `test/unit/collectinfo_analyzer/test_show_controller.py` — `stats → statistics`, the `stat` prefix still resolves unambiguously (regression guard), and `stats` is not a registered command.
- Full unit suite: **1379 passed, 1 skipped**. Coverage confirms every added line in `base_controller.py` is exercised.

### DIFF
--- a/lib/base_controller.py
+++ b/lib/base_controller.py
@@ -300,6 +300,7 @@ class BaseController(object):
     def complete(self, line):
         self._init()
         command = line.pop(0) if line else ""
+        command = self.aliases.get(command, command)
         commands = self.commands.get_key(command)
 
         logger.debug("Auto-complete: command {}".format(command))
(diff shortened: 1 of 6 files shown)


dwelch-spike (dismissed):
LGTM, nice quality of life feature, maybe more commands can support this in the future.

pvinh-spike (approved):
LGTM

PR #495, opened 2026-06-04
feat: TOOLS-3298 deprecate 'numeric' sindex type in favor of 'integer'

a-spiker:
## Summary

Server **8.1.3** renames the secondary-index `numeric` type to `integer`. This change makes `asadm` aware of that rename across both **creating** and **listing** secondary indexes, while staying fully backward-compatible across server versions.

- Adds `integer` as a first-class bin-type keyword; keeps `numeric` as a deprecated alias.
- `show sindex` requests the new v2 list format on supporting servers so the renamed `integer` type is reported.
- Adds the version constant `SERVER_SINDEX_INTEGER_TYPE_FIRST_VERSION = "8.1.3"`.

## `manage sindex create`

The wire type sent to the server is selected per the cluster's minimum build version (via `check_version_support` then `info_sindex_create`).

| User types | Server | Result |
|---|---|---|
| `integer` | >= 8.1.3 | sends `type=integer` |
| `integer` | < 8.1.3 | **error**: "requires server v. 8.1.3 or later. Use 'numeric'..." |
| `numeric` | >= 8.1.3 | **warning** (deprecated, replaced by `integer`) and sends `type=integer` |
| `numeric` | < 8.1.3 | sends `numeric` (legacy, no warning) |

## `show sindex` (list)

On servers >= 8.1.3, `info_sindex` sends `sindex-list:v=v2` instead of `sindex-list:` so the server reports the renamed `integer` type. The format is chosen per node from each node's own build version (the list is fanned out to all nodes), so mixed-version clusters are handled correctly. Servers < 8.1.3, or nodes whose build is unknown, fall back to the default (v1) format.

## Changes

- `lib/utils/constants.py`: new `SERVER_SINDEX_INTEGER_TYPE_FIRST_VERSION` constant.
- `lib/live_cluster/manage_controller.py`: new `do_integer` keyword and deprecated `numeric` alias; version guard that errors on `integer` for servers < 8.1.3; deprecation warning when `numeric` is upgraded to `integer`; updated help text and the bin-type-required error message.
- `lib/live_cluster/client/node.py`:
  - `info_sindex_create` translates the deprecated `numeric` alias forward to `integer` on supporting servers.
  - `info_sindex` requests the `v=v2` list format on servers >= 8.1.3 (gated per node on its own build version, with a fallback to v1 when the build is unavailable).

## Testing

- `make unit` (full suite green).
- Added/updated unit tests in `test/unit/live_cluster/test_manage_controller.py` and `test/unit/live_cluster/client/test_node.py` covering all four create behaviour cells plus the warning/error paths, and the new list version-gating (v1 on older servers, v2 on 8.1.3+, and the v1 fallback when the build is unknown).
- Branch coverage confirms every new line and both branches of the new guards/warnings are exercised.


dwelch-spike (dismissed):
LGTM. Did some local testing with an older server and observed passing unit tests for new functionality. Looks like code adheres to the description. Lets make sure this is integration tested by QE against the new server.

dwelch-spike (approved):
LGTM

PR #496, opened 2026-06-09
fix: TOOLS-3596 collectinfo silently drops nodes whose info calls error/time out

a-spiker:
## TOOLS-3596

Jira: https://aerospike.atlassian.net/browse/TOOLS-3596

## Problem

During `asadm collectinfo`, a node whose per-node `info_*` calls failed in the first parallel burst (most commonly `asyncio.TimeoutError` against the 1s default per-node timeout) could disappear entirely from `ascinfo.json`. Every `show`/`info` table in the offline analyzer was then missing the node while cluster-level data (`Cluster Size` in `summary`) still reported the full count, so the bundle was internally inconsistent. The exception was swallowed without a log line, making the failure undiagnosable from `collectinfo_debug.log`.

Observed live (see the Jira for the incident bundle analysis): a healthy 15-node cluster produced a 14-node bundle. The missing node answered identical commands in later rounds of the same run, confirming a transient first-burst timeout.

## Root causes

1. **Silent swallowing and erasure.** `async_return_exceptions` caught every exception and returned it as the per-node value with no log. Downstream, `util.filter_exceptions` deleted exception-valued keys and `GetStatisticsController` popped nodes whose namespace stats errored, leaving no trace.
2. **Dump gated on `as_map` keys.** `_get_collectinfo_data_json` iterated only over `as_map` (statistics + config). A node that lost every stats and config call in the first burst was dropped from the JSON even though its `meta_map`/`sys_map`/`histogram_map`/`latency_map` data had been retrieved successfully.
3. **The offline analyzer could not represent a partial node.** It registered only nodes carrying `as_stat.meta_data`, and several unguarded accesses either crashed commands or silently degraded the whole bundle when a node with partial data was present. Fixing only the collection side would have produced bundles the analyzer mishandles.
4. **A dead `raise_exception` kwarg masked real errors.** The decorator unconditionally overwrote `raise_exception=False`, so `info_logging_config` crashed on the returned exception object and recorded a misleading `AttributeError` instead of the original failure.

### Collection (live cluster)

- **Every queried node lands in the snapshot.** The expected node set (alive + selected) is captured before the collection bursts, and the new `_build_dump_map` merges over the union of that set and every section map's keys. A node whose stats/config calls failed still appears via its surviving sections; a node that produced nothing appears with an empty `as_stat`. Node ordering is sorted so bundles are stable across runs. A `WARNING` names any node with no Aerospike data and suggests a larger `--timeout`.
- **Collectinfo-only timeout raise.** `_run_collectinfo` raises the per-node timeout to `max(--timeout, 5s)` for the run via new `Cluster.set_timeout` / `Node.set_timeout` / `ASSocket.settimeout` (already-pooled sockets are updated in place). The restore runs first in the `finally` and only when the timeout was actually raised, so a teardown failure cannot leak the elevated value into interactive use. The interactive default stays 1s and an explicit larger `--timeout` is honored.
- Removed a no-op `filter_exceptions(json_dump)` call that operated on the already-serialized JSON string.

### Diagnostics

- `async_return_exceptions` logs every swallowed exception at DEBUG (call name, `ip:port`, exception repr), so future bundles record which command failed against which node and why. It now also honors `raise_exception=True` by raising the original exception, so `info_logging_config` surfaces the real error.
- `util.filter_exceptions` logs each dropped key (resolving the long-standing TODO there).
- `get_namespace` and `get_strong_consistency_namespace` share a `_remove_and_log_failed_nodes` helper that logs every per-node drop (error or empty) instead of erasing it silently.

### Offline analyzer (`asadm --collectinfo`)

- **Degraded nodes are visible.** The analyzer now registers every node present in `ascinfo.json`, not only meta-bearing ones. A rescued node appears in the bundle listing and in views with `N/E` / `~~` markers, the same rendering live mode uses for a node whose info calls fail, instead of being invisible while `Cluster Size` says otherwise.
- **Principal handling is best-effort.** The expected principal is computed over nodes with known ids rather than collapsing to `UNKNOWN_PRINCIPAL` the moment one id is missing. Since principal-scoped data (ACL) is stored only on the node it was collected from, `get_data(nodes="principal")` falls back to all nodes when the resolved principal has no data for the requested section, so `show users`/`show roles` still return the collected ACL even if the true principal's id was lost.
- **Crash and silent-degradation guards.** `_create_node_ip_map` previously raised `KeyError` on a node without `meta_data`, which a bare `except` turned into silently skipping the `aerospike.conf` original-config merge for the whole bundle. The node-id mappings used by `show udfs/sindex/racks` had the same unguarded access. Both now skip nodes without usable metadata.
- **`summary` no longer aborts on a partial node.** `compute_license_data_size` skips (with a warning about possible under-reporting) a host that reported namespace stats but whose build call failed, instead of raising and killing the whole summary. Partial nodes make this shape common.

## Approach notes

- The union + seeding happens in `_build_dump_map` because it is the single choke point where all section maps meet. The alternatives were worse: teaching each of the ~9 getters to preserve failed node keys needs that many coordinated changes, and per-call retries amplify load on the exact contention that caused the timeouts.
- 5s is not a guarantee on very large or loaded clusters. The difference is that a residual failure is now visible (warning plus a present-but-empty node) rather than a silent drop.
- A node that legitimately leaves the cluster mid-run now shows up as an empty node plus a warning. That noise is the accepted cost of never silently losing a node that was queried.

## Testing

- Full unit suite: **1460 passed, 1 skipped** (master baseline was 1425). New tests cover: the dump union, seeding, no-data warning, and stable ordering; timeout raise/restore including restore-on-teardown-failure and not lowering a larger `--timeout`; `set_timeout` fan-out down to pooled sockets; `raise_exception` propagation and `info_logging_config` surfacing the real error; the shared drop-and-log helper including the empty-result case; analyzer registration of empty-`as_stat` nodes; best-effort principal (known ids, none known, single node) and the principal-scope fallback; parser and node-id mapping guards; the license-size skip.
- `black --check` clean; `flake8` introduces no new warnings.
- Offline end-to-end against a synthetic bundle containing a degraded node (empty `as_stat`): before this PR the analyzer showed 2 of 3 nodes while `Cluster Size` said 3; after, the node appears with `N/E` markers and `summary`, `info`, and `show` commands run clean, with the principal marker on the right node.
- Manual end-to-end against a live cluster is still pending: verify `ascinfo.json` node count matches `Cluster Size` and the new diagnostics appear in `collectinfo_debug.log`.

## Known residuals (out of scope, tracked separately)

- With `--enable-ssh`, a single node's system-stats exception still aborts the whole snapshot (pre-existing `raise` in `_get_collectinfo_data_json`). It is loud rather than silent, but inconsistent with this PR's per-node degradation; worth its own ticket.
- No retry logic was added for failed info calls; a burst exceeding the raised timeout still loses the failed sections for that node, now visibly.

## Workaround for users on current releases

Re-run with `asadm --timeout 5` (or 10).


dwelch-spike:
**PR Review (version 2026.07.06)** — updated for re-review at `16a5154`

**Promises:**
1. Every queried node lands in `ascinfo.json` — no silent node drops (union merge + expected-node seeding).
2. Every swallowed per-node info-call failure is logged and diagnosable from `collectinfo_debug.log`.
3. A WARNING names any node for which collectinfo captured no Aerospike data.
4. Per-node timeout is raised to `max(--timeout, 5s)` for the collectinfo run only, and always restored.
5. The offline analyzer registers and renders degraded nodes (`N/E` markers) instead of hiding them or crashing.
6. Principal handling is best-effort; collected ACL data is still returned when the true principal's node id was lost.

| Gate | Status |
|------|--------|
| Contract | fail — promise 3 still under-delivers: the new predicate is defeated by the locally-derived `meta_data["ip"]` (see reopened F2 thread) |
| Impacted surface | pass |
| Failure paths | pass — TimeoutError/alive fix verified, including connect-time dead-node detection |
| Evidence | blind spot — the F2 test still uses a `meta_map` shape (`ip` omitted) that `_get_as_metadata` cannot produce; full unit suite at `16a5154` run locally: 1467 passed, 1 pre-existing environmental failure (`test_ssh.py`, fails identically on master) |
| Quality | pass |

**Prior findings:**
| Finding | Status |
|---------|--------|
| Logic: transient timeout flips `alive=False` on Python 3.12 (node.py) | Addressed in `16a5154` — thread resolved |
| Data Flow: "no Aerospike data" warning under-fires (collectinfo_controller.py) | **Unresolved** — fix attempt defeated by locally-derived `meta_data["ip"]`; one-line follow-up posted in thread |
| Logic: `node_id == ""` passes the known-ids filter (collectinfo_log.py) | Addressed in `16a5154` — thread resolved |
| Architecture: PRINCIPAL→ALL fallback fires for all-node types (collectinfo_log.py) | Addressed in `16a5154` — thread resolved |
| Performance: license warning fires per namespace × host (common.py) | Addressed in `16a5154` — thread resolved |
| Control Flow: timeout restore vs in-flight sockets / over-claiming comment (collectinfo_controller.py) | Addressed in `16a5154` — thread resolved |

No new issues introduced by `16a5154` beyond the F2 gap above.

lib/live_cluster/client/node.py
```diff
@@ -94,8 +93,19 @@ async def wrapper(*args, raise_exception=False, **kwargs):
         except Exception as e:
             exception = e
 
+        # Without this log, per-node info-call failures (e.g. asyncio.TimeoutError during a
```
dwelch-spike, line 100: From Logic agent:

Severity: important
Problem Confidence: high
Solution Confidence: medium

## Problem

On the Python this repo ships (Pipfile pins 3.12), the `except OSError` branch above this comment catches the very `asyncio.TimeoutError` this PR is about, and flips `args[0].alive = False`. A single transient >timeout call therefore marks a healthy node dead, and every subsequent cluster call in the same collectinfo burst excludes it (`Cluster.get_nodes` filters on `alive`, cluster.py:507) — so one slow call cascades into losing **all** remaining sections for that node in the snapshot, not just the one that timed out. The union+seeding still puts the node in the bundle (empty/partial), but the data loss is much larger than one failed call, and the 5s raise only narrows the window.

## Reasoning Chain

1. Since Python 3.11, `asyncio.TimeoutError` **is** the builtin `TimeoutError`, which **is an `OSError` subclass**. Verified empirically on 3.12.3: a `wait_for` timeout is caught by `except OSError`, not `except Exception`. (On ≤3.10 it derives from plain `Exception`, which is why the generic-branch assumption held historically — the Jira's "alive was never flipped, ruling out OSError" elimination is only valid pre-3.11.)
2. `alive = False` → `get_nodes()` excludes the node for every info call that resolves its node list afterwards. This matches the incident signature better than the Jira's reading: `3B3`'s single-round meta calls (`node`, `racks:`) had *zero* responses — consistent with mid-burst exclusion — and it recovered in round 2 because `needs_refresh()` returns True for dead nodes and `refresh_connection()` resurrects them.
3. The new debug log will at least record `TimeoutError()` now, but the classification itself remains: transient timeout ≡ node death on 3.12.

## Solution

Catch the timeout before the `OSError` branch and do not flip `alive`:

```python
except asyncio.TimeoutError as e:  # builtin TimeoutError on 3.11+
    exception = e
except (ASInfoNotAuthenticatedError, ASProtocolConnectionError) as e:
    ...
```

Caveat (why Solution Confidence is medium): a genuinely hung-but-accepting node would then never be marked dead by info calls, so verify nothing else relies on timeout-driven `alive` flips before changing it. If that's out of scope here, this deserves its own ticket referenced from TOOLS-3596 — it is the same failure mode this PR diagnoses.

lib/live_cluster/collectinfo_controller.py
```diff
@@ -527,14 +592,25 @@ async def _get_collectinfo_data_json(
             if node in masking_map:
                 dump_map[node]["as_stat"]["masking"] = masking_map[node]
 
-        snp_map = {}
-        snp_map[cluster_name] = dump_map
-        return snp_map
+        no_data_nodes = [
+            node for node in expected_node_keys if not dump_map[node]["as_stat"]
```
dwelch-spike: From Data Flow agent:

Severity: important
Problem Confidence: high
Solution Confidence: medium

## Problem

The "no Aerospike data" warning almost never fires in the case the PR is built around: a node that is *reachable* but whose every info call failed. Real collection seeds non-empty `as_stat` content for such a node, so `not dump_map[node]["as_stat"]` is False and the warning is skipped. The warning effectively only covers the narrower case where the node was dropped from `get_nodes()` mid-run (the `alive=False` path). Promise 3 of the PR ("a WARNING names any node with no Aerospike data") silently under-delivers.

## Reasoning Chain

1. `_get_as_metadata` (line 327) iterates `builds` keys; a node whose every meta call returned an exception is still a key in `builds`, and `_check_for_exception_and_set` stores `""` for each failed call — the node gets `meta_data` full of empty strings.
2. `_get_as_histograms` (line 371) and `_get_as_latency` (line 389) seed `map[node] = {}` for every node in the response *before* the exception check; `_get_as_user_agents` seeds `[]`.
3. So the all-calls-failed node's `as_stat` is `{"meta_data": {"node_id": "", ...}, "histogram": {}, "latency": {}, "user_agents": []}` — truthy → no warning, even though there is no usable data.
4. The unit tests (`test_node_with_no_data_anywhere_is_included_and_warned`) pass because they construct section maps *without* these seeded shapes — shapes real collection cannot produce.

## Solution

Make the predicate mean "no usable data" rather than "no keys": e.g. treat `as_stat` as empty when every section is empty and every `meta_data` value is `""` (or check specifically for absent/empty `statistics`, `config`, and `meta_data.node_id`). Add a test at the `_get_collectinfo_data_json` level whose mocked cluster calls return exception values, so the maps have production shapes. Solution Confidence medium because the exact "usable" predicate is a judgment call (e.g. a node with only sys_stat + seeded empties should presumably still warn).

lib/collectinfo_analyzer/collectinfo_handler/collectinfo_log.py
```diff
@@ -402,14 +413,20 @@ def get_summary(self, stanza=""):
 
     def get_expected_principal(self) -> str:
         try:
+            # A node without meta_data (info calls failed during collection, TOOLS-3596)
+            # keeps the default id "N/E"; compute a best-effort principal from the nodes
+            # whose ids are known instead of giving up on the whole snapshot.
+            known_ids = [n.node_id for n in self.nodes.values() if n.node_id != "N/E"]
```
dwelch-spike: From Logic agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

The `known_ids` filter excludes only the `"N/E"` default, but collection produces a second "unknown id" shape: `node_id == ""`. If *all* nodes carry `""`, `known_ids` is non-empty (`["", ...]`), the `UNKNOWN_PRINCIPAL` branch is skipped, and the loop returns the `"0"` sentinel as the expected principal instead of `UNKNOWN_PRINCIPAL`.

## Reasoning Chain

1. On the collection side, `_check_for_exception_and_set` (collectinfo_controller.py:289) stores `""` for a node whose `node` info call failed while its `build` call succeeded — so `meta_data.node_id == ""` is a real bundle shape this PR makes more common.
2. The analyzer's `_set_node_id` calls `set_node_id("")`, which stores `""` as-is (it only strips a leading `*`), not `"N/E"`.
3. `"" != "N/E"` → it enters `known_ids`. Mixed clusters are fine (`"".zfill(16)` never wins the max), but the all-degraded case returns `"0"`. Other new guards in this PR (log_handler.py `if not node_id`, `_create_node_ip_map` `if node_id == ""`) already treat `""` as unknown — this filter is the one spot that doesn't.

## Solution

```python
known_ids = [n.node_id for n in self.nodes.values() if n.node_id and n.node_id != "N/E"]
```

lib/collectinfo_analyzer/collectinfo_handler/collectinfo_log.py
```diff
@@ -352,6 +356,13 @@ def get_data(
         except Exception:
             pass
 
+        if nodes == NodeSelection.PRINCIPAL and principal_ip and not data:
```
dwelch-spike: From Architecture agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

The PRINCIPAL→ALL fallback fires for *every* data type, not just principal-scoped sections (acl/masking). For sections collected from all nodes, a caller that asked for one node's data can silently receive N nodes' worth, and views render duplicated rows.

## Reasoning Chain

1. For acl/masking the fallback is exactly right: in a healthy bundle only one node holds the section, so falling back to ALL returns precisely the holder — the new tests prove this.
2. But e.g. `get_xdr_filters` defaults to `nodes=PRINCIPAL` (collectinfo_analyzer/get_controller.py:185) over `config`-type data that exists on **every** node. If the resolved principal's `config` section is missing (its get-config calls failed — the partial-node world this PR creates), `data` is empty, the fallback fires, and the caller gets all nodes' xdr filters where it expected one node's.
3. Not wrong data, but noisy — and it masks the fact that the principal specifically failed.

## Solution

Gate the fallback on the result looking principal-scoped, e.g. only use the fallback result when it contains exactly one node (`len(all_data) == 1`), or restrict it to the principal-scoped types (`acl`, `masking`). Alternatively accept the noise and note it in the comment — but then the duplicated-row behavior should be a conscious choice, not an accident of the `config` case.

lib/utils/common.py
```diff
@@ -888,7 +888,15 @@ def compute_license_data_size(
             )
 
             if host_build_version is None:
-                raise Exception("could not find host %s in build responses" % host_id)
+                # A node can report namespace stats while its build call failed
+                # (TOOLS-3596); skip its contribution instead of aborting the whole
+                # summary.
+                logger.warning(
+                    "Could not find host %s in build responses; license usage may be "
```
dwelch-spike: From Performance agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

This warning sits inside the per-namespace × per-host loop, so a single degraded host produces one near-identical warning per namespace it appears in — a 20-namespace cluster logs ~20 copies during every `summary`, drowning the signal the warning is meant to provide.

## Reasoning Chain

1. `compute_license_data_size` iterates namespaces, and within each namespace iterates hosts; `host_id` missing from `server_builds` is a per-host condition that is identical for every namespace.
2. The host is missing from `server_builds` because its build call failed — a whole-node condition, not a per-namespace one.

## Solution

Collect the missing host ids in a set during the loops and emit one warning after, e.g. `"Could not find host(s) %s in build responses; license usage may be under-reported"`. Alternatively track already-warned hosts in a local set and skip repeats.

lib/live_cluster/collectinfo_controller.py
```diff
@@ -1009,6 +1098,10 @@ async def _run_collectinfo(
                     "Failed to archive collectinfo logs. See earlier errors for more details."
                 )
         finally:
+            # Restore the per-node timeout first so an elevated timeout cannot leak into
+            # subsequent interactive use if teardown fails (TOOLS-3596).
+            if collectinfo_timeout != original_timeout:
+                self.cluster.set_timeout(original_timeout)
```
dwelch-spike: From Control Flow agent:

Severity: minor
Problem Confidence: medium
Solution Confidence: medium

## Problem

The restore-before-teardown ordering is right, but on an abort path the restore can race still-running collection tasks: a socket checked out by an in-flight call at restore time keeps the elevated 5s timeout and returns to the pool with it, where subsequent *interactive* commands will reuse it.

## Reasoning Chain

1. `asyncio.gather` (default `return_exceptions=False`, as used in `_get_collectinfo_data_json`) propagates the first exception immediately but does **not** cancel its sibling tasks — they keep running while the exception unwinds into this `finally`.
2. `Node.set_timeout` updates only sockets currently *in* the pool; a socket popped by `_borrow_socket` for an in-flight call is invisible to it and re-enters the pool afterwards still carrying 5s.
3. Impact is bounded and benign-ish (a too-large timeout only delays failure detection), which is why this is minor — but it contradicts the comment's "cannot leak the elevated value into interactive use" claim on the error path. Problem Confidence medium: I traced the gather semantics and pool mechanics but did not construct a live repro.

## Solution

Either accept and soften the comment ("best-effort restore; in-flight sockets may briefly retain the raised timeout"), or make `ASSocket` read the timeout from its owning `Node` at call time instead of caching a copy per socket, which removes the stale-copy class entirely. The second is cleaner but touches every `wait_for` site in assocket.py.

lib/live_cluster/client/node.py
a-spiker: Thanks a lot for this catch, this one was really sharp.

You are right. On 3.12 asyncio.TimeoutError is an OSError subclass, so a transient timeout was hitting the OSError branch and flipping alive=False, which then cascaded the node out of the later batches in the same burst.

Fix: added a dedicated `except asyncio.TimeoutError` branch before the OSError one, so a transient timeout no longer marks the node dead. Debug log is kept as is.

One caveat we are aware of: a genuinely hung node will now not be marked dead via info timeouts, but connection errors still do that and dead nodes self heal on refresh, so we are fine with it for the collectinfo path. This will be in the next push.

lib/live_cluster/collectinfo_controller.py
a-spiker: Good catch, you are absolutely right.

The seeded meta_data (empty strings, plus node_names which comes from local state) was making as_stat truthy, so the warning was basically never firing for the exact case we built it for.

Fix: changed the predicate. Added a generic `util.has_content()` helper and `_as_stat_has_aerospike_data()` which checks for real non-empty data across the sections and ignores node_names. Also added a controller level test that feeds production shaped maps (empty-string meta plus seeded empty sections), so it will fail if this regresses. Coming in the next push.

lib/collectinfo_analyzer/collectinfo_handler/collectinfo_log.py
a-spiker: Nice one.

Fix: changed the filter to `if n.node_id and n.node_id != "N/E"`, so the empty string is also treated as unknown. Now an all-degraded snapshot returns UNKNOWN_PRINCIPAL instead of the "0" sentinel. Added a test for the all-empty-id case as well. Thanks for pointing this out.

lib/collectinfo_analyzer/collectinfo_handler/collectinfo_log.py
a-spiker: Agreed, it was firing for config type data also.

Fix: gated the fallback to principal scoped types only. Pulled the tuple into constants as `PRINCIPAL_SCOPED_TYPES = ("acl", "masking")`, and get_data now falls back only when type is in that set. So a single node config request will not silently return all nodes worth of data anymore. Added a test that a config request does not fall back. Will push shortly.

lib/utils/common.py
a-spiker: Right, it was logging once per namespace per host.

Fix: now collecting the missing host ids in a set inside the loop and logging a single consolidated warning after the loops finish. Added a test with 3 namespaces asserting only one warning fires. Thanks for flagging this.

lib/live_cluster/collectinfo_controller.py
a-spiker: Fair point, and thanks for the careful trace, the comment was over claiming.

Two things done: (1) softened the comment to say the restore is best effort and in-flight sockets may briefly retain the raised timeout until they are recycled, and (2) wrapped the restore in its own try/finally so a failure there can never skip the logger and color teardown.

lib/live_cluster/client/node.py
dwelch-spike: Resolved by 16a51544b9a36fcf9ef403bd53b913b6a95e35fe — the dedicated `except asyncio.TimeoutError` branch is ordered before the alive-flipping branches, which is exactly right on both interpreters (on 3.12 the name is the builtin `TimeoutError`/OSError subclass, on ≤3.10 a distinct `Exception` subclass). Verified that connect-time dead-node detection is unaffected: `connect()` checks `isinstance(self.node_id, Exception)` and sets `alive=False` itself. Both new tests are meaningful. https://github.com/aerospike/aerospike-admin/blob/16a51544b9a36fcf9ef403bd53b913b6a95e35fe/lib/live_cluster/client/node.py#L87-L90

lib/collectinfo_analyzer/collectinfo_handler/collectinfo_log.py
dwelch-spike: Resolved by 16a51544b9a36fcf9ef403bd53b913b6a95e35fe — `if n.node_id and n.node_id != "N/E"` treats the empty string as unknown, and the all-empty-ids test locks in UNKNOWN_PRINCIPAL. https://github.com/aerospike/aerospike-admin/blob/16a51544b9a36fcf9ef403bd53b913b6a95e35fe/lib/collectinfo_analyzer/collectinfo_handler/collectinfo_log.py#L428-L432

lib/collectinfo_analyzer/collectinfo_handler/collectinfo_log.py
dwelch-spike: Resolved by 16a51544b9a36fcf9ef403bd53b913b6a95e35fe — fallback gated on `PRINCIPAL_SCOPED_TYPES = ("acl", "masking")`, which matches exactly the type keys the analyzer uses (log_handler.py:236 `type="masking"`, :249 `type="acl"`), and the new test proves a `config` request no longer falls back. https://github.com/aerospike/aerospike-admin/blob/16a51544b9a36fcf9ef403bd53b913b6a95e35fe/lib/collectinfo_analyzer/collectinfo_handler/collectinfo_log.py#L363-L368

lib/utils/common.py
dwelch-spike: Resolved by 16a51544b9a36fcf9ef403bd53b913b6a95e35fe — missing hosts accumulate in a set and produce one consolidated warning after the loops; the test asserts a single warning across three namespaces. https://github.com/aerospike/aerospike-admin/blob/16a51544b9a36fcf9ef403bd53b913b6a95e35fe/lib/utils/common.py#L930-L936

lib/live_cluster/collectinfo_controller.py
dwelch-spike: Resolved by 16a51544b9a36fcf9ef403bd53b913b6a95e35fe — comment softened to best-effort, and wrapping the restore in its own try/finally so a `set_timeout` failure can never skip logger/color teardown is a genuine robustness improvement beyond what was asked. https://github.com/aerospike/aerospike-admin/blob/16a51544b9a36fcf9ef403bd53b913b6a95e35fe/lib/live_cluster/collectinfo_controller.py#L1120-L1128

lib/live_cluster/collectinfo_controller.py
dwelch-spike: From Data Flow agent:

Severity: important
Problem Confidence: high
Solution Confidence: high
Status: still open at 16a5154 — the new predicate still returns True for the production shape of a fully-failed node, so the warning is still suppressed in exactly the case it targets.

## Problem

`_as_stat_has_aerospike_data` excludes only `node_names` as locally derived, but `meta_data["ip"]` is local state too: `info_ip_port` just returns `create_key(self.ip, self.port)` (node.py:1215-1223) and never touches the network, so it succeeds — and is non-empty — for every queried node, including one whose every real info call timed out. The new test passes only because its `meta_map` for node B omits the `ip` key, a shape `_get_as_metadata` cannot produce (it sets `ip` for every node in `builds`).

Note the (correct) TimeoutError fix in this same commit makes the gap *more* certain to occur: the timing-out node now stays `alive` through the whole burst, so it is guaranteed to be in `builds` and receive the seeded meta including `ip`.

## Reasoning Chain

1. Ran the new predicate at 16a5154 against the true production shape — all info-derived meta values `""`, plus `"ip": "10.0.0.1:3000"` and `node_names`, with seeded empty `statistics`/`config`/`histogram`/`latency`/`user_agents`:
   `_as_stat_has_aerospike_data(as_stat)` → `True` (warning suppressed).
2. Same input minus the `ip` key (the test's shape) → `False` (warning fires).
3. Audit of all 13 fields `_get_as_metadata` sets: `ip` (via `info_ip_port`) and `node_names` (via `cluster.get_node_names()`) are the only two that never hit the network; `endpoints`, `services`, `udf`, `health`, `best_practices`, `jobs`, `feature-key`, `release`, `asd_build`, `edition`, `node_id` all come from real info calls and become `""` on failure.

## Solution

Exclude `ip` alongside `node_names`:

```python
LOCALLY_DERIVED_META = ("node_names", "ip")  # set without any info call succeeding
...
if any(
    util.has_content(v) for k, v in section.items() if k not in LOCALLY_DERIVED_META
):
```

and add `"ip": "10.0.0.1:3000"` to node B's `meta_map` in `test_node_with_only_seeded_empty_sections_is_warned` so the test pins the production shape. A short comment naming *why* those two fields are excluded will keep the next locally-derived field from reintroducing this.

a-spiker: Good catch. ip is locally derived via info_ip_port and never fails, so it kept the predicate truthy; excluded it alongside node_names (constants.COLLECTINFO_LOCALLY_DERIVED_META_KEYS) and added a production-shape test that drives _get_collectinfo_data_json with all info calls returning TimeoutError, coming in the next push.

dwelch-spike (approved):
LGTM

PR #502, opened 2026-06-29
Show Used% for shmem primary/secondary index + highlight the self node

Alb0t:
Two related namespace-usage display improvements, plus a small refactor and the in-app docs for both.

1. Used% for shmem indexes with indexes-memory-budget — fills in a column that was previously blank.
2. Self-node highlight — the local node asadm is running on is now highlighted in cyan in the Node column (name prefixed with `@`), distinct from the existing green principal highlight (prefixed with `*`).

image
image

### Problem

- Used%: Used% for primary/secondary index only resolved for flash/pmem index types (via index_mounts_used_pct or a bytes/limit division). For shmem indexes with indexes-memory-budget configured (7.1+), the column rendered blank even though both the usage bytes and the budget were available.
- Self node: Output highlighted the cluster principal (green) but gave no indication of which node you were physically on — useful context when running asadm directly on a cluster node.

### Changes

- templates.py — Add a third fallback to the Projectors.Any chain for primary and secondary Used%: index_used_bytes / indexes-memory-budget (and the sindex equivalent). Placed last in the chain so flash/pmem nodes are unaffected; when no budget is configured the division returns None and the column stays hidden, exactly as before. Numerators use only the 7.0+ consolidated stat names — the budget denominator is 7.1+, so pre-7.0 memory_used_* fallbacks would be dead code (matches how lib/utils/common.py pairs these stats in the summary).
- cluster.py — Add Cluster.get_self_node(), returning the node_id of the first node flagged localhost (via the is_localhost() accessor), or "".
- collectinfo_log.py — Add _CollectinfoSnapshot.get_self_node() returning "" (a capture has no self node), keeping the snapshot duck-type compatible with Cluster for the shared CliView render paths.
- decleration.py — Add bold_cyan_alert formatter, following the existing *_alert factory pattern.
- templates.py — Apply bold_cyan_alert to the Node column for the self node, listed after green_alert so principal wins the renderer's first-match formatter selection. Add a shared _node_marker helper that prefixes `*` (principal) / `@` (self node) on the Node name column of every sheet and on the Node ID column of info network / show roster, so the signal survives copy/paste, piping, and --no-color. Degrades to no marker on sheets without a Node ID column.
- view.py — Thread self_node into the common dict at every render site. Extract the repeated dict(principal=…, self_node=…) construction (~20 sites) into a single CliView._common(cluster, **extra) helper.
- live_cluster_root_controller.py — Document the color + marker legend (green/`*` = expected principal, cyan/`@` = self node) at the top of the base help menu.
- Apply repo-pinned black (26.3.1) formatting.

### Tests

- NodeHighlightingTest (test_templates.py) renders the real info_network_sheet to JSON and asserts: green + `*` on the principal, cyan + `@` on the self node, no format/marker on other nodes, green precedence when a node is both principal and self, no markers when self_node is empty (collectinfo), and no crash on sheets without a Node ID column.
- test_get_self_node / test_get_self_node_returns_empty_when_no_localhost (test_cluster.py) cover the new Cluster method directly.
- Collectinfo compat test asserts CliView._common works against a spec'd _CollectinfoSnapshot, pinning the duck-type contract.
- The shmem Used% test asserts both the primary (25.0 %) and secondary (10.0 %) index percentages.

### Behavior

- Used% now populates for shmem primary/secondary index when indexes-memory-budget is set; unchanged for flash/pmem and when no budget is configured.
- Self node renders cyan with an `@` name prefix in the Node column of info / show output; principal remains green with `*`. When a node is both, green/`*` wins (principal takes precedence).
- Collectinfo mode (asadm -cf) renders exactly as before: self_node resolves to "" so nothing is highlighted or marked.

#### Notes / known limitations (pre-existing, not introduced here)

- Green = expected principal, computed as the highest node-id among the nodes asadm is connected to (get_expected_principal). When you don't seed the full cluster, this can differ from the real paxos_principal — green may land on the highest-id node of your subset. The new help legend calls this out.
- Self-node detection is IPv4-centric. It relies on a localhost/127.0.0.1 seed or an exact-string match of hostname -I against the node's service addresses. IPv6 is best-effort: ::1 isn't special-cased, and the comparison does no address normalization, so cyan may not appear on IPv6 deployments. Can be hardened in a follow-up.
=

### DIFF
--- a/lib/live_cluster/client/cluster.py
+++ b/lib/live_cluster/client/cluster.py
@@ -815,8 +815,10 @@ class Cluster(AsyncObject):
         self.node_lookup = LookupDict()
 
     def get_self_node(self):
+        # Returns the first node flagged localhost; if multiple co-located
+        # nodes are flagged, dict-iteration order decides which wins.
         for node in self.nodes.values():
-            if node.localhost:
+            if node.is_localhost():
                 return node.node_id
         return ""
 
--- a/lib/live_cluster/client/cluster.py
+++ b/lib/live_cluster/client/cluster.py
@@ -814,5 +814,11 @@ class Cluster(AsyncObject):
         self.nodes = {}
         self.node_lookup = LookupDict()
 
+    def get_self_node(self):
+        for node in self.nodes.values():
+            if node.localhost:
+                return node.node_id
+        return ""
+
     def get_seed_nodes(self):
         return list(self._seed_nodes)
--- a/lib/view/templates.py
+++ b/lib/view/templates.py
@@ -131,7 +131,7 @@ node_field = Field(
     Projectors.String("node_names", None),
     formatters=(
         Formatters.bold_cyan_alert(
-            lambda edata: edata.record["Node ID"] == edata.common.get("self_node")
+            lambda edata: edata.record["Node ID"] == edata.common["self_node"]
         ),
         Formatters.green_alert(
             lambda edata: edata.record["Node ID"] == edata.common["principal"]
--- a/lib/view/templates.py
+++ b/lib/view/templates.py
@@ -126,15 +126,33 @@ def weighted_avg(values: Iterable[float], weights: Iterable[float]):
 # Common fields.
 #
 
+
+def _node_marker(node_id, common):
+    """Plain-text counterpart of the green/cyan node highlighting, so the
+    signal survives copy/paste and --no-color: "*" = expected principal,
+    "@" = self node. Principal wins when a node is both."""
+    common = common or {}
+    if node_id:
+        if node_id == common.get("principal"):
+            return "*"
+        if node_id == common.get("self_node"):
+            return "@"
+    return ""
+
+
 node_field = Field(
     "Node",
     Projectors.String("node_names", None),
+    converter=(
+        lambda edata: _node_marker(edata.record.get("Node ID"), edata.common)
+        + edata.value
+    ),
     formatters=(
-        Formatters.bold_cyan_alert(
-            lambda edata: edata.record["Node ID"] == edata.common["self_node"]
-        ),
         Formatters.green_alert(
             lambda edata: edata.record["Node ID"] == edata.common["principal"]
+        ),
+        Formatters.bold_cyan_alert(
+            lambda edata: edata.record["Node ID"] == edata.common["self_node"]
         ),
     ),
 )
@@ -155,15 +173,14 @@ info_network_sheet = Sheet(
             "Node ID",
             Projectors.String("node_ids", None),
             converter=(
-                lambda edata: (
-                    "*" + edata.value
-                    if edata.value == edata.common["principal"]
-                    else edata.value
-                )
+                lambda edata: _node_marker(edata.value, edata.common) + edata.value
             ),
             formatters=(
                 Formatters.green_alert(
                     lambda edata: edata.record["Node ID"] == edata.common["principal"]
+                ),
+                Formatters.bold_cyan_alert(
+                    lambda edata: edata.record["Node ID"] == edata.common["self_node"]
                 ),
             ),
             align=FieldAlignment.right,
@@ -339,7 +356,6 @@ info_namespace_usage_sheet = Sheet(
                             Projectors.Number(
                                 "ns_stats",
                                 "index_used_bytes",
-                                "memory_used_index_bytes",
                             ),
                             Projectors.Number(
                                 "ns_stats",
@@ -425,7 +441,6 @@ info_namespace_usage_sheet = Sheet(
                             Projectors.Number(
                                 "ns_stats",
                                 "sindex_used_bytes",
-                                "memory_used_sindex_bytes",
                             ),
                             Projectors.Number(
                                 "ns_stats",
@@ -2517,15 +2532,14 @@ show_roster = Sheet(
             "Node ID",
             Projectors.String("node_ids", None),
             converter=(
-                lambda edata: (
-                    "*" + edata.value
-                    if edata.value == edata.common["principal"]
-                    else edata.value
-                )
+                lambda edata: _node_marker(edata.value, edata.common) + edata.value
             ),
             formatters=(
                 Formatters.green_alert(
                     lambda edata: edata.record["Node ID"] == edata.common["principal"]
+                ),
+                Formatters.bold_cyan_alert(
+                    lambda edata: edata.record["Node ID"] == edata.common["self_node"]
                 ),
             ),
             align=FieldAlignment.left,
--- a/lib/view/templates.py
+++ b/lib/view/templates.py
@@ -130,6 +130,9 @@ node_field = Field(
     "Node",
     Projectors.String("node_names", None),
     formatters=(
+        Formatters.bold_cyan_alert(
+            lambda edata: edata.record["Node ID"] == edata.common.get("self_node")
+        ),
         Formatters.green_alert(
             lambda edata: edata.record["Node ID"] == edata.common["principal"]
         ),
@@ -332,6 +335,17 @@ info_namespace_usage_sheet = Sheet(
                                 "index-type.mounts-size-limit",
                             ),
                         ),
+                        Projectors.Div(  # shmem with indexes-memory-budget (7.1+)
+                            Projectors.Number(
+                                "ns_stats",
+                                "index_used_bytes",
+                                "memory_used_index_bytes",
+                            ),
+                            Projectors.Number(
+                                "ns_stats",
+                                "indexes-memory-budget",
+                            ),
+                        ),
                     ),
                     converter=Converters.ratio_to_pct,
                     aggregator=ComplexAggregator(
@@ -405,6 +419,17 @@ info_namespace_usage_sheet = Sheet(
                             Projectors.Number(
                                 "ns_stats",
                                 "sindex-type.mounts-size-limit",
+                            ),
+                        ),
+                        Projectors.Div(  # shmem with indexes-memory-budget (7.1+)
+                            Projectors.Number(
+                                "ns_stats",
+                                "sindex_used_bytes",
+                                "memory_used_sindex_bytes",
+                            ),
+                            Projectors.Number(
+                                "ns_stats",
+                                "indexes-memory-budget",
                             ),
                         ),
                     ),
--- a/lib/view/view.py
+++ b/lib/view/view.py
@@ -102,6 +102,16 @@ class CliView(object):
         return " (" + str(timestamp) + ")"
 
     @staticmethod
+    def _common(cluster, **extra):
+        """Build the 'common' dict passed to sheet rendering. Extra keyword
+        arguments are merged in for sheets that need additional fields."""
+        return dict(
+            principal=cluster.get_expected_principal(),
+            self_node=cluster.get_self_node(),
+            **extra,
+        )
+
+    @staticmethod
     @reserved_modifiers
     def info_network(
         stats,
@@ -146,9 +156,8 @@ class CliView(object):
         common_key = util.find_most_frequent(cluster_keys)
         common_principal = util.find_most_frequent(cluster_principals)
 
-        common = dict(
-            principal=cluster.get_expected_principal(),
-            self_node=cluster.get_self_node(),
+        common = CliView._common(
+            cluster,
             common_size=common_size,
             common_key=common_key,
             common_principal=common_principal,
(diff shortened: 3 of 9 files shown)


kportertx:
**PR Review (version 2026.06.16)**

**Promises:**
1. Used% populates for shmem primary/secondary index when `indexes-memory-budget` is set (7.1+); unchanged for flash/pmem and blank when no budget.
2. The self node is highlighted cyan in the Node column, distinct from the green principal; when a node is both, green (principal) wins.
3. `Cluster.get_self_node()` returns the localhost node_id, or "".
4. ~20 repeated `dict(principal=...)` common-dict sites are refactored into `CliView._common(cluster, **extra)` with no behavior change.
5. The base help menu documents the color legend.

| Gate | Status |
|------|--------|
| Contract | **Fail** — promise 4 ("no behavior change") broken: collectinfo mode crashes; promise 2 (green wins) inverted |
| Impacted surface | **Fail** — `_common` reaches the collectinfo render path, which passes a non-`Cluster` object |
| Failure paths | **Fail** — `AttributeError` in all collectinfo info/show |
| Evidence | **Fail** — the two new behaviors have no behavioral test coverage |
| Quality | Pass — only a minor style nit |

**5 findings** posted inline (1 Critical, 2 Important, 2 Minor).

The Used% shmem fallback (promise 1) is correctly implemented — verified the `Projectors.Any`/`Div`/`Number` chain: shmem+budget resolves via the new last-in-chain `Div`, and a missing budget raises `NoEntryException` so the column stays blank, exactly as described. The blocker is that the `_common` refactor added `cluster.get_self_node()`, which exists on `Cluster` but not on the `_CollectinfoSnapshot` that collectinfo mode passes as `cluster` — so `asadm -cf` info/show output crashes. That, plus the missing tests (which would have caught both the crash and the color-precedence inversion), are the required fixes.

lib/view/view.py
```diff
@@ -101,6 +101,16 @@ def _get_timestamp_suffix(timestamp):
 
         return " (" + str(timestamp) + ")"
 
+    @staticmethod
+    def _common(cluster, **extra):
+        """Build the 'common' dict passed to sheet rendering. Extra keyword
+        arguments are merged in for sheets that need additional fields."""
+        return dict(
+            principal=cluster.get_expected_principal(),
+            self_node=cluster.get_self_node(),
```
kportertx, line 110: From Logic agent: **Critical:** `_common` unconditionally calls `cluster.get_self_node()`, but the shared `CliView.info_*`/`show_*` methods are also used by **collectinfo** mode, where the object passed as `cluster` is a `_CollectinfoSnapshot` (`collectinfo_analyzer/collectinfo_handler/collectinfo_log.py:68`), not a `Cluster`. That class implements `get_node_names`, `get_node_ids`, and `get_expected_principal` (so the old `dict(principal=cluster.get_expected_principal())` worked) but has **no `get_self_node` and no `__getattr__`**. So every collectinfo `info`/`show` command routed through these methods (e.g. `show_controller.py:167` `self.view.show_config(..., get_cinfo_log_at(...))`) now raises `AttributeError: '_CollectinfoSnapshot' object has no attribute 'get_self_node'`.

This breaks promise 4 ("no behavior change") — it's a hard regression for `asadm -cf`. The unit tests miss it because the live path uses a `MagicMock` that auto-stubs `get_self_node`.

**Fix:** add `get_self_node(self)` to `_CollectinfoSnapshot` returning `""` (a capture has no "self" node — mirrors how `get_expected_principal` is dual-implemented), or degrade gracefully in `_common`: `self_node=getattr(cluster, "get_self_node", lambda: "")()`.

lib/view/templates.py
```diff
@@ -130,6 +130,9 @@ def weighted_avg(values: Iterable[float], weights: Iterable[float]):
     "Node",
     Projectors.String("node_names", None),
     formatters=(
+        Formatters.bold_cyan_alert(
```
kportertx: From Control Flow agent: **Important:** The Node column renders **cyan, not green**, when a node is both the self node and the principal — the inverse of the PR's stated "green composes over cyan (principal wins)." The renderer does not compose formatters; `base_rsheet.py:843` returns the **first** formatter whose predicate matches and stops:
```python
for name, formatter in self.decleration.formatters:
    format_fn = formatter(edata)
    if format_fn is not None:
        return name, format_fn
```
Here `bold_cyan_alert` (self) is listed **before** `green_alert` (principal), so for a self+principal node the cyan branch matches first and green never runs. This is common (connecting to a single node, or the highest-node-id node, makes self == expected principal).

**Fix:** list `green_alert` before `bold_cyan_alert` in `node_field` so principal (green) wins, matching the contract and the help legend.

test/unit/view/test_view.py
```diff
@@ -57,7 +57,10 @@ def test_show_roster(self):
         node_names = {"1.1.1.1": "1.1.1.1 is my name"}
         node_ids = {"1.1.1.1": "ABCD"}
         principal = "test-principal"
-        common = {"principal": principal}
+        common = {
+            "principal": principal,
+            "self_node": self.cluster_mock.get_self_node.return_value,
```
kportertx, line 62: From Tests agent: **Important:** The two new behaviors have no behavioral coverage. All 22 `self_node` test edits are mock-plumbing of this form — they set the expected `common` dict to `get_self_node.return_value` (a `MagicMock` sentinel) and assert `sheet.render` received it. That verifies the `_common` wiring (promise 4) but exercises neither the cyan highlight nor the Used% logic:
- **No test renders the Node column and asserts** `bold_cyan_alert` is applied to the self node (and green to the principal, and which wins when both) — `grep cyan test/` returns nothing. This gap is why the precedence inversion (separate comment) is undetected.
- **`Cluster.get_self_node()` has no unit test** (localhost → node_id; none → "") — `test_cluster.py` has `test_get_expected_principal` as the natural sibling but no `test_get_self_node`. New logic, fully uncovered.

A collectinfo-mode render test would also have caught the Critical `AttributeError`. **Fix:** add a Node-column render test asserting the cyan token on self and green precedence, plus a direct `get_self_node` unit test.

test/unit/view/test_view.py
```diff
@@ -1666,6 +1723,42 @@ def test_info_namespace_usage_in_memory_no_used_pct(self):
...
+
+        f = io.StringIO()
+        with redirect_stdout(f):
+            CliView.info_namespace_usage(
+                ns_stats, service_stats, self.cluster_mock, timestamp="test-stamp"
+            )
+        output = f.getvalue()
+
+        self.assertIn("Used%", output)
+        self.assertIn("25.0 %", output)
```
kportertx: From Tests agent: **Minor:** This test feeds `sindex_used_bytes` + `indexes-memory-budget` (which should render the secondary-index Used% as 10.0%) but asserts only the **primary** index value (`"25.0 %"`). The new secondary-index shmem `Projectors.Div` (templates.py ~421) runs but its output is never checked — deleting or breaking that sindex projector would not fail this test. Add an assertion for the sindex percentage (e.g. `assertIn("10.0 %", output)`) so the sindex half of promise 1 has real evidence.

lib/live_cluster/client/cluster.py
```diff
@@ -814,5 +814,11 @@ async def close(self):
         self.nodes = {}
         self.node_lookup = LookupDict()
 
+    def get_self_node(self):
+        for node in self.nodes.values():
+            if node.localhost:
```
kportertx: From Style agent: **Minor:** `get_self_node` reads the raw `node.localhost` attribute, but `node.py` already exposes an accessor for exactly this (`def is_localhost(self): return self.localhost`) — prefer `node.is_localhost()` for consistency with the rest of the codebase. Also worth a one-line comment that this returns the *first* node flagged localhost: if `_is_any_my_ip(hostname -I)` marks multiple co-located nodes as localhost, the highlighted node is dict-iteration-order-dependent. Not a correctness issue for a cosmetic highlight, and the `""` empty-string default is consistent with `get_expected_principal`'s own fallback (keep it, don't switch to `None`).

lib/view/templates.py
a-spiker: when the node you connected to is also the principal (single-node clusters always, and any time you point asadm at the principal), the Node column renders cyan, not green. Principal node should have the priority

lib/view/templates.py
```diff
@@ -332,6 +335,17 @@ def usage_weighted_avg(edatas: list[EntryData]):
                                 "index-type.mounts-size-limit",
                             ),
                         ),
+                        Projectors.Div(  # shmem with indexes-memory-budget (7.1+)
```
a-spiker, line 355: The two new shmem Div projectors for Used% (primary and secondary) mix version epochs, the denominator indexes-memory-budget is 7.1+, but the numerators fall back to the pre-7.0 names memory_used_index_bytes / memory_used_sindex_bytes, which were replaced by index_used_bytes / sindex_used_bytes in 7.0. 

Since any server that has the budget (7.1+) always emits the consolidated names and never the memory_used_* ones, those fallbacks are dead and misleading. This matches how the summary handles it (lib/utils/common.py:1882-1908 pairs indexes-memory-budget only with the 7.0+ consolidated metrics).

lib/view/templates.py
```diff
@@ -130,6 +130,9 @@ def weighted_avg(values: Iterable[float], weights: Iterable[float]):
     "Node",
     Projectors.String("node_names", None),
     formatters=(
+        Formatters.bold_cyan_alert(
+            lambda edata: edata.record["Node ID"] == edata.common["self_node"]
+        ),
```
a-spiker: The self-node signal is color-only, so it disappears the moment the output is copied, piped, or viewed in a no-color/--no-color terminal. We already consider that a problem for the principal: on top of the green highlight, the Node ID column prepends a * for the principal (converter at templates.py:157-163, and again in show_roster at templates.py:2519-2525) so a plain-text copy still shows which node is principal. The new cyan self-node highlight has no such fallback, so "which node did asadm connect to?" is unrecoverable from copied output.


(^ is just a suggestion, use whatever reads best as long as it's distinct from *.) Note the current * marker only appears on the two sheets that show a Node ID column, while cyan appears on all of them, so putting the self-marker on node_field also makes coverage consistent. Please also update the color-legend help text to mention the marker, not only the color, so copy/no-color users know what the glyph means.

a-spiker (changes requested):
MInor comments + Existing comments

lib/view/view.py
Alb0t: Added `_CollectinfoSnapshot.get_self_node()` returning `""` (a capture has no self node), mirroring how `get_expected_principal` is dual-implemented. Also added a compat test that calls `CliView._common` against a mock spec'd to `_CollectinfoSnapshot`, so removing the method fails CI rather than only failing at runtime in `asadm -cf`.

lib/view/templates.py
Alb0t: Reordered `green_alert` before `bold_cyan_alert` at all three sites so principal wins the first-match formatter selection. There's now a behavioral test asserting green when a node is both principal and self; I verified it catches the bug by temporarily re-introducing the inverted order (test fails with `'bold-cyan-alert' != 'green-alert'`).

test/unit/view/test_view.py
Alb0t: Added `NodeHighlightingTest` in test_templates.py, which renders the real `info_network_sheet` through the JSON renderer and asserts the actual `format` tags and converted values: green + `*` on principal, cyan + `@` on self, nothing on plain nodes, precedence when both, and no markers when `self_node` is empty. Also added `test_get_self_node` (localhost → node_id) and `test_get_self_node_returns_empty_when_no_localhost` next to `test_get_expected_principal`.

test/unit/view/test_view.py
Alb0t: Added `assertIn("10.0 %", output)` so the secondary-index projector is pinned alongside the primary.

lib/live_cluster/client/cluster.py
Alb0t: Switched to `node.is_localhost()` and added the comment noting first-match/dict-iteration-order behavior when multiple nodes are flagged.

lib/view/templates.py
Alb0t: Removed. The new shmem Div numerators now use only the consolidated 7.0+ names (`index_used_bytes` / `sindex_used_bytes`), consistent with the summary pairing in lib/utils/common.py.

lib/view/templates.py
Alb0t: Went with `@` for the self node. Per your consistency note, the marker now applies to the Node *name* column on every sheet (not just the two Node ID sheets) via a shared `_node_marker` converter — principal names get `*`, self gets `@`, and it degrades gracefully on the one sheet without a Node ID column (`node_info_responses`). The help legend documents both the colors and the glyphs.

a-spiker (approved):
LGTM

PR #516, opened 2026-07-14
fix: TOOLS-3949 asadm execute-mode dispatch (double-run, batch truncation, ';' parsing, Ctrl-C)

a-spiker:
## What

Fixes four related defects in the `-e` / execute-mode command dispatch path in `asadm.py`. Details and full repro/verification in [TOOLS-3949](https://aerospike.atlassian.net/browse/TOOLS-3949).

## Bugs fixed

1. **`do_*` commands run twice in `-e` mode.** `asadm -e cake` ran the command once explicitly (`await shell.onecmd(line)`) and again via `cmdloop`. Only affected shell-local `do_*` commands (`cake`, `exit`, `quit`, `EOF`); `cake` was the only visible case. Removed the redundant call.

2. **`precmd` truncated a command batch at the first shell-local command.** `-e "cake; show config"` ran only `cake` and silently dropped `show config`. Non-terminator shell-local commands now dispatch inline and the batch loop continues; `exit`/`quit`/`EOF` keep the intentional early return.

3. **Semicolon only separated commands when followed by whitespace.** `-e "info;show config"` failed with `command not found`, and a quoted `;` embedded in a longer token (e.g. `"err;"`) was wrongly split. Removed `;` from `shlex` wordchars so an unquoted `;` always separates and a quoted `;` embedded in a longer token stays literal. Known limitation: a value consisting solely of `;` (`';'`, `";"`, `\;`) still splits; this is pinned by tests and a universal quote-tracking fix is deferred to a follow-up ticket.

4. **Ctrl-C during `-e` re-ran the command instead of exiting.** `cmdloop` recursed on `KeyboardInterrupt`/`SystemExit` even in single-command mode. Made the retry iterative; execute mode now re-raises and exits 130.

## Behavior change (release note)

Unquoted `;` now always separates commands. Unquoted values containing `;`, such as `asinfo -v set-config:context=service;proto-fd-max=9999` or equivalent `manage config` values, are silently split at the `;` and the resulting pieces are executed as separate commands, so the intended command only partially executes. The old parser preserved the `;` when it was not followed by a space. Quote the value (`asinfo -v "set-config:context=service;proto-fd-max=9999"`) to keep the `;` literal.

## Verification

- `-e "cake"`: output once, exit 0
- `-e "cake; show config"`, `-e "show config;cake;info network"`: all commands run, in order, once each
- `-e "info;show config"`: both run
- `-e "info network; exit; show config"`: `exit` terminates the batch (unchanged)
- Quoted `;` embedded in a token preserved: `"set-config:context=service;proto-fd-max=9999"`, `'p;w'`, `"err;"`; a standalone quoted/escaped `;` (`';'`, `";"`, `\;`) still splits (known limitation, pinned by tests)
- SIGINT during `-e cake`: exits 130, no restart
- Interactive mode (piped stdin) works; `test/unit/test_aerospike_shell.py` (37 tests) pass

[TOOLS-3949]: https://aerospike.atlassian.net/browse/TOOLS-3949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/asadm.py
+++ b/asadm.py
@@ -364,32 +364,22 @@ class AerospikeShell(cmd.Cmd, AsyncObject):
         #       new parser or correct shlex behavior.
         commands = []
         command = []
-        build_token = ""
 
         # Maybe someday we should not allow most of the characters below without
         # quotes surrounding them.  These characters below define what can be in
         # an unquotes string.
-        lexer.wordchars += r"`~!@#$;%^&*()_-+={}[]|:<>,./\?"
+        lexer.wordchars += r"`~!@#$%^&*()_-+={}[]|:<>,./\?"
         lexer.escapedquotes += "'"
 
         try:
             for token in lexer:
-                build_token += token
-
                 if token == ";":
                     if command:
                         commands.append(command)
                         command = []
-                elif token.endswith(";"):
-                    command.append(build_token[:-1])
-                    commands.append(command)
-                    command = []
                 else:
-                    command.append(build_token)
-                build_token = ""
+                    command.append(token)
             else:
-                if build_token:
-                    command.append(build_token)
                 if command:
                     commands.append(command)
 
@@ -461,7 +451,11 @@ class AerospikeShell(cmd.Cmd, AsyncObject):
 
         for line in lines:
             if line[0] in self.commands:
-                return " ".join(line)
+                if line[0] in ("exit", "quit", "EOF"):
+                    return " ".join(line)
+
+                await self.onecmd(" ".join(line))
+                continue
 
             if len(lines) > max_commands_to_print_header:
                 if len(line) > 1 and any(
@@ -923,7 +917,6 @@ async def main():
                 command_index_to_print_from=command_index_to_print_from,
             )
 
-            await shell.onecmd(line)
             func = shell.onecmd
             args = (line,)
 
@@ -972,21 +965,23 @@ async def cmdloop(
     use_yappi: bool,
     single_command: bool,
 ):
-    try:
-        if use_yappi:
-            yappi.start()
-            await func(*args)
-            yappi.get_func_stats().print_all()
-        else:
-            await func(*args)
-    except (KeyboardInterrupt, SystemExit):
-        if not single_command:
+    while True:
+        try:
+            if use_yappi:
+                yappi.start()
+                await func(*args)
+                yappi.get_func_stats().print_all()
+            else:
+                await func(*args)
+            return
+        except (KeyboardInterrupt, SystemExit):
+            if single_command:
+                raise
             shell.intro = (
                 terminal.fg_red()
                 + "\nTo exit asadm utility please run the 'exit' command."
                 + terminal.fg_clear()
             )
-        await cmdloop(shell, func, args, use_yappi, single_command)
 
 
 def parse_commands(file):
--- a/asadm.py
+++ b/asadm.py
@@ -369,7 +369,7 @@ class AerospikeShell(cmd.Cmd, AsyncObject):
 
         # Maybe someday we should not allow most of the characters below without
         # quotes surrounding them.  These characters below define what can be in
-        # an unquotes string.
+        # an unquoted string.
         lexer.wordchars += r"`~!@#$%^&*()_-+={}[]|:<>,./\?"
         lexer.escapedquotes += "'"
 
--- a/asadm.py
+++ b/asadm.py
@@ -92,6 +92,8 @@ CMD_FILE_MULTI_LINE_COMMENT_END = "*/"
 
 MULTILEVEL_COMMANDS = ["show", "info", "manage"]
 
+TERMINATOR_COMMANDS = ("exit", "quit", "EOF")
+
 DEFAULT_PROMPT = "Admin> "
 PRIVILEGED_PROMPT = "Admin+> "
 ADMIN_DEFAULT_PROMPT = "ADMIN> "
@@ -379,10 +381,9 @@ class AerospikeShell(cmd.Cmd, AsyncObject):
                         command = []
                 else:
                     command.append(token)
-            else:
-                if command:
-                    commands.append(command)
 
+            if command:
+                commands.append(command)
         except ValueError as e:
             raise ShellException(e)
 
@@ -450,11 +451,14 @@ class AerospikeShell(cmd.Cmd, AsyncObject):
             return ""
 
         for line in lines:
-            if line[0] in self.commands:
-                if line[0] in ("exit", "quit", "EOF"):
-                    return " ".join(line)
+            if line[0] in TERMINATOR_COMMANDS:
+                return " ".join(line)
 
-                await self.onecmd(" ".join(line))
+            if line[0] in self.commands:
+                try:
+                    await self.onecmd(" ".join(line))
+                except Exception as e:
+                    logger.error(e)
                 continue
 
             if len(lines) > max_commands_to_print_header:
@@ -503,13 +507,14 @@ class AerospikeShell(cmd.Cmd, AsyncObject):
                 """
                 Interrupt in the middle of executing a command.
                 """
-                pass
+                if self.execute_only_mode:
+                    raise KeyboardInterrupt
             except Exception as e:
                 logger.error(e)
             finally:
                 for signal in signals:
                     loop.remove_signal_handler(signal)
-        return ""  # line was handled by execute
+        return ""
 
     # overloaded to support async
     async def onecmd(self, line):
@@ -872,80 +877,76 @@ async def main():
     args = ()
     single_command = True
     real_stdout = sys.stdout
-    if not execute_only_mode:
-        if not shell.connected:
-            sys.exit(1)
+    f = None
 
-        func = shell.cmdloop
-        single_command = False
-
-    else:
-        commands_arg = cli_args.execute
-        max_commands_to_print_header = 1
-        command_index_to_print_from = 1
-        if os.path.isfile(commands_arg):
-            commands_arg = parse_commands(commands_arg)
-            max_commands_to_print_header = 0
-            command_index_to_print_from = 0
-
-        if cli_args.out_file:
-            try:
-                f = open(str(cli_args.out_file), "w")
-                sys.stdout = f
-                disable_coloring()
-                max_commands_to_print_header = 0
-                command_index_to_print_from = 0
-            except Exception as e:
-                print(e)
-
-        def cleanup():
-            try:
-                sys.stdout = real_stdout
-                if f:
-                    f.close()
-            except Exception:
-                pass
-
-        if shell.connected:
-            # Print admin port visual cue message in execute mode
-            if shell._has_admin_nodes():
-                print(ADMIN_PORT_VISUAL_CUE_MSG)
-
-            line = await shell.precmd(
-                commands_arg,
-                max_commands_to_print_header=max_commands_to_print_header,
-                command_index_to_print_from=command_index_to_print_from,
-            )
-
-            func = shell.onecmd
-            args = (line,)
-
-        else:
-            if "collectinfo" in commands_arg:
-                logger.warning(
-                    "Collecting only System data. Not able to connect any cluster with "
-                    + str(seeds)
-                    + "."
-                )
-
-                func = common.collect_sys_info(port=cli_args.port)
-
-                cleanup()
-                sys.exit(1)
-
-            cleanup()
-
-    if func:
-        await cmdloop(shell, func, args, use_yappi, single_command)
-
-    await shell.close()
-
-    try:
+    def cleanup():
         sys.stdout = real_stdout
         if f:
-            f.close()
-    except Exception:
-        pass
+            try:
+                f.close()
+            except OSError as e:
+                logger.error(e)
+
+    try:
+        if not execute_only_mode:
+            if not shell.connected:
+                sys.exit(1)
+
+            func = shell.cmdloop
+            single_command = False
+
+        else:
+            commands_arg = cli_args.execute
+            max_commands_to_print_header = 1
+            command_index_to_print_from = 1
+            if os.path.isfile(commands_arg):
+                commands_arg = parse_commands(commands_arg)
+                max_commands_to_print_header = 0
+                command_index_to_print_from = 0
+
+            if cli_args.out_file:
+                try:
+                    f = open(str(cli_args.out_file), "w")
+                    sys.stdout = f
+                    disable_coloring()
+                    max_commands_to_print_header = 0
+                    command_index_to_print_from = 0
+                except Exception as e:
+                    print(e)
+
+            if shell.connected:
+                # Print admin port visual cue message in execute mode
+                if shell._has_admin_nodes():
+                    print(ADMIN_PORT_VISUAL_CUE_MSG)
+
+                line = await shell.precmd(
+                    commands_arg,
+                    max_commands_to_print_header=max_commands_to_print_header,
+                    command_index_to_print_from=command_index_to_print_from,
+                )
+
+                func = shell.onecmd
+                args = (line,)
+
+            else:
+                if "collectinfo" in commands_arg:
+                    logger.warning(
+                        "Collecting only System data. Not able to connect any cluster with "
+                        + str(seeds)
+                        + "."
+                    )
+
+                    common.collect_sys_info(port=cli_args.port)
+
+                    sys.exit(1)
+
+        if func:
+            await cmdloop(shell, func, args, use_yappi, single_command)
+    finally:
+        try:
+            await shell.close()
+        finally:
+            cleanup()
 
     sys.exit(get_exit_code())
 
--- a/test/unit/test_aerospike_shell.py
+++ b/test/unit/test_aerospike_shell.py
@@ -519,6 +519,7 @@ class CleanLineTest(unittest.TestCase):
         """Known limitation: posix shlex strips quotes before we see the token."""
         self.assertEqual(self.clean("info ';'"), [["info"]])
         self.assertEqual(self.clean('info ";"'), [["info"]])
+        self.assertEqual(self.clean(r"info \;"), [["info"]])
         self.assertEqual(
             self.clean("grep -s ';' statistics"), [["grep", "-s"], ["statistics"]]
         )
--- a/test/unit/test_aerospike_shell.py
+++ b/test/unit/test_aerospike_shell.py
@@ -574,5 +574,36 @@ class PrecmdDispatchTest(unittest.IsolatedAsyncioTestCase):
         shell.ctrl.execute.assert_not_called()
 
 
+class CmdloopTest(unittest.IsolatedAsyncioTestCase):
+    """Module-level cmdloop re-raises interrupts in single-command (execute)
+    mode and retries func, resetting shell.intro, in interactive mode."""
+
+    async def test_single_command_reraises_keyboard_interrupt(self):
+        func = AsyncMock(side_effect=KeyboardInterrupt)
+        with self.assertRaises(KeyboardInterrupt):
+            await asadm.cmdloop(Mock(), func, (), False, True)
+        func.assert_awaited_once()
+
+    async def test_single_command_reraises_system_exit(self):
+        func = AsyncMock(side_effect=SystemExit)
+        with self.assertRaises(SystemExit):
+            await asadm.cmdloop(Mock(), func, (), False, True)
+        func.assert_awaited_once()
+
+    async def test_normal_completion_runs_once(self):
+        func = AsyncMock(return_value=None)
+        await asadm.cmdloop(Mock(), func, ("line",), False, False)
+        func.assert_awaited_once_with("line")
+
+    async def test_interactive_retries_and_sets_intro(self):
+        func = AsyncMock(side_effect=[KeyboardInterrupt(), None])
+        shell = Mock()
+        await asadm.cmdloop(shell, func, (), False, False)
+        self.assertEqual(func.await_count, 2)
+        self.assertIn(
+            "To exit asadm utility please run the 'exit' command", shell.intro
+        )
+
+
 if __name__ == "__main__":
     unittest.main()
--- a/test/unit/test_aerospike_shell.py
+++ b/test/unit/test_aerospike_shell.py
@@ -705,7 +705,10 @@ class ExecuteModeDoubleRunTest(unittest.IsolatedAsyncioTestCase):
     async def test_emptyline_is_noop(self):
         """Stock cmd.Cmd re-runs lastcmd on a blank line; the override must stay."""
         shell = object.__new__(AerospikeShell)
+        shell.lastcmd = "info network"
+        shell.onecmd = Mock()
         self.assertIsNone(shell.emptyline())
+        shell.onecmd.assert_not_called()
 
 
 if __name__ == "__main__":
--- a/test/unit/test_aerospike_shell.py
+++ b/test/unit/test_aerospike_shell.py
@@ -3,6 +3,7 @@ import asadm
 
 import asyncio
 import unittest
+from types import SimpleNamespace
 from unittest.mock import AsyncMock, Mock, patch, call
 from lib.base_controller import ShellException
 from lib.utils import async_object
@@ -473,8 +474,7 @@ class AdminHomeDirTest(unittest.IsolatedAsyncioTestCase):
 
 
 class CleanLineTest(unittest.TestCase):
-    """clean_line does not touch instance state, so exercise it directly on a
-    bare instance to avoid the async cluster-connect setup."""
+    """Bare instance skips the async cluster-connect setup."""
 
     def clean(self, line):
         shell = object.__new__(AerospikeShell)
@@ -497,8 +497,6 @@ class CleanLineTest(unittest.TestCase):
         )
 
     def test_semicolon_without_spaces_splits(self):
-        """An unquoted ';' separates commands even without surrounding
-        whitespace - ';' was removed from the lexer wordchars."""
         self.assertEqual(
             self.clean("show config;show statistics"),
             [["show", "config"], ["show", "statistics"]],
@@ -513,9 +511,17 @@ class CleanLineTest(unittest.TestCase):
             [["show", "config"], ["show", "statistics"]],
         )
 
-    def test_quoted_semicolon_is_literal(self):
-        """A quoted ';' stays inside its token and does not split the command."""
+    def test_quoted_semicolon_embedded_in_token_is_literal(self):
         self.assertEqual(self.clean("info 'a;b'"), [["info", "a;b"]])
+        self.assertEqual(self.clean("info foo';'bar"), [["info", "foo;bar"]])
+
+    def test_standalone_quoted_semicolon_still_splits(self):
+        """Known limitation: posix shlex strips quotes before we see the token."""
+        self.assertEqual(self.clean("info ';'"), [["info"]])
+        self.assertEqual(self.clean('info ";"'), [["info"]])
+        self.assertEqual(
+            self.clean("grep -s ';' statistics"), [["grep", "-s"], ["statistics"]]
+        )
 
     def test_unterminated_quote_raises_shell_exception(self):
         with self.assertRaises(ShellException):
@@ -523,12 +529,10 @@ class CleanLineTest(unittest.TestCase):
 
 
 class PrecmdDispatchTest(unittest.IsolatedAsyncioTestCase):
-    """precmd routes do_* commands through onecmd and everything else through
-    ctrl.execute, running every command in a multi-command line."""
-
     def make_shell(self):
         shell = object.__new__(AerospikeShell)
-        shell.commands = {"exit", "quit", "EOF", "cake"}
+        shell.commands = set(asadm.TERMINATOR_COMMANDS) | {"cake"}
+        shell.execute_only_mode = False
         shell.ctrl = Mock()
         shell.ctrl.execute = AsyncMock(return_value="")
         shell.onecmd = AsyncMock(return_value=None)
@@ -573,11 +577,44 @@ class PrecmdDispatchTest(unittest.IsolatedAsyncioTestCase):
         shell.onecmd.assert_awaited_once_with("cake")
         shell.ctrl.execute.assert_not_called()
 
+    async def test_ctrl_execute_failure_is_logged_and_batch_continues(self):
+        shell = self.make_shell()
+        shell.ctrl.execute = AsyncMock(side_effect=[Exception("boom"), ""])
+        with patch("asadm.asyncio.get_event_loop", return_value=Mock()):
+            with patch("asadm.logger") as mock_logger:
+                result = await shell.precmd("info network ; info statistics")
+        self.assertEqual(result, "")
+        self.assertEqual(shell.ctrl.execute.await_count, 2)
+        mock_logger.error.assert_called_once()
+
+    async def test_inline_do_command_failure_is_logged_and_batch_continues(self):
+        shell = self.make_shell()
+        shell.onecmd = AsyncMock(side_effect=Exception("boom"))
+        with patch("asadm.asyncio.get_event_loop", return_value=Mock()):
+            with patch("asadm.logger") as mock_logger:
+                result = await shell.precmd("cake ; info network")
+        self.assertEqual(result, "")
+        mock_logger.error.assert_called_once()
+        shell.ctrl.execute.assert_called_once_with(["info", "network"])
+
+    async def test_cancel_reraises_keyboard_interrupt_in_execute_mode(self):
+        shell = self.make_shell()
+        shell.execute_only_mode = True
+        shell.ctrl.execute = AsyncMock(side_effect=asyncio.CancelledError)
+        with patch("asadm.asyncio.get_event_loop", return_value=Mock()):
+            with self.assertRaises(KeyboardInterrupt):
+                await shell.precmd("info network")
+
+    async def test_cancel_is_swallowed_in_interactive_mode(self):
+        shell = self.make_shell()
+        shell.execute_only_mode = False
+        shell.ctrl.execute = AsyncMock(side_effect=asyncio.CancelledError)
+        with patch("asadm.asyncio.get_event_loop", return_value=Mock()):
+            result = await shell.precmd("info network")
+        self.assertEqual(result, "")
+
 
 class CmdloopTest(unittest.IsolatedAsyncioTestCase):
-    """Module-level cmdloop re-raises interrupts in single-command (execute)
-    mode and retries func, resetting shell.intro, in interactive mode."""
-
     async def test_single_command_reraises_keyboard_interrupt(self):
         func = AsyncMock(side_effect=KeyboardInterrupt)
         with self.assertRaises(KeyboardInterrupt):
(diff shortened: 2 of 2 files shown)


dwelch-spike:
**PR Review (version 2026.07.06)** — re-review at f9cc7d5

**Promises:**
1. Shell-local `do_*` commands (`cake`, `exit`, `quit`, `EOF`) run exactly once in `-e` execute mode.
2. A multi-command `-e` batch continues past non-terminator shell-local commands; `exit`/`quit`/`EOF` still terminate the batch.
3. An unquoted `;` always separates commands; a quoted `;` embedded in a token is literal (narrowed by agreement — standalone quoted/escaped `;` is a pinned, documented limitation).
4. Ctrl-C/SystemExit during `-e` propagates so asadm exits (130 for SIGINT) instead of re-running; interactive retry is iterative.
5. Interactive-mode behavior otherwise unchanged.
6. The new unit tests are regression tests for each of the four bugs.

| Gate | Status |
|------|--------|
| Contract | FAIL (narrowly) — code promises all verified; the PR body's "Behavior change (release note)" section still says "now require quotes" instead of the agreed "silently split and partially execute" wording |
| Impacted surface | PASS — main() restructure traced mode-by-mode (6 paths) as control-flow equivalent; it also fixes a latent UnboundLocalError in the old cleanup |
| Failure paths | PASS — Ctrl-C re-raise, nested finally, signal-handler removal ordering, unconnected-shell close(), and exception-masking paths all traced clean |
| Evidence | PASS — 37/37 at HEAD; bug-1, Ctrl-C, containment, and iterative-retry pins all mutation-verified genuine; one minor weak pin (emptyline) posted inline |
| Quality | PASS |

**Prior findings:**
| Prior finding | Status |
|---------------|--------|
| Control Flow: execute-mode Ctrl-C swallowed in precmd (asadm.py) | Addressed in fd3b5ce — verified + mutation-pinned |
| Complexity: cleanup skipped on raise path (asadm.py) | Addressed in fd3b5ce — try/finally verified on all six mode paths |
| Data Flow: unquoted `;` silent partial execution (asadm.py) | **Partially addressed** — typo fixed, tripwire-by-tests verified, batch pre-validation deferred to follow-up ticket (agreed); release-note reword in PR body still pending |
| Complexity: magic terminator tuple / stop flag / stale comment (asadm.py) | Addressed in fd3b5ce — `TERMINATOR_COMMANDS` hoisted; author's stop-protocol rebuttal verified correct |
| Control Flow: uncontained inline dispatch (asadm.py) | Addressed in fd3b5ce — containment + both tests, mutation-pinned |
| Style: for/else without break (asadm.py) | Addressed in fd3b5ce |
| Security: standalone quoted `;` still splits (tests) | **Partially addressed** — narrowed claim pinned by tests (all three spellings), universal fix deferred (agreed); release-note narrowing in PR body still pending |
| Tests: bug 1 (double-run) unpinned | Addressed in fd3b5ce — `ExecuteModeDoubleRunTest` mutation-verified: re-adding the deleted line flips exactly that test red |
| Tests: iterative conversion unpinned | Addressed in fd3b5ce — 2000-interrupt test fails with RecursionError against the recursive implementation (mutation-verified) |

**1 new finding** posted inline (0 Critical, 0 Important, 1 Minor): `test_emptyline_is_noop` does not actually pin the override (mutation-verified — deleting the override leaves 37/37 green).

Remaining before approval: (1) reword the PR body's release-note section as agreed in the two open threads; (2) optionally strengthen the emptyline test. CI is green on f9cc7d5. Codecov's 65.71% patch coverage is dominated by mechanically relocated main() lines; all four core fixes are covered and mutation-verified.

asadm.py
```diff
@@ -972,21 +965,23 @@ async def cmdloop(
...
+        try:
+            if use_yappi:
+                yappi.start()
+                await func(*args)
+                yappi.get_func_stats().print_all()
+            else:
+                await func(*args)
+            return
+        except (KeyboardInterrupt, SystemExit):
+            if single_command:
```
dwelch-spike, line 979: From Control Flow agent:

Severity: important
Problem Confidence: high
Solution Confidence: medium

## Problem

The PR claims "Ctrl-C during `-e` exits 130," but this `raise` only guards a window that is nearly empty. In execute mode, every real command runs inside `shell.precmd(commands_arg, ...)` (asadm.py:914) — *before* `cmdloop` is entered with `func = shell.onecmd` and `args` already reduced to `""` or `"exit"`. Inside `precmd`, SIGINT is bound to `task.cancel` (asadm.py:488-489), the resulting `CancelledError` is swallowed (`except asyncio.CancelledError: pass`, asadm.py:502-506), and the `for line in lines` loop proceeds to the NEXT command in the batch. Consequences:

1. `asadm -e "info; collectinfo"` + Ctrl-C during `info` cancels only `info` and still runs `collectinfo`.
2. The run then exits via `sys.exit(get_exit_code())` (asadm.py:950) — nothing sets a non-zero code on the cancel path (lib/utils/logger.py only sets it on error/critical), so an aborted run reports success (0), not 130. A wrapping script cannot distinguish an interrupted run from a completed one.

Promise 4 therefore holds only in the sub-millisecond gaps between commands, not during command execution — which is virtually all of `-e` wall-clock time.

## Reasoning Chain

1. Traced execute-mode wiring post-diff: main():914 runs the whole batch via precmd; cmdloop later dispatches `onecmd("")` (a no-op — `emptyline` is overridden at asadm.py:576) or `onecmd("exit")`. The protected region excludes command execution by construction.
2. `loop.add_signal_handler(SIGINT, task.cancel)` replaces the default SIGINT handler, so no KeyboardInterrupt is raised while a command runs; Ctrl-C manifests solely as `CancelledError` on the awaited task.
3. That exception is caught and discarded; control falls through `finally` (handler removal) into the next loop iteration. The comment at asadm.py:482-487 justifies this for interactive `watch`; execute mode inherited it incidentally.
4. Two independent traces reached the same conclusion. Settling check: `asadm -e "watch 3 5 info; info network"` + Ctrl-C.

## Solution

Make the cancellation path mode-aware — `self.execute_only_mode` already exists on the shell:

```python
except asyncio.CancelledError:
    if self.execute_only_mode:
        raise KeyboardInterrupt
```

KeyboardInterrupt then propagates from main():914 (no enclosing try) through `asyncio.run` to the `__main__` handler (asadm.py:1025-1026) → `sys.exit(130)` — exactly the promised behavior. Alternatively, skip installing the cancel handlers in execute mode so SIGINT raises natively. Unchecked edges: whether any collectinfo flow relies on Ctrl-C skipping a hung node while the run continues, and asyncio teardown of in-flight cluster connections on the re-raise path.

asadm.py
```diff
@@ -972,21 +965,23 @@ async def cmdloop(
...
+                raise
```
dwelch-spike, line 980: From Complexity agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

This `raise` propagates through main() past `await shell.close()` (asadm.py:941) and past the stdout-restore/`f.close()` block (asadm.py:943-948). main() already contains two copies of that cleanup — the nested `cleanup()` at ~901-907 and the inline copy at 943-948 — and this PR adds a third control-flow path that skips both. Practical damage is limited (the process is exiting; CPython flushes streams at shutdown), but with `-o out_file` the user's redirected output now depends on interpreter-shutdown finalization, and `shell.close()` (client connection teardown) is silently skipped on the exact path this PR introduces.

## Reasoning Chain

1. Contract promise 4 requires the exception to propagate in execute mode; the raise does that.
2. Trace the propagation: main():939 `await cmdloop(...)` → raises → lines 941-948 never run → caught at the module entrypoint (asadm.py:1025) → `sys.exit(130)`.
3. So shell.close(), sys.stdout restoration, and f.close() are all skipped on interrupt — and the duplicated cleanup blocks show main() was written assuming straight-line flow.

## Solution

Wrap the tail of main() in try/finally:

```python
try:
    if func:
        await cmdloop(shell, func, args, use_yappi, single_command)
finally:
    await shell.close()
    sys.stdout = real_stdout
    ...
```

This changes nothing on the happy path and makes the raise path safe by construction. Consider deduplicating the two existing cleanup blocks while there.

asadm.py
```diff
@@ -364,32 +364,22 @@ def clean_line(self, line):
         #       new parser or correct shlex behavior.
         commands = []
         command = []
-        build_token = ""
 
         # Maybe someday we should not allow most of the characters below without
         # quotes surrounding them.  These characters below define what can be in
         # an unquotes string.
-        lexer.wordchars += r"`~!@#$;%^&*()_-+={}[]|:<>,./\?"
+        lexer.wordchars += r"`~!@#$%^&*()_-+={}[]|:<>,./\?"
```
dwelch-spike, line 373: From Data Flow agent:

Severity: important
Problem Confidence: high
Solution Confidence: low

## Problem

The release note frames the behavior change as "unquoted values containing ';' now require quotes." The actual failure mode for existing scripts is worse than a rejection: the value is silently split, and the *truncated head executes as a valid — possibly far more destructive — command* before the tail errors out. Empirically, with the exact lexer config at HEAD:

- old: `asinfo -v truncate:namespace=test;set=s1` → `[['asinfo','-v','truncate:namespace=test;set=s1']]`
- new: → `[['asinfo','-v','truncate:namespace=test'], ['set=s1']]`

precmd dispatches these in order: `asinfo -v truncate:namespace=test` runs first — truncating the entire namespace instead of one set — then `set=s1` fails with `logger.error` (exit code 2). The error arrives after the mutating head has already executed. This is the shape of real scripted usage: callers single-quote the whole `-e '...'` argument for the OS shell, leaving inner values unquoted, and command files fed through `parse_commands` contain raw unquoted lines.

Separately, nothing at this line documents that ';' is *deliberately* absent from wordchars. The next person debugging shlex mangling of asinfo values will "fix" it by adding ';' back and silently reintroduce bug 3 — the only tripwire is a unit test three directories away. (Also: "unquotes" → "unquoted" in the comment above.)

## Reasoning Chain

1. Ran old and new `clean_line` side by side on the truncate example; the old lexer kept one token, the new one splits mid-word.
2. Traced precmd's loop: commands execute sequentially; the head runs via ctrl.execute before the tail fragment errors, so the exit-code-2 safety net fires only after the mutating command completed.
3. Checked other ';' consumers for containment: lib's `split(";")` sites (node.py, get_controller.py, collectinfo_controller.py, common.py) parse info-protocol *responses*, and asinfo mode (`-a`) has its own regex splitter (asadm.py:699) — both untouched. The blast radius is confined to shell-parsed command lines, but within it the consequence is partial execution, not an error.

## Solution

The lexer change is the point of the PR; reverting it reintroduces bug 3. Missing is a guard between "split happened" and "head executes," plus documentation. In increasing cost: (a) add a one-line comment here — "';' is deliberately NOT in wordchars: an unquoted ';' must tokenize alone so it separates commands; a quoted ';' remains literal" — and make the release note say unquoted ';' values now *silently split and partially execute*, naming asinfo/manage as the risky commands; (b) in precmd, when clean_line yields multiple commands and a later fragment's head is not a known command, refuse the whole batch before executing anything — cheap, and turns silent truncation into a hard error; (c) provenance-aware lexing (see the quoted-';' finding on the tests). Option (b) uniformly covers value-bearing arguments I did not audit individually (e.g. manage config string values).

asadm.py
```diff
@@ -461,7 +451,11 @@ async def precmd(
 
         for line in lines:
             if line[0] in self.commands:
-                return " ".join(line)
+                if line[0] in ("exit", "quit", "EOF"):
```
dwelch-spike: From Complexity agent:

Severity: important
Problem Confidence: high
Solution Confidence: medium

## Problem

precmd() now has three dispatch modes and documents none of them: (a) controller commands run inline via `ctrl.execute`, (b) non-terminator `do_*` commands run inline via `await self.onecmd(...)` with the return value thrown away, (c) `exit`/`quit`/`EOF` are returned as a string for the caller to dispatch. The magic tuple `("exit", "quit", "EOF")` carries the entire reason this works, and that reason appears nowhere in the code.

The reason matters: `do_exit`/`do_quit`/`do_EOF` return a stop flag that is only honored when the caller does `stop = await self.onecmd(line)` (asadm.py:428). The inline dispatch at line 457 silently discards onecmd's return value. Meanwhile `self.commands` — the set that decides what is shell-local at all — is built dynamically by a `do_*` reflection scan (asadm.py:274-281). Two sources of truth for one concept: any future `do_*` command that returns True (or an exit-style alias not added to the tuple) gets its stop signal silently swallowed and the batch keeps going. The aliasing pattern already exists — `do_quit`/`do_EOF` are just delegations to `do_exit`.

And the trailing comment is now wrong: `return ""  # line was handled by execute` (asadm.py:512) — after this change the line may have been handled by onecmd, not execute.

## Reasoning Chain

1. Old precmd returned the first shell-local command for the cmd framework to dispatch; the new code dispatches non-terminators itself and continues.
2. `await self.onecmd(" ".join(line))` discards the result; do_exit returns True precisely so line 428 can see it.
3. Therefore terminators MUST take path (c) — but that invariant lives only in a hardcoded 3-element tuple with zero commentary, parallel to a reflection-built set that grows automatically.
4. Line 512's comment predates path (b) and now contradicts it.

## Solution

Minimum: hoist the tuple to a named constant (`TERMINATOR_COMMANDS = ("exit", "quit", "EOF")`) with a comment stating the invariant — "these do_* commands return a stop flag that must flow through the caller's `stop = await self.onecmd(line)`; dispatching them inline would discard it" — and fix the line-512 comment. Better: honor the protocol instead of special-casing names — `if await self.onecmd(" ".join(line)): return ""` — which makes any stop-returning do_* terminate the batch automatically; verify no do_* returns a truthy non-stop value and that the interactive path at 427-429 still receives its stop before adopting this.

asadm.py
```diff
@@ -461,7 +451,11 @@ async def precmd(
...
+                    return " ".join(line)
+
+                await self.onecmd(" ".join(line))
```
dwelch-spike: From Control Flow agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

The two dispatch paths inside this loop now have asymmetric failure behavior. Controller commands run under `try: ... except Exception as e: logger.error(e)` (asadm.py:507-508) — a failing command is logged and the batch continues. The new inline shell-local dispatch has no containment: an exception aborts the remaining batch and propagates out of precmd. In interactive mode it escapes `AerospikeShell.cmdloop` (try/finally only), is not caught by the module-level cmdloop (which catches only KeyboardInterrupt/SystemExit), reaches the module entrypoint's `except Exception: pass` (asadm.py:1027-1028), and the session dies silently with exit code 0.

Today the only inline-dispatchable command is `do_cake`, so the trigger surface is small — but the trap is structural and armed for the next `do_*` anyone adds. The new tests don't cover it either: PrecmdDispatchTest's mocks never raise, so neither the log-and-continue policy of the ctrl route nor the abort policy of this route is pinned anywhere.

## Reasoning Chain

1. `self.commands` is built by reflection over `do_*` (asadm.py:276-281); everything outside the terminator tuple takes this uncontained path.
2. Traced the exception route: precmd → AerospikeShell.cmdloop (no except) → module cmdloop (KeyboardInterrupt/SystemExit only) → main():939 unprotected → asyncio.run → `except Exception: pass` → exit 0, no message.
3. The adjacent ctrl.execute branch proves the intended failure policy for a batch member is log-and-continue; this branch silently institutes a second, harsher policy.

## Solution

Mirror the containment: `try: await self.onecmd(...) except Exception as e: logger.error(e)` and continue. Add two tests: (a) ctrl.execute raises on the first command of a batch — assert the second still executes and precmd returns ""; (b) pin the inline-onecmd failure behavior, whichever policy is chosen. (The `except Exception: pass` at the entrypoint turning any crash into exit 0 is pre-existing and out of this PR's scope, but it is what makes this abort silent.)

asadm.py
```diff
@@ -364,32 +364,22 @@ def clean_line(self, line):
...
                         commands.append(command)
                         command = []
-                elif token.endswith(";"):
-                    command.append(build_token[:-1])
-                    commands.append(command)
-                    command = []
                 else:
-                    command.append(build_token)
-                build_token = ""
+                    command.append(token)
```
dwelch-spike, line 383: From Style agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

The rewritten token loop keeps a `for ... else:` with no `break` anywhere in the loop body. A loop-`else` without `break` is unconditionally executed straight-line code dressed up as a conditional epilogue (pylint W0120). It misleads twice: every reader stops to hunt for the `break` that justifies it, and a maintainer who later adds a `break` silently changes whether the trailing `if command: commands.append(command)` flush runs — turning a structural vestige into a latent truncation bug in the very function this PR just fixed for truncation. It was defensible noise when the old build_token bookkeeping lived here; now that the body is four lines, the `else:` is the most confusing thing left in the function.

## Reasoning Chain

1. Python's for/else executes the else suite iff the loop was not exited by `break`; grep of the loop body: no break exists, so the else suite runs on every non-exception path — equivalent to dedented code.
2. This PR rewrote the loop body and its else suite wholesale; keeping the construct encodes a falsehood about the control flow.
3. Exception paths are unaffected either way: a ValueError from the lexer skips both forms identically and lands in the except.

## Solution

Dedent the flush:

```python
for token in lexer:
    if token == ";":
        if command:
            commands.append(command)
            command = []
    else:
        command.append(token)
if command:
    commands.append(command)
```

Zero behavior change on all paths including the exception path; the new CleanLineTest cases cover the trailing-command flush.

test/unit/test_aerospike_shell.py
```diff
@@ -469,3 +470,140 @@ def test_admin_home_skipped_in_execute_mode(self):
...
+    def test_leading_and_trailing_semicolons_ignored(self):
+        self.assertEqual(self.clean(";show config;"), [["show", "config"]])
+
+    def test_multiple_consecutive_semicolons(self):
+        self.assertEqual(
+            self.clean("show config ;; show statistics"),
+            [["show", "config"], ["show", "statistics"]],
+        )
+
+    def test_quoted_semicolon_is_literal(self):
```
dwelch-spike: From Security agent:

Severity: important
Problem Confidence: high
Solution Confidence: low

## Problem

The contract states "a quoted ';' is always literal." It is not. A token consisting of exactly one semicolon is treated as a command separator even when the user quoted or escaped it — verified empirically against the lexer configuration at HEAD:

- `info ';'` → `[['info']]` — the quoted ';' silently vanishes
- `grep -s ';' statistics` → `[['grep', '-s'], ['statistics']]` — two bogus commands
- `grep -s ";" statistics` and `grep -s \; statistics` → same
- Only embedded cases (`'a;b'`, `foo';'bar`) are literal — and that is all this test covers, so the gap is unpinned in either direction.

The docstring ("A quoted ';' stays inside its token and does not split the command") and the PR body's "Quoted ';' preserved" claim more than the code delivers. The failure is silent: the input splits into two wrong commands and both are handed to the dispatcher. The hole itself predates this PR, but the PR's stated contract, release note, and this test's name assert that quoting works — for a value that IS ';', it cannot.

## Reasoning Chain

1. `clean_line` builds `shlex.shlex(line, posix=True)`; POSIX mode strips quotes and resolves escapes before yielding tokens.
2. Therefore the token produced by `';'`, `";"`, or `\;` is the one-character string `;` — bit-for-bit identical to an unquoted separator.
3. The separator test is `token == ";"` on the post-lex string, at which point the quoting information has been irrecoverably destroyed — no tuning of `wordchars` can fix this.
4. Verified by executing the function's exact logic on all three spellings; independently reproduced by three separate traces.

## Solution

Decide the intended behavior first. If a lone quoted ';' argument is out of scope (plausible — grep/filter patterns are the realistic victims), narrow the claim: fix this docstring to "a quoted ';' embedded in a longer token is literal," note the limitation in the release note, and add tests pinning the standalone `';'`, `";"`, and `\;` behavior as known. If the promise must hold universally, the separator decision has to be made while quoting information still exists — e.g. a pre-pass over the raw line tracking quote/escape state that splits on top-level unquoted ';' only, then shlex per segment. Any lexing rework risks disturbing the exotic wordchars set; proposed fixes diverged across reviewers, hence low solution confidence.

test/unit/test_aerospike_shell.py
```diff
@@ -469,3 +470,140 @@ def test_admin_home_skipped_in_execute_mode(self):
...
+    async def test_exit_after_command_stops_and_skips_rest(self):
+        shell = self.make_shell()
+        with patch("asadm.asyncio.get_event_loop", return_value=Mock()):
+            result = await shell.precmd("cake ; exit ; info network")
+        self.assertEqual(result, "exit")
+        shell.onecmd.assert_awaited_once_with("cake")
+        shell.ctrl.execute.assert_not_called()
+
+
+class CmdloopTest(unittest.IsolatedAsyncioTestCase):
```
dwelch-spike: From Tests agent:

Severity: important
Problem Confidence: high
Solution Confidence: medium

## Problem

Bug 1 — the double-run, the headline defect of this PR — has no regression test. The fix is the single deleted line in main() (`await shell.onecmd(line)` before `func = shell.onecmd`), but no test exercises main()'s execute-mode dispatch. Mutation-verified: re-adding that exact line at PR HEAD and re-running `pytest test/unit/test_aerospike_shell.py` leaves **all 29 tests green**. Reintroduce the double dispatch and CI stays green.

CmdloopTest.test_normal_completion_runs_once does not cover it: it asserts cmdloop calls func once, which was never the bug — the double dispatch lived in main(), upstream of cmdloop. The suite tests the two halves (precmd, cmdloop) in isolation; the bug was the wiring.

Related untested invariant: the exactly-once property now also depends on `AerospikeShell.emptyline()` staying a no-op override (asadm.py:576). In `-e "cake"`, precmd runs `onecmd("cake")` inline (setting `lastcmd`), then cmdloop calls `onecmd("")` — with stock `cmd.Cmd.emptyline`, that re-runs `lastcmd`, i.e. the double-run comes back through a different door.

## Reasoning Chain

1. The bug's mechanism required both call sites in main() to fire; only a test that drives main()'s execute path (or an extracted equivalent) can catch its return.
2. The only tests that call `asadm.main()` (AdminHomeDirTest) run with `shell.connected = False`, so control never reaches the dispatch block.
3. Confirmed by mutation: re-inserted the deleted line → 29/29 pass. (For contrast: reverting the precmd, clean_line, or cmdloop fixes each fails 2-3 of the new tests — those three are genuinely pinned.)

## Solution

Add a main()-level test in the AdminHomeDirTest style: patch `asadm.AerospikeShell` to return a mock with `connected = True`, run main() with execute args (`-e "cake"`), stub `precmd` to return `""`, and assert `shell.onecmd` is awaited exactly once across the whole run (with precmd's return value). A second cheap pin: assert `AerospikeShell.emptyline()` returns None without invoking `lastcmd`. The mock plumbing is tedious, which is exactly why this path is worth automating — nobody re-verifies it by hand.

test/unit/test_aerospike_shell.py
```diff
@@ -469,3 +470,140 @@ def test_admin_home_skipped_in_execute_mode(self):
...
+            await asadm.cmdloop(Mock(), func, (), False, True)
+        func.assert_awaited_once()
+
+    async def test_normal_completion_runs_once(self):
+        func = AsyncMock(return_value=None)
+        await asadm.cmdloop(Mock(), func, ("line",), False, False)
+        func.assert_awaited_once_with("line")
+
+    async def test_interactive_retries_and_sets_intro(self):
+        func = AsyncMock(side_effect=[KeyboardInterrupt(), None])
```
dwelch-spike, line 637: From Tests agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

This test does not pin the iterative (bounded-stack) conversion of cmdloop — it passes unchanged against the old recursive implementation. Mutation-verified: reverting cmdloop to the recursive version fails only the two single_command re-raise tests; this one stays green. So the "recursion → iteration" half of bug 4's fix is guarded by nothing; someone could reintroduce recursion in the interactive branch and CI would not notice.

Smaller, related: exit code 130 for SIGINT is claimed in the PR body but only manually verified — the mapping lives in the `if __name__ == "__main__"` guard (asadm.py:1022-1026), unreachable from unit tests. Acceptable for a 3-line guard, but it stays permanently untestable without a subprocess-level test.

## Reasoning Chain

1. Old code recursed once per interrupt in interactive mode; this test's `side_effect=[KeyboardInterrupt(), None]` is exactly one retry — well within stack depth, so recursive and iterative implementations are indistinguishable.
2. Ran this test against the reverted (recursive) cmdloop: passes.
3. The two single_command tests do fail against old code, so bug 4's user-facing half is pinned; only the bounded-stack half is not.

## Solution

Add one test with more interrupts than the recursion limit:

```python
func = AsyncMock(side_effect=[KeyboardInterrupt()] * 2000 + [None])
await asadm.cmdloop(Mock(), func, (), False, False)
self.assertEqual(func.await_count, 2001)
```

Fails with RecursionError against the recursive implementation, passes against the loop — a one-line pin for the bounded-stack claim.

asadm.py
```diff
@@ -972,21 +965,23 @@ async def cmdloop(
...
+            return
+        except (KeyboardInterrupt, SystemExit):
+            if single_command:
+                raise
             shell.intro = (
                 terminal.fg_red()
                 + "\nTo exit asadm utility please run the 'exit' command."
                 + terminal.fg_clear()
             )
-        await cmdloop(shell, func, args, use_yappi, single_command)
```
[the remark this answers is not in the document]
a-spiker: Added `CmdloopTest`: KeyboardInterrupt and SystemExit re-raise in single-command mode, normal completion runs once, interactive retry sets `shell.intro`, plus a 2000-interrupt case pinning the iterative conversion (fails with RecursionError against the recursive implementation).

asadm.py
a-spiker: Implemented the suggested fix: the CancelledError handler in `precmd` now raises KeyboardInterrupt when `self.execute_only_mode` is set, so Ctrl-C aborts the whole batch and exits 130 via the `__main__` handler. Interactive mode keeps the swallow for the `watch` use case. Pinned by `test_cancel_reraises_keyboard_interrupt_in_execute_mode` and `test_cancel_is_swallowed_in_interactive_mode`.

On the unchecked edge: collectinfo only inherited the skip-a-hung-node behavior incidentally, and aborting on Ctrl-C in execute mode is the promised behavior, so no carve-out was added.

asadm.py
a-spiker: Done. `main()`'s tail is wrapped in try/finally, and the two cleanup copies are deduplicated into a single hoisted `cleanup()`. The finally is nested (`try: await shell.close()` / `finally: cleanup()`) so stdout restore and `f.close()` survive even a BaseException out of `close()`. `cleanup()` also no longer swallows close errors: a failed `f.close()` is logged (exit code 2) instead of shipping a truncated `-o` file with exit 0.

asadm.py
a-spiker: Agreed on the failure mode. Taken option (a): the "unquotes" typo is fixed, and the release note will be reworded to say unquoted `;` values silently split and partially execute, naming asinfo/manage as the risky commands. The tripwire also exists as tests: re-adding `;` to wordchars fails the CleanLineTest split cases.

Option (b), refusing the batch when a later fragment's head is not a known command, needs command-tree validation in precmd before anything executes; proposing that as a follow-up ticket rather than growing this PR.

asadm.py
a-spiker: The tuple is hoisted to a module constant `TERMINATOR_COMMANDS`, the terminator check is hoisted above the `self.commands` membership test, and the stale `# line was handled by execute` comment is gone.

The stop-protocol alternative (`if await self.onecmd(...): return ""`) doesn't work here: returning `""` never stops the outer loop, because `onecmd("")` routes to the no-op `emptyline()` and yields `stop=None`. Terminators must return the line so the caller's `stop = await self.onecmd(line)` sees `do_exit`'s True; running `do_exit` inline would also close the controller mid-precmd. That invariant is pinned by `PrecmdDispatchTest` and `test_emptyline_is_noop`.

asadm.py
a-spiker: Done as suggested: the inline onecmd dispatch is wrapped in `try/except Exception: logger.error(e)` and the batch continues. Both suggested tests added: `test_ctrl_execute_failure_is_logged_and_batch_continues` and `test_inline_do_command_failure_is_logged_and_batch_continues`.

asadm.py
a-spiker: Done, the flush is dedented out of the `else`.

test/unit/test_aerospike_shell.py
a-spiker: Adopted the narrowing. The test is renamed `test_quoted_semicolon_embedded_in_token_is_literal`, and a new `test_standalone_quoted_semicolon_still_splits` pins all three spellings (`';'`, `";"`, `\;`) as a documented limitation, with the docstring noting posix shlex strips quotes before tokenizing. The release note will be narrowed to "a quoted `;` embedded in a longer token is literal".

A universal fix needs a quote-tracking pre-pass over the raw line; deferring to the same follow-up ticket as option (b) on the clean_line thread.

test/unit/test_aerospike_shell.py
a-spiker: Added `ExecuteModeDoubleRunTest.test_do_command_dispatched_exactly_once`: drives `main()`'s execute path with a connected mock shell and asserts `onecmd` is awaited exactly once, so re-adding the deleted line flips it red. Also added `test_emptyline_is_noop` to pin the second door.

test/unit/test_aerospike_shell.py
a-spiker: Added exactly the suggested test: 2000 interrupts then None, asserts `await_count == 2001`; it fails with RecursionError against the recursive implementation. Agreed the `__main__` 130 mapping stays untestable without a subprocess test; leaving that as-is.

a-spiker:
All 9 inline findings addressed; the fixes land in the next push:

- Promise 4 implemented: execute-mode CancelledError re-raises as KeyboardInterrupt, Ctrl-C aborts the batch and exits 130. Interactive swallow kept for `watch`. Both pinned by tests.
- Promise 3 narrowed: embedded quoted `;` is literal; standalone quoted/escaped `;` pinned as a known limitation, release note to be reworded (silent split + partial execution, asinfo/manage named).
- Bug 1 is now mutation-pinned via a main()-level test (`ExecuteModeDoubleRunTest`), plus an `emptyline` no-op pin.
- cmdloop retry/re-raise fully tested, including a 2000-interrupt bounded-stack pin.
- main() cleanup deduplicated into one `cleanup()` under try/finally; inline dispatch failure containment mirrors the ctrl.execute path.

test/unit/test_aerospike_shell.py
```diff
@@ -469,3 +471,242 @@ def test_admin_home_skipped_in_execute_mode(self):
...
+        ):
+            with self.assertRaises(SystemExit):
+                await asadm.main()
+
+        shell.onecmd.assert_awaited_once_with("")
+
+    async def test_emptyline_is_noop(self):
+        """Stock cmd.Cmd re-runs lastcmd on a blank line; the override must stay."""
+        shell = object.__new__(AerospikeShell)
+        self.assertIsNone(shell.emptyline())
```
dwelch-spike, line 710: From Tests agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

`test_emptyline_is_noop` does not pin the `emptyline` override. Deleting the override from AerospikeShell leaves all 37 tests green (mutation-verified). Stock `cmd.Cmd.emptyline` is `if self.lastcmd: return self.onecmd(self.lastcmd)`, and `cmd.Cmd.lastcmd` is the class attribute `''`; this test's bare `object.__new__(AerospikeShell)` shell never sets `lastcmd`, so the falsy default makes stock emptyline return None as well — the assertion passes with or without the override. The docstring's claim ("the override must stay") is not enforced.

## Reasoning Chain

1. Mutation: removed the `emptyline` override from asadm.py → full suite: 37 passed, 0 failed.
2. Probe: stock emptyline only re-dispatches when `lastcmd` is truthy; the class default is `''`.
3. Counter-probe: with `shell.lastcmd = "info network"` and a mocked `onecmd`, the pristine override returns None with onecmd uncalled, while the override-deleted mutant calls `onecmd(lastcmd)` — a lastcmd-set variant does distinguish the regression.

## Solution

Exercise the guarded path:

```python
async def test_emptyline_is_noop(self):
    shell = object.__new__(AerospikeShell)
    shell.lastcmd = "info network"  # stock cmd.Cmd would re-dispatch this
    shell.onecmd = Mock()
    self.assertIsNone(shell.emptyline())
    shell.onecmd.assert_not_called()
```

Empirically verified: this variant fails against the override-deleted mutant and passes at HEAD.

asadm.py
dwelch-spike: Resolved by fd3b5ce — verified: the CancelledError handler re-raises KeyboardInterrupt when `execute_only_mode` is set; traced the full propagation (signal handlers removed in precmd's finally before the KI leaves; main()'s new finally runs close+cleanup; `__main__` maps to exit 130). Interactive swallow preserved. Mutation-verified: reverting to `pass` fails exactly `test_cancel_reraises_keyboard_interrupt_in_execute_mode`, and the interactive-swallow test stays green.

asadm.py
dwelch-spike: Resolved by fd3b5ce — verified: main()'s tail is try/finally with the nested `try: await shell.close() / finally: cleanup()`, cleanup deduplicated. Traced all six mode paths (interactive/execute × connected/not, collectinfo, out_file failure) as control-flow equivalent — the restructure also fixes a latent UnboundLocalError in the old interactive-mode cleanup, and `shell.close()` is safe on unconnected shells (swallow-all, idempotent down to Cluster.close). Exit codes preserved on every path.

asadm.py
dwelch-spike: Resolved by fd3b5ce — verified: `TERMINATOR_COMMANDS` module constant, terminator check hoisted above the `self.commands` membership test (semantics-preserving since do_exit/quit/EOF always populate the set), stale comment gone. Your stop-protocol rebuttal is correct — returning "" routes to the no-op emptyline yielding stop=None, and inline do_exit would close the controller mid-precmd; returning the terminator line is the only mechanism that stops cmd.Cmd's loop.

asadm.py
dwelch-spike: Resolved by fd3b5ce — verified: inline onecmd dispatch wrapped in try/except with `logger.error` and the batch continues (KeyboardInterrupt/SystemExit correctly not swallowed). Both tests present; mutation-verified: removing the containment fails exactly `test_inline_do_command_failure_is_logged_and_batch_continues`. Bonus: an inline do_* failure now sets exit code 2 instead of silently exiting 0 via the entrypoint's except-pass.

asadm.py
dwelch-spike: Resolved by fd3b5ce — verified: flush dedented out of the else, still inside the try so a lexer ValueError skips it identically; trailing-command, empty-line, and unterminated-quote behavior unchanged.

test/unit/test_aerospike_shell.py
dwelch-spike: Resolved by fd3b5ce — mutation-verified: re-inserting `await shell.onecmd(line)` in main() flips exactly `test_do_command_dispatched_exactly_once` red (36 others green). Mock fidelity checked against main()'s real call surface — all patched names match, every attribute read on the driven path is supplied, and `_has_admin_nodes` is correctly a sync Mock. Note: the companion `test_emptyline_is_noop` does NOT yet pin the override (bare shell's `lastcmd` is the falsy class default) — new minor comment posted with a one-line strengthening; not blocking this thread since the main() pin is the core concern here.

test/unit/test_aerospike_shell.py
dwelch-spike: Resolved by fd3b5ce — mutation-verified: restoring the recursive cmdloop fails `test_interactive_retry_is_iterative_not_recursive` with RecursionError while the single-retry test stays green. Agreed on leaving the `__main__` 130 mapping to manual/subprocess verification.

asadm.py
dwelch-spike: From Data Flow agent:

Severity: important
Problem Confidence: high
Solution Confidence: high
Status: partially addressed at f9cc7d5 — code items done, PR-body edit pending

## Problem

The code-side items are verified: "unquoted" typo fixed, and the tripwire-by-tests claim holds (re-adding `;` to wordchars fails the CleanLineTest split cases). Deferring batch pre-validation (option b) to a follow-up ticket is a reasonable scope call. However, the PR body's "Behavior change (release note)" section still reads "now require quotes around the value" — the agreed rewording (unquoted `;` values silently split and *partially execute*, naming asinfo/manage) has not landed yet.

## Reasoning Chain

1. Checked the PR body at f9cc7d5: the release-note section is unchanged from the original text.
2. This thread's reply committed to the reword; the release note is the only remaining delta between the agreed fix and the PR.

## Solution

Edit the PR body's release-note section as agreed (and link the follow-up ticket for option (b) once filed). This thread can be resolved after that edit — no further code change needed.

test/unit/test_aerospike_shell.py
dwelch-spike: From Security agent:

Severity: important
Problem Confidence: high
Solution Confidence: high
Status: partially addressed at f9cc7d5 — tests done, PR-body edit pending

## Problem

The narrowing is implemented and verified: `test_quoted_semicolon_embedded_in_token_is_literal` covers both embedded spellings, and `test_standalone_quoted_semicolon_still_splits` pins all three standalone spellings (`';'`, `";"`, `\;`) as a documented limitation with an accurate docstring. Deferring the universal quote-tracking fix to the follow-up ticket is fine. Remaining: the PR body still claims "quoted `;` is always literal" (bug 3 description) and "Quoted `;` preserved" (verification section) without the agreed "embedded in a longer token" qualifier.

## Reasoning Chain

1. Verified both new tests at f9cc7d5 against the empirical probe cases from the original finding — all pinned.
2. Checked the PR body: the bug-3 description and verification bullets retain the unqualified claim this thread agreed to narrow.

## Solution

Narrow the PR-body wording as agreed ("a quoted `;` embedded in a longer token is literal; a value consisting solely of `;` is a known limitation"). Resolvable after that edit — no further code change needed.

test/unit/test_aerospike_shell.py
dwelch-spike: Resolved in https://github.com/aerospike/aerospike-admin/pull/516/commits/3f1dab1130c64de768aa8b5908f1c32bf3d12872

dwelch-spike (approved):
LGTM

PR #519, opened 2026-07-19
TOOLS-3662: ship astools.conf via copy-if-absent install script

a-spiker:
## What

Standalone tool packages ship no `astools.conf`, so a fresh install leaves the user without the config template. This adds it back without re-introducing conffile-ownership conflicts with the `aerospike-tools` bundle.

- New `pkg/astools.conf.sample`: byte-identical copy of the canonical `etc/astools.conf` template.
- New `pkg/postinst.sh` (deb `--after-install` + `--after-upgrade`, rpm `%post`): seeds `/etc/aerospike/astools.conf` from `/opt/aerospike/doc/asadm/astools.conf.sample` **only when the file is absent**.
- New `pkg/mac-scripts/postinstall`: same copy-if-absent logic for the macOS `.pkg`, target-volume aware.
- `pkg/Makefile`: stage the owned sample under `/opt/aerospike/doc/asadm/` (mac: `/usr/local/aerospike/doc/asadm/`); wire the scripts.
- Removed the stale, unreferenced `pkg/astools.conf`.

The live `/etc/aerospike/astools.conf` is deliberately **not owned** by the package. Result: standalone tools and the `aerospike-tools` bundle coexist with no dpkg/rpm conflict over the path, and user edits are never touched.

Wired as **both** `--after-install` and `--after-upgrade`: fpm's deb template routes an install-only hook into `after_upgrade()` that never runs on upgrade or reinstall, so both are required for existing installs to get the file.

## Tests

New `.github/bin/test/test_astools_conf.bats` runs in the linux and mac install jobs (kept out of the shared `test_execute.bats`, which is reused by the docker smoke job on the host):
- file created on install and matches the sample
- file recreated after removal + reinstall (guards the after-upgrade wiring)
- user edits survive reinstall

Docker smoke test asserts the image carries `/etc/aerospike/astools.conf`.

TOOLS-3662

### DIFF
--- a/.github/bin/test/test_astools_conf.bats
+++ b/.github/bin/test/test_astools_conf.bats
@@ -1,0 +1,32 @@
+#!/usr/bin/env bats
+#
+# Config-file lifecycle tests. Run ONLY from the linux-install and mac-install
+# jobs (never the docker smoke job, which shims tool binaries onto the host).
+# Requires two env vars exported by the install step:
+#   ASTOOLS_SAMPLE    - absolute path to the packaged astools.conf.sample
+#   ASTOOLS_REINSTALL - command that reinstalls the package from its local file
+
+setup() {
+  CONF=/etc/aerospike/astools.conf
+  SAMPLE="${ASTOOLS_SAMPLE}"
+  if [ "$(id -u)" -eq 0 ]; then SUDO=""; else SUDO="sudo"; fi
+}
+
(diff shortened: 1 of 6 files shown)


dwelch-spike (approved):
As long as we call out in the release notes and docs that the config file is left behind on uninstall this LGTM.

PR #524, opened 2026-07-28
ci(pkg): dynamic-dependency allowlist guard over the PyInstaller bundle

a-spiker:
## Problem

The asadm one-dir PyInstaller bundle deliberately loads glibc, libstdc++, libgcc_s, zlib, and libcrypt from the host (`excludes_binaries` in pyinstaller-build.spec) while the deb/rpm declare no dependencies at all. Nothing verifies the host-loaded set stays within that intent: a bundled library silently gaining a new host dependency (directly or transitively, e.g. via the bundled `less`) only surfaces at install-test time, and only if it breaks `--help`. Same hardening as aerospike/aql#84, aerospike/aerospike-tools-backup#164, aerospike/aerospike-benchmark#155, aerospike/asconfig#126.

## Changes

- **.github/bin/build_package.sh**: `assert_dynamic_deps` gate between build and packaging. Sweeps every ELF in `build/bin/asadm` (executables plus `_internal/`), subtracts libraries shipped inside the bundle, and fails on any remaining `DT_NEEDED` outside the allowlist (glibc family, libstdc++/libgcc_s, zlib, libcrypt, loaders). Covers transitive deps of bundled libraries too. Runs in every job of both build-and-release and the PR-time build-check.

No package metadata changes: the allowed host libs are present wherever apt/dnf exist (apt and dnf themselves require libstdc++ and zlib), matching the spec's exclusion intent.

## Verification

- Guard logic validated in a container against real ELFs: flags an unexpected host dep, passes once the lib is bundled, and then flags the bundled lib's own missing transitive dep.
- Dry-run Build and Release dispatch on this branch (release=false) exercises the sweep across the full 10-distro × 2-arch matrix; results posted below.

### DIFF
--- a/.github/bin/build_package.sh
+++ b/.github/bin/build_package.sh
@@ -2,6 +2,35 @@
 set -euo pipefail
 [[ -n "${DEBUG:-}" ]] && set -x
 
+function assert_dynamic_deps() {
+	local bundle="$GIT_DIR/build/bin/asadm"
+	local allowed="libc.so.6 libm.so.6 libpthread.so.0 libdl.so.2 librt.so.1 libutil.so.1
+		libgcc_s.so.1 libstdc++.so.6 libz.so.1 libcrypt.so.1
+		ld-linux-x86-64.so.2 ld-linux-aarch64.so.1"
+	local provided
+	provided=$(find "$bundle" -name 'lib*.so*' -printf '%f\n')
+
+	local elf lib needed fail=0
+	while IFS= read -r elf; do
+		needed=$(readelf -d "$elf" 2>/dev/null | awk '/\(NEEDED\)/ { gsub(/[][]/, "", $NF); print $NF }') || true
+		[ -z "$needed" ] && continue
+		for lib in $needed; do
+			if printf '%s\n' "$provided" | grep -qxF "$lib"; then
+				continue
+			fi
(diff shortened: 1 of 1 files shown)


a-spiker:
Verification: dry-run Build and Release dispatch on this branch ([run 30398376066](https://github.com/aerospike/aerospike-admin/actions/runs/30398376066), release=false) green across the full matrix; job logs show `assert_dynamic_deps: all host-loaded DT_NEEDED entries within allowlist` in every Linux job.

Downloaded all 20 Linux packages and re-ran the sweep offline over every ELF in each extracted bundle:

| Package family | Declared deps | Host-loaded set (bundle-provided libs subtracted) |
|---|---|---|
| deb debian11, rpm el8 (both arches) | none / /bin/sh | glibc family (incl. libutil), libgcc_s, libz |
| all other debs and rpms (both arches) | none / /bin/sh | glibc family, libgcc_s, libz |

Nothing outside the allowlist anywhere. Notably the bundled `less` does not leak host libtinfo (PyInstaller bundles it), and nothing in the bundle currently needs host libstdc++ or libcrypt; those stay in the allowlist as declared headroom matching the spec's `excludes_binaries` intent.

pvinh-spike (approved):
LGTM

PR #525, opened 2026-07-30
ci: update get-pip.py sha256 pin for pip 26.2

a-spiker:
## Problem

The build-and-release workflow fails on the ubuntu26.04 path:

```
sha256sum: WARNING: 1 computed checksum did NOT match
```

`install_deps.sh` downloads `get-pip.py` from bootstrap.pypa.io and verifies it against a pinned sha256 before bootstrapping pip. PyPA republished the file (it now bundles pip 26.2), so the old pin no longer matches and `set -e` aborts the job.

## Fix

Update the pinned hash to the current file (`25b5c39a...`). Verified the downloaded file is the genuine PyPA base85-encoded pip 26.2 bundle before pinning.

The `GET_PIP_SHA256` env var remains as an override for future rotations.


pvinh-spike (approved):
LGTM

PR #528, opened 2026-07-31
fix(pkg): embed distro in deb revision per Debian convention

a-spiker:
## What

Moves the distro into the deb **revision**, per Debian convention and matching server packages (`aerospike-server-community_8.1.3.0-66debian13_arm64.deb`). Follow-up to the earlier `-1` iteration change.

| | Filename |
|---|---|
| Before | `aerospike-asadm_5.0.3-1_debian12_x86_64.deb` |
| After | `aerospike-asadm_5.0.3-1debian12_x86_64.deb` |

Debian filenames carry exactly three underscore-separated fields (`name_version_arch`); vendors encode the distro inside the revision (`-0ubuntu0.24.04.1`, `.pgdg120+1`, `~bpo11+1`, server's `-66debian13`).

## How

- `pkg/Makefile`: `DEB_ITERATION = $(ITERATION)$(BUILD_DISTRO)`; deb fpm gets `--iteration 1<distro>`, so control `Version` is e.g. `5.0.3-1debian12` and per-distro builds carry distinct Versions.
- rpm unchanged: dist tag already sits in the Release position (`-1.el9`), which is the rpm equivalent.
- CI test-install deb glob and `docker/docker-build.sh` updated; the relaxed glob (`*<distro>_*`) matches both new and previous layouts.
- JFrog upload side verified: shared-workflows `get_codename_for_deb` substring-matches the distro anywhere in the filename, so pool/codename mapping is unaffected.

## Verification

`make -n` dry runs: `aerospike-asadm_9.9.9-1ubuntu24.04_x86_64.deb`; rc case `9.9.9-rc1-1debian12` parses as upstream `9.9.9-rc1` + revision `1debian12` under dpkg's last-hyphen rule.

## Coordination

Bundle PR citrusleaf/aerospike-tools#436 already fetches with distro-agnostic globs covering both layouts.

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -437,7 +437,7 @@ jobs:
           case "${{ matrix.distro }}" in
             ubuntu*|debian*)
               export DEBIAN_FRONTEND=noninteractive
-              pkg=$(find ./packages -name "*_${{ matrix.distro }}_*${arch}*.deb" | head -1)
+              pkg=$(find ./packages -name "*${{ matrix.distro }}_*${arch}*.deb" | head -1)
               [[ -n "${pkg}" ]] || { echo "ERROR: no .deb found for ${{ matrix.distro }}/${arch}" >&2; exit 1; }
               apt-get install -y "${pkg}"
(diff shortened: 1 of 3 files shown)


pvinh-spike (approved):
LGTM

PR #530, opened 2026-08-03
fix(pkg): fold rcN into the package iteration, keep it out of package names and docker tags

a-spiker:
## Problem

Release-candidate tags leak into every artifact name and docker tag. For 5.0.3-rc1:

```
aerospike-asadm_5.0.3-rc1-1debian12_x86_64.deb
aerospike-asadm-5.0.3_rc1-1.el9.x86_64.rpm
aerospike-asadm-5.0.3-rc1-macos-14-ARM64.pkg
aerospike/aerospike-asadm:5.0.3-rc1{,-amd64,-arm64}
```

## Change

`rcN` becomes the package iteration, which is exactly what the rpm `Release` and deb revision fields are for. Nothing carries `rcN` any more:

| version | deb | rpm | macOS .pkg | tar | docker tag |
|---|---|---|---|---|---|
| 5.0.3-rc1 | `aerospike-asadm_5.0.3-1debian12_x86_64.deb` | `aerospike-asadm-5.0.3-1.el9.x86_64.rpm` | `aerospike-asadm-5.0.3-1-macos-14-ARM64.pkg` | `aerospike-asadm-debian12-5.0.3-1-x86_64.tar` | `5.0.3-1` |
| 5.0.3-rc2 | `...5.0.3-2debian12...` | `...5.0.3-2.el9...` | `...5.0.3-2-macos-14...` | `...5.0.3-2...` | `5.0.3-2` |
| 5.0.3 | `...5.0.3-1debian12...` | `...5.0.3-1.el9...` | `...5.0.3-1-macos-14...` | `...5.0.3-1...` | `5.0.3-1` |

`VERSION` itself is untouched. It stays the git tag, the release bundle version, the JFrog build number, and what `asadm --version` prints, so the rc identity is not lost anywhere it matters.

New `.github/bin/pkg_release.sh` is the single source of the split, shared by `pkg/Makefile`, the workflow and `docker/docker-build.sh`, so the built name and the name the docker build looks for cannot drift apart.

## Edge cases

- **No pre-release tag** (`5.0.3`): iteration 1, identical to today's names.
- **Branch / feature builds** (`5.0.3-rc2-master.abc123456`): the branch qualifier stays in the version, iteration is still 2 -> `aerospike-asadm_5.0.3-master.abc123456-2debian12_x86_64.deb`.
- **`rc01`** normalizes to iteration 1, so it cannot produce a second file name that dpkg/rpm then compare as *equal* to rc1's.
- **Non-rc pre-release** (`5.0.3-beta1`): left in the version, so two different pre-releases can never collapse onto the same package name. Only `rcN` is folded.
- **Iteration >= 10** stays unambiguous: the deb revision is `<iteration><distro>` and the rpm Release is followed by `.<distro>`.
- **`5.0.3-rc1` and a bare `5.0.3` map to the same package name** (`5.0.3-1`). That is intended: rc1 *is* iteration 1 of 5.0.3, and the release flow promotes the rc that passed rather than re-cutting a bare version.
- `docker-build.sh` gained `-i/--iteration` to set the iteration directly (rebuilds of the same version).
- Untouched deliberately: the `find` globs in the install/smoke-test jobs and `build-check.yml` match on distro+arch, not on the version, so they keep working.

## Test

- `make -f pkg/Makefile -n deb rpm tar osx-pkg` for `5.0.3-rc2`, `5.0.3`, `5.0.3-rc2-master.abc123456`: names and `--version`/`--iteration` as tabled above; `pkgbuild --version` gets `5.0.3.2` / `5.0.3.1` / `5.0.3.master.abc123456.2`.
- `docker/docker-build.sh -N -t -v <version>` resolves the right .deb URL and image tag for the same set, and `-i 5` overrides the iteration.
- `pkg_release.sh` checked against all shapes above; `shellcheck -S warning` clean; workflow YAML parses.

Sibling PR in aerospike-tools applies the same scheme to the bundle and teaches its component download the new layout.


pvinh-spike (approved):
LGTM

PR #533, opened 2026-08-06
fix(ci): sign and notarize macOS pkgs; gate the release on it

pvinh-spike:
## Summary

`notarize-mac-artifacts` has been a **silent no-op**. Run [30840082628](https://github.com/aerospike/aerospike-admin/actions/runs/30840082628) (asadm 5.0.3-2) shipped four **unsigned, un-notarized** macOS pkgs while the job reported success — in 44s, far too fast for a notarization round-trip.

This is the same defect as citrusleaf/aerospike-tools#440, found by checking whether that bug reproduced here. It does, identically.

## Root cause

`artifact-glob` was `"**/*.pkg"`. The shared signer's `matches_glob()` ([`sign-mac-artifacts/entrypoint.sh:375`](https://github.com/aerospike/shared-workflows/blob/v3.6.0/.github/workflows/sign-mac-artifacts/entrypoint.sh#L375)) is **not a globstar matcher**. It short-circuits only on the exact literal `**/*`, and otherwise does a bash `case` match against the basename, then the relative path:

```bash
if [[ $pattern == "**/*" ]]; then return 0; fi     # literal string compare
filename=$(basename "$file")
case "$filename" in $pattern) return 0 ;; esac
case "$file"     in $pattern) return 0 ;; esac
```

In a `case` pattern, `**` is just `*` followed by `*` — so `**/*.pkg` requires a **literal `/`** in the subject. But `organize-built-artifacts` merges the matrix artifacts with `merge-multiple: true`, which **flattens** the tree, so the pkgs arrive as bare filenames and match neither candidate:

```
signed-mac-output/aerospike-asadm-5.0.3-2-macos-14-ARM64.pkg
signed-mac-output/aerospike-asadm-5.0.3-2-macos-15-ARM64.pkg
signed-mac-output/aerospike-asadm-5.0.3-2-macos-15-intel-X64.pkg
signed-mac-output/aerospike-asadm-5.0.3-2-macos-26-ARM64.pkg
...
==> Processing artifacts for signing

Signing complete
  Files signed:  0
  Files skipped: 24 (not matching glob or not a signable type)
```

The entrypoint then **exits 0**. The pkgs were re-uploaded untouched, and nothing downstream noticed — `sign-artifacts` is GPG-only, and the macOS test jobs ran no signature checks.

### Why this was never caught

The input's default is `artifact-glob: "**/*"`, which the matcher bypasses entirely via that literal string compare — so the default always works. No upstream macOS test overrides it, meaning the `**/`-prefixed path is **never exercised**. The only test anywhere passing an explicit glob is the *Windows* smoke test, using `"*.exe"` — the single-star form. The bug is only reachable by a caller passing an explicit `**/` glob, which is exactly what we did.

## The fix

`artifact-glob: "*.pkg"`. Because `matches_glob()` tries the **basename first**, this is layout-independent — correct for flat *and* nested trees, unlike the pattern it replaces. Verified by running the real `matches_glob()` over the exact 24 filenames from this run:

| glob | result |
|---|---|
| `**/*.pkg` | 0 signed, 24 skipped ❌ |
| `*.pkg` | **4 signed**, 20 skipped ✅ (rpms/debs correctly ignored) |

**Considered and rejected:** restructuring the artifact tree so `**/*.pkg` matches. It works, but makes signing depend on an accident of directory depth in an unrelated job — anyone later simplifying `organize-built-artifacts` silently re-breaks it. `*.pkg` is one line and layout-independent.

## Two new gates

Added to `test-install-mac-and-execute`, which already has the shipping artifact on a mac runner (all four macOS versions):

**1. `Verify .pkg is signed and notarized`** — `pkgutil --check-signature` asserting the *Aerospike Developer ID Installer* identity specifically (not merely that *some* signature exists), plus `xcrun stapler validate`. Confirmed on macOS that these exit non-zero on an unsigned pkg and on a missing ticket. This also generally covers the signer's zero-match no-op, since the entrypoint itself still exits 0 on `Files signed: 0`.

**2. `Verify installed files are root-owned`** — signing **activates a path that was dormant** while the glob was broken. When the signer codesigns an embedded binary it rebuilds the pkg Payload with an unprivileged `tar -xf | cpio -o --format odc` and re-flattens **without regenerating `Bom`** ([entrypoint.sh:252](https://github.com/aerospike/shared-workflows/blob/v3.6.0/.github/workflows/sign-mac-artifacts/entrypoint.sh#L252)), recording the runner's uid instead of root. Reproduced locally:

```
original pkg payload:   -rwxr-xr-x  1 0    0    ./usr/local/bin/tsttool
after cpio repack:      -rwxr-xr-x  1 502  20   ./usr/local/bin/tsttool
```

`pkgbuild` records `0:0`; the repack records the invoking uid. On a GitHub runner that's uid 501 — **the primary admin account on a customer Mac** — so root-installed binaries would land user-writable. The existing bats tests only exercise the CLI, so this would ship unnoticed.

The check asserts `/usr/local/aerospike/bin/asadm` exists *before* scanning, so a changed install location can't make it pass vacuously. It scans only `/usr/local/aerospike`; `pkg/mac-scripts/postinstall` writes solely to `/etc/aerospike` and runs as root, so it can't false-positive here.

## shared-workflows v3.6.0 → v3.7.0

Also bumps all four reusable-workflow refs to v3.7.0. Each job's `uses@tag` moves in lockstep with its `gh-workflows-ref` — that ref is what the reusable workflow checks out to find `entrypoint.sh`, so a mismatch would run v3.6.0 script logic under a v3.7.0 workflow.

**This bump does not fix the bug and is functionally inert.** For the four workflows this repo consumes, v3.7.0 changes only dependency pins:

- `harden-runner` v2.19.3 → v2.19.4 (all four)
- `setup-gpg` repinned v3.5.0 → v3.6.0 in `sign-artifacts` — and that action's logic is **byte-identical** between those two commits (README-only diff)

`sign-mac-artifacts/entrypoint.sh` is byte-identical, and `matches_glob()` is unchanged in v3.7.0, v4.0.0, **and on `main`** — no released version fixes this. The `artifact-glob` change above is what makes signing work.

Verified every input and secret this repo passes is still accepted by v3.7.0's `workflow_call` schemas (6+6, 3+3, 11, 8): no unknown keys, no missing required ones, and all four `uses@tag` / `gh-workflows-ref` pairs agree.
## Verification

- Ran the real `matches_glob()` (lifted verbatim) over the actual 24 filenames from the failing run — see table above.
- Confirmed `pkgutil --check-signature` exits 1 on an unsigned pkg and `xcrun stapler validate` exits 66 with no ticket.
- Reproduced the uid downgrade with `pkgbuild` + the entrypoint's exact `cpio` repack.
- Confirmed the install location (`pkg/Makefile:129`) and that `postinstall` cannot affect the ownership scan.
- YAML parses; every `run:` body passes `bash -n`.

**Not yet proven:** `actionlint` was unavailable locally, and the `pkgutil`/`stapler` assertions have only been exercised against locally-built pkgs. **The real confirmation is a release run where `notarize-mac-artifacts` reports `Files signed: 4` and both new steps pass** — please trigger one before merging.

## Follow-up (other repo)

- [ ] **`aerospike/shared-workflows`**: `entrypoint.sh` should exit non-zero when `notarize: true` and `signed_count == 0`; `matches_glob()` should either implement `**/` properly or reject it loudly; and the non-default glob path needs test coverage. That is the highest-value fix — it turns this class of failure from silent-green into a hard stop for every consumer, instead of each repo discovering it after shipping.

## Release impact

asadm 5.0.3-2's macOS pkgs are unsigned and un-notarized and should not ship — users would hit Gatekeeper rejection. A rebuild is needed after this merges.

## Review follow-up (applied)

A review ran against this PR; findings were addressed in a follow-up commit. All fixes are consumer-side — **shared-workflows is unchanged.**

| Finding | Status |
|---|---|
| Identity grep dropped the Team ID, was position-blind, and used a regex where a fixed string was meant | **Fixed** — full identity, leaf-pinned, `grep -qF` |
| No trust evaluation (`Status:` unread; `stapler` does not check the chain) | **Fixed** — added `spctl --assess --type install` as the gate |
| Ownership check missed the `/usr/local/bin` payload root; tested uid only; could not distinguish its own failure from success | **Fixed** — both roots scanned, mode bits tested, error swallowing removed, non-vacuity asserted |
| Verified pkg was not provably the installed pkg | **Fixed** — resolved once, exact-count asserted, exported as `VERIFIED_PKG` |
| Verification coverage coupled to the test matrix | **Open** — needs a dedicated `verify-mac-signing` job; see below |
| Embedded binaries signed with hardened runtime and no entitlements | **Open — blocking**; see below |
| Bundle payload assembled from unverified downloads (bundle repo only) | **Open** — pre-existing; needs its own PR |

### Blocking follow-up before a real release: entitlements

Fixing the glob activates the shared signer's embedded-binary codesigning, which runs `codesign --deep --force --options runtime --timestamp --sign` with **no `--entitlements`** (verified: zero occurrences in `entrypoint.sh@v3.7.0`). This repo's legacy path signed with `disable-library-validation`, and asadm's PyInstaller tree additionally with `allow-unsigned-executable-memory`. A binary that needs those dies at exec under library validation.

The fix is available consumer-side and is better than an upstream input, because the signer **skips** binaries already signed with the expected identity:

```bash
if [[ -n $current_authority && $current_authority == *"$SIGNING_IDENTITY"* ]]; then
    echo "    Already signed with correct identity, skipping: $rel"
```

So pre-signing embedded binaries in `build-mac-artifacts` with `--entitlements` both preserves the entitlements **and** leaves `signed_any=false`, which skips the `cpio` repack entirely — removing the ownership-downgrade defect at its root instead of detecting it after the fact.

Not implemented here: it requires the Apple Application cert and keychain setup wired into the build job across six repos, and this workflow cannot be executed from a PR (`workflow_dispatch`-only), so it would land untested in six release pipelines. **This PR instead makes that failure fail closed and legible** — every entrypoint must execute after install, with the error naming signing rather than surfacing as a bats failure.

### Also still open

- **Coverage coupling.** Verification runs per test-matrix shard, so adding a build target without a matching test target would still ship that pkg unverified. The fix is a dedicated `verify-mac-signing` job that loops every `*.pkg` and asserts an expected count, with `sign-artifacts` depending on it. Not landed untested.
- **Mutable tag pins.** The four highest-privilege callees are pinned to `@v3.7.0`, a mutable tag, while `actions/checkout` is SHA-pinned. Pre-existing and arguably deliberate policy; if tags are meant to float, the mitigation is tag protection on `aerospike/shared-workflows`.

### Evidence status — unchanged and still the main gap

None of these steps has executed. This workflow is `workflow_dispatch`-only, so every green check here is a linter or scanner. What closes it: one `release=false` dispatch with `artifact-glob` reverted to `**/*.pkg` showing the verify step **red**, and one with the fix showing all assertions **green** on all four mac runners. The ownership predicate now also tests group/world-writable bits, which needs that first run to confirm no legitimate installed file trips it.

What I did verify locally: YAML parses, every `run:` block passes `bash -n`, and the new ownership logic was exercised against a sandbox tree for pass, missing-entrypoint, vacuous-scan, world-writable, and non-executing-entrypoint cases — pass reaches all three assertions and exits 0; each failure case exits 1.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -655,12 +655,14 @@ jobs:
           echo "==> installed files are root-owned and not group/world-writable"
 
           # The shared signer applies hardened runtime (`codesign --options
-          # runtime`) with NO --entitlements, and exposes no input for them. This
-          # repo's legacy mac path signed with disable-library-validation (and
-          # allow-unsigned-executable-memory for asadm's PyInstaller tree). A
-          # binary that needs those dies at exec under library validation, so
-          # assert execution here, where the message names signing -- rather than
-          # later in the bats suite, where it reads as a test failure.
+          # runtime`) with NO --entitlements, and exposes no input for them.
+          # asadm's PyInstaller tree is the case that needs them: the
+          # aerospike-tools bundle path signs it with disable-library-validation
+          # and allow-unsigned-executable-memory (pkg/asadm-entitlements lives in
+          # that repo, not this one). A binary that needs those dies at exec
+          # under library validation, so assert execution here, where the message
+          # names signing -- rather than later in the bats suite, where it reads
+          # as a test failure.
           for rel in asadm/asadm asadm/asinfo; do
             b="/usr/local/aerospike/bin/${rel}"
             [[ -x "${b}" ]] || continue
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -353,7 +353,7 @@ jobs:
     needs:
       - get-version
       - organize-built-artifacts
-    uses: aerospike/shared-workflows/.github/workflows/reusable_sign-mac-artifacts.yaml@v3.6.0
+    uses: aerospike/shared-workflows/.github/workflows/reusable_sign-mac-artifacts.yaml@v3.7.0
     permissions:
       contents: read
       id-token: write
@@ -365,7 +365,7 @@ jobs:
       artifact-glob: "*.pkg"
       signing-identity: "Developer ID Application: Aerospike, Inc. (22224RFU67)"
       installer-identity: "Developer ID Installer: Aerospike, Inc. (22224RFU67)"
-      gh-workflows-ref: "v3.6.0"
+      gh-workflows-ref: "v3.7.0"
       gh-retention-days: 1
     secrets:
       apple-application-cert: ${{ secrets.APPLE_APPLICATION_CERT }}
@@ -379,11 +379,11 @@ jobs:
     needs:
       - get-version
       - notarize-mac-artifacts
-    uses: aerospike/shared-workflows/.github/workflows/reusable_sign-artifacts.yaml@v3.6.0
+    uses: aerospike/shared-workflows/.github/workflows/reusable_sign-artifacts.yaml@v3.7.0
     with:
       gh-artifact-name: signed-artifacts-${{ needs.get-version.outputs.VERSION }}
       gh-unsigned-artifacts: unsigned-artifacts-${{ needs.get-version.outputs.VERSION }}
-      gh-workflows-ref: "v3.6.0"
+      gh-workflows-ref: "v3.7.0"
     secrets:
       gpg-private-key: ${{ secrets.GPG_SECRET_KEY }}
       gpg-public-key: ${{ secrets.GPG_PUBLIC_KEY }}
@@ -611,7 +611,7 @@ jobs:
     needs:
       - get-version
       - organize-signed-artifacts
-    uses: aerospike/shared-workflows/.github/workflows/reusable_deploy-artifacts.yaml@v3.6.0
+    uses: aerospike/shared-workflows/.github/workflows/reusable_deploy-artifacts.yaml@v3.7.0
     with:
       jf-project: database
       jf-build-name: "aerospike-asadm"
@@ -623,7 +623,7 @@ jobs:
       dry-run: ${{ inputs.release != true }}
       jf-build-id: ${{ needs.get-version.outputs.VERSION }}
       jf-metadata-build-id: ${{ needs.get-version.outputs.VERSION }}-metadata
-      gh-workflows-ref: "v3.6.0"
+      gh-workflows-ref: "v3.7.0"
 
   docker-build-arch:
     name: Docker build, test & push (${{ matrix.arch }})
@@ -824,7 +824,7 @@ jobs:
       - get-version
       - upload-artifacts-to-jfrog
       - docker-manifest
-    uses: aerospike/shared-workflows/.github/workflows/reusable_create-release-bundle.yaml@v3.6.0
+    uses: aerospike/shared-workflows/.github/workflows/reusable_create-release-bundle.yaml@v3.7.0
     with:
       jf-project: "database"
       jf-build-names: "aerospike-asadm:${{ needs.get-version.outputs.VERSION }},aerospike-asadm-docker:${{ needs.get-version.outputs.VERSION }}"
@@ -832,7 +832,7 @@ jobs:
       oidc-provider-name: database-gh-aerospike
       oidc-audience: database-gh-aerospike
       version: ${{ needs.get-version.outputs.VERSION }}
-      gh-workflows-ref: v3.6.0
+      gh-workflows-ref: v3.7.0
       dry-run: false
 
   # Tag-last: created only after every publish step succeeds. A failure
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -532,12 +532,22 @@ jobs:
           # verified here, and an ambiguous match fails instead of silently
           # picking find's first result.
           pat="*-${{ matrix.os }}-${{ runner.arch }}.pkg"
-          n=$(find ./packages -type f -name "${pat}" | wc -l | tr -d ' ')
+          # One find, reused for the count, the error message and the result --
+          # three separate walks could disagree with each other. (mapfile would
+          # be the idiomatic reader, but the macOS runners ship bash 3.2, which
+          # has no mapfile builtin.) A find that fails outright fails this
+          # assignment under `set -e`.
+          matches=$(find ./packages -type f -name "${pat}")
+          if [[ -n "${matches}" ]]; then
+            n=$(printf '%s\n' "${matches}" | wc -l | tr -d ' ')
+          else
+            n=0
+          fi
           [[ "${n}" -eq 1 ]] || {
             echo "ERROR: expected exactly 1 .pkg matching ${pat}, found ${n}" >&2
-            find ./packages -type f -name "${pat}" >&2
+            printf '%s\n' "${matches}" >&2
             exit 1; }
-          pkg=$(find ./packages -type f -name "${pat}")
+          pkg="${matches}"
           echo "VERIFIED_PKG=${pkg}" >> "$GITHUB_ENV"
 
           # Capture once. An unsigned pkg exits non-zero here; a pkg signed by
@@ -667,10 +677,24 @@ jobs:
           # PyInstaller) so the shared signer will re-sign them -- if signing
           # ever silently skipped them again, an unsigned binary would ship and
           # be killed outright on arm64.
+          # `-perm +111` is any-exec-bit, and on macOS it is the ONLY spelling:
+          # /usr/bin/find here is BSD find, whose `-perm` grammar is
+          # `[-|+]mode`. GNU find's `/mode` is a hard error ("illegal mode
+          # string"), so do not "modernize" this -- this step only ever runs on
+          # macOS runners.
+          #
+          # Resolved into a variable rather than read from a process
+          # substitution: there, a find that errors out feeds an empty loop and
+          # "every embedded Mach-O is signed" prints for a scan that never ran.
+          # The emptiness check below makes that failure loud instead.
+          execs=$(find /usr/local/aerospike -type f -perm +111)
+          [[ -n "${execs}" ]] || {
+            echo "ERROR: found 0 executable files under /usr/local/aerospike -- the Mach-O signature scan would pass vacuously" >&2
+            exit 1; }
           unsigned=$(while IFS= read -r b; do
               file "${b}" | grep -q "Mach-O" || continue
               codesign -dv "${b}" >/dev/null 2>&1 || echo "${b}"
-            done < <(find /usr/local/aerospike -type f -perm +111))
+            done <<< "${execs}")
           if [[ -n "${unsigned}" ]]; then
             echo "ERROR: unsigned Mach-O binary(ies) in the installed tree -- check the notarize-mac-artifacts job" >&2
             printf '%s\n' "${unsigned}" | head -20 >&2
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -635,6 +635,23 @@ jobs:
           done
           echo "==> all entrypoints execute under hardened runtime"
 
+          # Every embedded Mach-O must carry a signature. The outer pkg being
+          # signed says nothing about its contents, and the mac staging step
+          # strips foreign signatures (Apple-signed binaries bundled by
+          # PyInstaller) so the shared signer will re-sign them -- if signing
+          # ever silently skipped them again, an unsigned binary would ship and
+          # be killed outright on arm64.
+          unsigned=$(while IFS= read -r b; do
+              file "${b}" | grep -q "Mach-O" || continue
+              codesign -dv "${b}" >/dev/null 2>&1 || echo "${b}"
+            done < <(find /usr/local/aerospike -type f -perm +111))
+          if [[ -n "${unsigned}" ]]; then
+            echo "ERROR: unsigned Mach-O binary(ies) in the installed tree -- check the notarize-mac-artifacts job" >&2
+            printf '%s\n' "${unsigned}" | head -20 >&2
+            exit 1
+          fi
+          echo "==> every embedded Mach-O is signed"
+
       - name: Install bats
         run: brew install bats-core
 
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -359,9 +359,11 @@ jobs:
       id-token: write
     with:
       gh-unsigned-artifacts: unsigned-artifacts-${{ needs.get-version.outputs.VERSION }}
-      # The signer's matches_glob() is a bash `case` match, not a real globstar
-      # matcher, so "**/*.pkg" requires a literal "/" and never matches our flat
-      # artifact tree -- every pkg gets skipped and the job still exits 0.
+      # As observed in shared-workflows v3.7.0: matches_glob() is a bash `case`
+      # match, not a real globstar matcher, so "**/*.pkg" requires a literal "/"
+      # and never matches our flat artifact tree -- every pkg gets skipped and
+      # the job still exits 0. Do not "fix" this back to "**/*.pkg" unless the
+      # pinned signer above implements real globstar matching.
       artifact-glob: "*.pkg"
       signing-identity: "Developer ID Application: Aerospike, Inc. (22224RFU67)"
       installer-identity: "Developer ID Installer: Aerospike, Inc. (22224RFU67)"
@@ -518,47 +520,120 @@ jobs:
           path: ./packages
 
       - name: Verify .pkg is signed and notarized
+        env:
+          # Must match `installer-identity` on the notarize-mac-artifacts job.
+          # A reusable-workflow `with:` cannot read `env`, so this is the one
+          # place the literal is repeated -- keep the two in sync.
+          INSTALLER_IDENTITY: "Developer ID Installer: Aerospike, Inc. (22224RFU67)"
         run: |
           set -euo pipefail
-          pkg=$(find ./packages -name "*-${{ matrix.os }}-${{ runner.arch }}.pkg" | head -1)
-          [[ -n "${pkg}" ]] || { echo "ERROR: no .pkg found for ${{ matrix.os }}-${{ runner.arch }}" >&2; exit 1; }
-          echo "==> pkgutil --check-signature ${pkg}"
-          pkgutil --check-signature "${pkg}" || {
-            echo "ERROR: ${pkg} is not signed -- check the notarize-mac-artifacts job" >&2; exit 1; }
-          pkgutil --check-signature "${pkg}" | grep -q "Developer ID Installer: Aerospike, Inc." || {
-            echo "ERROR: ${pkg} is not signed by the Aerospike Developer ID Installer identity" >&2; exit 1; }
-          echo "==> xcrun stapler validate ${pkg}"
+          # Resolve the pkg ONCE and export it. The install step below consumes
+          # ${VERIFIED_PKG}, so it cannot install a different file than the one
+          # verified here, and an ambiguous match fails instead of silently
+          # picking find's first result.
+          pat="*-${{ matrix.os }}-${{ runner.arch }}.pkg"
+          n=$(find ./packages -type f -name "${pat}" | wc -l | tr -d ' ')
+          [[ "${n}" -eq 1 ]] || {
+            echo "ERROR: expected exactly 1 .pkg matching ${pat}, found ${n}" >&2
+            find ./packages -type f -name "${pat}" >&2
+            exit 1; }
+          pkg=$(find ./packages -type f -name "${pat}")
+          echo "VERIFIED_PKG=${pkg}" >> "$GITHUB_ENV"
+
+          # Capture once. An unsigned pkg exits non-zero here; a pkg signed by
+          # the WRONG identity exits 0 and is caught by the assertion below.
+          sig=$(pkgutil --check-signature "${pkg}") || {
+            echo "ERROR: ${pkg} is not signed -- check the notarize-mac-artifacts job" >&2
+            printf '%s\n' "${sig}" >&2
+            exit 1; }
+          printf '%s\n' "${sig}"
+
+          # Assert the FULL configured identity, Team ID included, at the leaf
+          # ("1. ") position. Org names are not unique across Apple teams; the
+          # Team ID is. -F so "(", ")" and "." are literal, not regex.
+          printf '%s\n' "${sig}" | grep -qF "1. ${INSTALLER_IDENTITY}" || {
+            echo "ERROR: ${pkg} leaf signer is not ${INSTALLER_IDENTITY}" >&2
+            printf '%s\n' "${sig}" >&2
+            exit 1; }
+
+          # Neither pkgutil's exit code nor stapler evaluates trust: pkgutil
+          # reports an untrusted or expired chain on its Status: line, and
+          # stapler only checks that a ticket is present and matches. spctl runs
+          # the assessment a customer's Mac actually applies.
+          spctl --assess --type install -vv "${pkg}" || {
+            echo "ERROR: ${pkg} is rejected by Gatekeeper -- check the notarize-mac-artifacts job" >&2
+            exit 1; }
           xcrun stapler validate "${pkg}" || {
-            echo "ERROR: ${pkg} has no stapled notarization ticket -- check the notarize-mac-artifacts job" >&2; exit 1; }
+            echo "ERROR: ${pkg} has no stapled notarization ticket -- check the notarize-mac-artifacts job" >&2
+            exit 1; }
 
       - name: Install .pkg locally
         run: |
           set -euo pipefail
-          pkg=$(find ./packages -name "*-${{ matrix.os }}-${{ runner.arch }}.pkg" | head -1)
-          [[ -n "${pkg}" ]] || { echo "ERROR: no .pkg found for ${{ matrix.os }}-${{ runner.arch }}" >&2; exit 1; }
-          sudo installer -pkg "${pkg}" -target /
-          echo "ASTOOLS_REINSTALL=sudo installer -pkg ${pkg} -target /" >> "$GITHUB_ENV"
+          # ${VERIFIED_PKG} is exported by the verify step above, so we can
+          # only install the exact file that was verified.
+          sudo installer -pkg "${VERIFIED_PKG}" -target /
+          echo "ASTOOLS_REINSTALL=sudo installer -pkg ${VERIFIED_PKG} -target /" >> "$GITHUB_ENV"
           echo "ASTOOLS_SAMPLE=/usr/local/aerospike/doc/asadm/astools.conf.sample" >> "$GITHUB_ENV"
 
-      - name: Verify installed files are root-owned
+      - name: Verify installed files are root-owned and entrypoints execute
         run: |
           set -euo pipefail
-          # The shared signer rebuilds the .pkg Payload with an unprivileged
-          # `tar -xf | cpio -o` whenever it codesigns an embedded binary, which
-          # records the runner's uid instead of root and leaves the Bom stale.
-          # On a customer Mac that uid is the primary admin account, so
-          # root-installed binaries would land user-writable.
-          # Assert the tree exists first, so a changed install location can't
-          # make this check pass vacuously.
-          [[ -d /usr/local/aerospike/bin/asadm ]] || {
-            echo "ERROR: /usr/local/aerospike/bin/asadm missing after install -- did the pkg install location change?" >&2; exit 1; }
-          bad=$(find /usr/local/aerospike ! -user root -exec ls -ldn {} + 2>/dev/null | head -20 || true)
+          # Assert the payload actually landed. A bare directory test would pass
+          # on an empty tree, which is the vacuous pass this guard prevents.
+          for rel in asadm/asadm asadm/asinfo; do
+            [[ -e "/usr/local/aerospike/bin/${rel}" ]] || {
+              echo "ERROR: /usr/local/aerospike/bin/${rel} missing after install -- did the pkg install location change?" >&2
+              exit 1; }
+          done
+
+          # Prove the scan is non-vacuous before trusting an empty result.
+          scanned=$(find /usr/local/aerospike -type f | wc -l | tr -d ' ')
+          [[ "${scanned}" -ge 1 ]] || {
+            echo "ERROR: ownership scan examined 0 files under /usr/local/aerospike" >&2
+            exit 1; }
+          echo "==> scanning ${scanned} installed file(s)"
+
+          # As observed in shared-workflows v3.7.0: when the signer codesigns an
+          # embedded binary it rebuilds the pkg Payload with an unprivileged
+          # `tar -xf | cpio -o` and re-flattens without regenerating Bom, so the
+          # payload records the build runner's uid rather than root. Mode bits are
+          # tested too -- the hazard is a writable file, not merely a non-root one.
+          # The pkg installs TWO roots (/usr/local/aerospike and the
+          # /usr/local/bin symlinks), so both are scanned; /usr/local/bin is
+          # filtered by -lname because Homebrew owns unrelated entries there.
+          # No 2>/dev/null and no `|| true`: a find that cannot walk the tree must
+          # fail the step, not read as success.
+          bad=$(
+            { find /usr/local/aerospike \( ! -user root -o -perm -0002 -o -perm -0020 \) -print
+              find /usr/local/bin -maxdepth 1 -lname '*aerospike/bin/*' ! -user root -print
+            }
+          ) || {
+            echo "ERROR: ownership scan failed -- see stderr above" >&2
+            exit 1; }
           if [[ -n "${bad}" ]]; then
-            echo "ERROR: installed files are not root-owned (see notarize-mac-artifacts payload rebuild)" >&2
-            echo "${bad}" >&2
+            echo "ERROR: $(printf '%s\n' "${bad}" | wc -l | tr -d ' ') installed path(s) not root-owned or group/world-writable" >&2
+            printf '%s\n' "${bad}" | head -20 | while IFS= read -r p; do ls -ldn "${p}" >&2; done
             exit 1
           fi
-          echo "==> all files under /usr/local/aerospike are root-owned"
+          echo "==> installed files are root-owned and not group/world-writable"
+
+          # The shared signer applies hardened runtime (`codesign --options
+          # runtime`) with NO --entitlements, and exposes no input for them. This
+          # repo's legacy mac path signed with disable-library-validation (and
+          # allow-unsigned-executable-memory for asadm's PyInstaller tree). A
+          # binary that needs those dies at exec under library validation, so
+          # assert execution here, where the message names signing -- rather than
+          # later in the bats suite, where it reads as a test failure.
+          for rel in asadm/asadm asadm/asinfo; do
+            b="/usr/local/aerospike/bin/${rel}"
+            [[ -x "${b}" ]] || continue
+            "${b}" --version >/dev/null 2>&1 || "${b}" --help >/dev/null 2>&1 || {
+              echo "ERROR: ${b} installed but will not execute -- likely hardened runtime without entitlements (see notarize-mac-artifacts)" >&2
+              codesign -d --entitlements - "${b}" >&2 || true
+              exit 1; }
+          done
+          echo "==> all entrypoints execute under hardened runtime"
 
       - name: Install bats
         run: brew install bats-core
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -359,7 +359,10 @@ jobs:
       id-token: write
     with:
       gh-unsigned-artifacts: unsigned-artifacts-${{ needs.get-version.outputs.VERSION }}
-      artifact-glob: "**/*.pkg"
+      # The signer's matches_glob() is a bash `case` match, not a real globstar
+      # matcher, so "**/*.pkg" requires a literal "/" and never matches our flat
+      # artifact tree -- every pkg gets skipped and the job still exits 0.
+      artifact-glob: "*.pkg"
       signing-identity: "Developer ID Application: Aerospike, Inc. (22224RFU67)"
       installer-identity: "Developer ID Installer: Aerospike, Inc. (22224RFU67)"
       gh-workflows-ref: "v3.6.0"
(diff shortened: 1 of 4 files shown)


pvinh-spike:

**Promises:**
1. macOS .pkg artifacts are Apple-signed (Developer ID Installer: Aerospike, Inc.) and notarized+stapled before release.
2. The signing step can no longer silently no-op: an unsigned/unstapled pkg fails the pipeline.
3. Installed files remain root-owned after `installer` (guarding the shared signer's unprivileged payload repack).
4. The new checks cannot pass vacuously.
5. The shared-workflows v3.6.0 -> v3.7.0 bump is behavior-neutral.
6. No change to Linux (.rpm/.deb) packaging or any non-macOS path.

| Gate | Status |
|------|--------|
| Contract | **fail** — promise 1 not kept (no trust evaluation); promise 3 enforced over part of the payload; promise 5 unverified from this PR |
| Impacted surface | **fail** — the verified pkg is not provably the shipped pkg; coverage is coupled to the test matrix |
| Failure paths | **fail** — ownership check cannot distinguish its own failure from success |
| Evidence | **blind spot** — `workflow_dispatch`-only, so every green check here is a linter/scanner. **None of the new steps has ever executed.** |
| Quality | **fail** — six-way copy-paste of a release gate, with drift already present in the introducing PR |
| Backward Compat | **fail** — signing now activates embedded-binary codesigning with hardened runtime and no entitlements |

**5 findings** posted inline.

### The one that matters most

Fixing the glob *activates* `codesign_pkg_contents()` in the shared signer, which signs embedded Mach-Os with `--options runtime` and **no `--entitlements`**. This repo's own legacy path signs with `disable-library-validation` (and `allow-unsigned-executable-memory` for asadm's PyInstaller tree). See the inline comment on the `artifact-glob` line.

### Verified against upstream (two candidate findings dropped)

I read `sign-mac-artifacts/entrypoint.sh@v3.7.0` directly rather than speculate:

- **The signer does staple** — `xcrun stapler staple` with 6 retries, and it runs `stapler validate` itself. So `stapler validate` here will not hard-fail every build. (It does make this check partly redundant, but re-checking the shipped artifact across a job boundary is defensible.)
- **Pre-fix component pins are not a notary blocker** — `codesign_pkg_contents()` signs unsigned embedded Mach-Os before submission.

Also dropped after adversarial verification: a claimed `pipefail`/SIGPIPE false-failure on `pkgutil | grep -q`. Measured: `pkgutil` takes the stdio default, a Darwin pipe's `st_blksize` is 16384, and `--check-signature` emits ~900 B (measured on four real Apple-signed pkgs) — one `write(2)`, ~13x headroom, 0/200 trials failed. The buffering is what makes this gate deterministic. The doubled `pkgutil` call was also cleared: unsigned exits 1, wrong-identity exits 0 and only the grep catches it, so the two calls catch two distinct failure modes.

### Also noted (not posted inline)

- **Six-way duplication.** ~35 identical lines across six repos. The original bug was a *silent skip*, so the assertion arguably belongs in `reusable_sign-mac-artifacts.yaml` where it would have caught it at the source in all six repos at once. Drift has already begun: this PR's guards diverge (`-d .../bin` in the bundle vs `-f .../bin/<tool>` in four tools).
- **Mutable tag pins.** The four highest-privilege callees (Apple certs, GPG key, JFrog OIDC) are pinned to `@v3.7.0`, a mutable tag, while `actions/checkout` three lines away is SHA-pinned. Pre-existing, and possibly deliberate policy — but this PR re-affirms those lines, and it inverts the usual risk ordering.
- **Comment staleness.** The "uid 501 is the primary admin account" sentence stakes a runner-image detail the code does not depend on (`! -user root` is correct at any uid). The `matches_glob()` claims should be anchored to "as observed in v3.7.0".

### What would close the Evidence gate

Two `release=false` dispatches, linked here: one with `artifact-glob` reverted to `**/*.pkg` showing the verify step **red** (negative control), one with the fix showing both new steps **green** on all four mac runners. Promise 3 additionally needs a manual repack experiment, since `pkgbuild --ownership recommended` normalizes the payload to root:wheel — so a green ownership check may be indistinguishable from a vacuous one.

.github/workflows/build-and-release.yml
```diff
@@ -353,16 +353,19 @@ jobs:
...
     permissions:
       contents: read
       id-token: write
     with:
       gh-unsigned-artifacts: unsigned-artifacts-${{ needs.get-version.outputs.VERSION }}
-      artifact-glob: "**/*.pkg"
+      # The signer's matches_glob() is a bash `case` match, not a real globstar
+      # matcher, so "**/*.pkg" requires a literal "/" and never matches our flat
+      # artifact tree -- every pkg gets skipped and the job still exits 0.
+      artifact-glob: "*.pkg"
```
pvinh-spike, line 367: From Architecture agent:

Severity: critical
Problem Confidence: high
Solution Confidence: medium

## Problem

This one line is load-bearing in a way the PR does not account for. Fixing the glob does not only make the *outer* pkg get signed — it activates `codesign_pkg_contents()` in the shared signer, which walks every embedded Mach-O and signs the unsigned ones. That path has never run before, because the broken glob skipped every pkg.

The signer signs them like this (`sign-mac-artifacts/entrypoint.sh@v3.7.0:235-236`):

```bash
codesign --deep --force --options runtime --timestamp \
    --sign "$SIGNING_IDENTITY" "$binary"
```

`--options runtime` enables the hardened runtime. There is **no `--entitlements`** — the shared workflow exposes no input for it, and none of the six call sites passes one.

This repo already states that is not sufficient. `pkg/Makefile.mac:35-43` signs the same binaries with entitlements:

- `pkg/entitlements` → `com.apple.security.cs.disable-library-validation`, applied to `asbackup`, `asrestore`, `aql`, `asbench`, `asconfig`
- `pkg/asadm-entitlements` → `disable-library-validation` **and** `com.apple.security.cs.allow-unsigned-executable-memory`, applied to every executable in asadm's PyInstaller `--onedir` tree, with a comment explaining why that tree needs per-binary treatment

So the effect of this PR is to sign those binaries under hardened runtime *minus* the two entitlements the repo documents as required. asadm is the most exposed: a PyInstaller onedir bundle with library validation **enabled** cannot load its own `_internal` dylibs unless every one carries a matching Team ID signature, and `allow-unsigned-executable-memory` exists precisely because CPython needs it.

Failure sequence:
1. Merge; dispatch a release.
2. `notarize-mac-artifacts` signs each embedded Mach-O with `--options runtime`, no entitlements.
3. Notarization succeeds (the binaries *are* signed), staple succeeds, and every check added by this PR passes.
4. `bats .github/bin/test/test_execute.bats` runs `<tool> --help` on the installed pkg and the binary fails to launch — reported as an *install test* failure, not a signing failure.
5. Or worse, if the tool under test happens not to exercise the affected load path, it ships and fails on a customer Mac.

## Reasoning Chain

1. The PR's own comment on the ownership step says the signer "codesigns an embedded binary" and repacks — so the author knows the embedded path activates. The entitlements consequence is not addressed anywhere.
2. Read the upstream signer at the pinned tag: the codesign invocation has `--options runtime` and no `--entitlements`, and `grep -c -- --entitlements` over that file is 0.
3. Read this repo's legacy signing path: it passes `--entitlements` on every binary, with a distinct, stricter plist for asadm and a comment justifying it. That is direct in-repo evidence that hardened runtime without those entitlements was tried and found insufficient.
4. Hardened runtime with library validation enabled is strictly more restrictive than with `disable-library-validation`. Signing that was previously skipped is now applied. Therefore the binaries move from "unsigned, unrestricted" to "signed, maximally restricted" in one step.
5. Nothing in the new gates detects this: signature present, ticket stapled, identity matches. The check surface and the failure surface do not intersect.

## Solution

Settle it before merge by reading whether `reusable_sign-mac-artifacts.yaml@v3.7.0` accepts an entitlements input.

- **If it does:** pass the equivalent of `pkg/entitlements` (and `pkg/asadm-entitlements` for asadm) from each call site, and move the plist into the repo that owns the binary — asadm needs one and `aerospike-admin/pkg/` has none today.
- **If it does not:** that is a blocking gap upstream, not something to paper over here. Either add the input, or keep inner-binary signing in-repo (as `pkg/Makefile.mac` does today) and let the shared signer only `productsign` + notarize the outer pkg. Note the signer hard-errors on a binary already signed with a different identity, so pre-signing in-repo with the *same* Developer ID is compatible — it takes the "already signed with correct identity, skipping" branch.

Either way, the smoke test is `<tool> --help` on an installed pkg from a real dispatch — which the existing bats suite already does, so the negative signal exists as soon as a run happens. Tradeoff: if the entitlements turn out to be unnecessary for a given tool, passing them is still harmless (they only relax).

.github/workflows/build-and-release.yml
```diff
@@ -514,6 +517,20 @@ jobs:
...
 
+      - name: Verify .pkg is signed and notarized
+        run: |
+          set -euo pipefail
+          pkg=$(find ./packages -name "*-${{ matrix.os }}-${{ runner.arch }}.pkg" | head -1)
+          [[ -n "${pkg}" ]] || { echo "ERROR: no .pkg found for ${{ matrix.os }}-${{ runner.arch }}" >&2; exit 1; }
+          echo "==> pkgutil --check-signature ${pkg}"
+          pkgutil --check-signature "${pkg}" || {
+            echo "ERROR: ${pkg} is not signed -- check the notarize-mac-artifacts job" >&2; exit 1; }
+          pkgutil --check-signature "${pkg}" | grep -q "Developer ID Installer: Aerospike, Inc." || {
```
pvinh-spike: From Security agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

The identity assertion drops the Team ID. Ten lines above, this workflow declares exactly what it expects:

```yaml
installer-identity: "Developer ID Installer: Aerospike, Inc. (22224RFU67)"
```

but the gate greps only for `Developer ID Installer: Aerospike, Inc.` — the discriminating half, `(22224RFU67)`, is absent.

Apple's Developer ID leaf CN is `Developer ID Installer: <Org Name> (<TeamID>)`. **Org names are not unique across Apple Developer teams; the Team ID is.** Any enrolled team whose org name is "Aerospike, Inc." — a duplicate/re-enrollment after a cert compromise, an acquired entity, a second team created for another purpose — produces a CN that satisfies this grep. The gate would green-light a pkg signed with a Developer ID that is not the one this workflow configured.

Two further weaknesses in the same expression:

1. **Position-blind.** `pkgutil --check-signature` prints the whole `Certificate Chain:`. `grep -q` matches anywhere in it, so it does not assert the string is the *leaf* (position-1) signer.
2. **Unescaped regex.** `Inc.` is a BRE — `.` is any-char. Harmless today (it matches the literal `.`), but it is a regex where a fixed string is meant.

Trigger sequence, no attacker CI access needed: a cert rotation produces a second Aerospike-named Developer ID under a new Team ID → someone edits `installer-identity` above → the signer uses it → this grep still matches the org-name prefix → `stapler validate` passes (it was genuinely notarized) → green on a pkg whose identity no longer matches what the org intends to ship, with no signal in CI.

## Reasoning Chain

1. The `with:` block names the full identity including the Team ID; the verifier checks a strict prefix of that same string, in the same file, ~150 lines apart.
2. Apple's uniqueness guarantee for these certs is the Team ID, not the org name — so a match that excludes it is strictly weaker than the assertion the workflow already declares.
3. `grep -q` over multi-line chain output cannot distinguish leaf from intermediate, so "identity appears" != "identity signed it".
4. Therefore the check proves "some cert in the chain has an Aerospike-looking org name" — not promise 1.
5. The two strings being duplicated literals in one file is itself the drift hazard: nothing forces them to agree.

## Solution

Assert the full identity as a fixed string, pinned to the leaf position, from one source of truth:

```bash
IDENTITY='Developer ID Installer: Aerospike, Inc. (22224RFU67)'
sig=$(pkgutil --check-signature "${pkg}") || {
  echo "ERROR: ${pkg} is not signed -- check the notarize-mac-artifacts job" >&2
  printf '%s\n' "${sig}" >&2; exit 1; }
printf '%s\n' "${sig}" | grep -qF "1. ${IDENTITY}" || {
  echo "ERROR: ${pkg} is not signed by ${IDENTITY}" >&2
  printf '%s\n' "${sig}" >&2; exit 1; }
```

`grep -qF` (fixed string, so `(`/`)`/`.` are literal) and the `1. ` prefix pins the leaf. Better still: hoist the identity into a job-level `env:` consumed both by `installer-identity:` and by this step, so they cannot drift.

Tradeoff: the `1. ` prefix depends on `pkgutil`'s output layout, which Apple can change across the macos-14…macos-26 span in this matrix. Pair it with the `spctl` assessment from the companion comment so a layout change fails closed rather than silently passing. Also note the failure path should print the captured output — right now a signing regression gives you a one-line error and no certificate chain to diagnose it with.

.github/workflows/build-and-release.yml
```diff
@@ -514,6 +517,20 @@ jobs:
           name: signed-artifacts-${{ needs.get-version.outputs.VERSION }}
           path: ./packages
 
+      - name: Verify .pkg is signed and notarized
+        run: |
+          set -euo pipefail
+          pkg=$(find ./packages -name "*-${{ matrix.os }}-${{ runner.arch }}.pkg" | head -1)
+          [[ -n "${pkg}" ]] || { echo "ERROR: no .pkg found for ${{ matrix.os }}-${{ runner.arch }}" >&2; exit 1; }
+          echo "==> pkgutil --check-signature ${pkg}"
```
pvinh-spike: From Security agent:

Severity: important
Problem Confidence: medium
Solution Confidence: high

## Problem

None of the three new checks evaluates **trust**. Together they establish only:

- `pkgutil --check-signature "$pkg"` (exit code) — a signature parses. `man pkgutil` documents no exit codes at all, so this leans on undocumented behavior. Measured: unsigned → exit 1 (good). **Not** established: signed-by-untrusted-cert and signed-by-expired-cert, both of which `pkgutil` reports on the `Status:` line — which this script never reads.
- the `grep` — a string appears in the output.
- `xcrun stapler validate` — a stapled ticket exists and matches this pkg. It does **not** evaluate the signing chain, its validity period, or revocation.

So there is a false-accept corridor: a pkg whose installer cert has expired, or whose chain no longer builds to Apple's Developer ID root (intermediate rotated out, cert revoked after issuance, wrong keychain contents in the signer), can pass all three checks and then be rejected by Gatekeeper on a customer Mac. A cert that was valid at signing+notarization time and expired afterward satisfies both `stapler validate` and the grep.

That leaves promise 2 half-kept: the *no-op* case is closed, the *bad-signature* case is not. And promise 1 — "signed and notarized" as a customer experiences it — is not what is being measured. The property a customer's Mac applies is a Gatekeeper assessment.

Settling check, one command with a signed pkg on a runner: sign with an untrusted/expired installer identity and run `pkgutil --check-signature that.pkg; echo $?`. If it exits 0 with `Status: signed by untrusted certificate`, the exit-code gate is vacuous for every untrusted signature and this becomes high confidence. (Could not complete locally: `productsign` refuses to sign with an untrusted identity — `CSSMERR_TP_NOT_TRUSTED` — so it needs a real untrusted-but-signed pkg.)

## Reasoning Chain

1. Enumerated what each of the three commands actually guarantees. The only trust-bearing input is `pkgutil`'s exit status, which is undocumented — an undocumented status is not a security contract.
2. `pkgutil` *does* distinguish trusted / untrusted / expired, via the printed `Status:` line. The script pipes that output to `grep` for the org name and discards the rest, so the verdict field is read by nothing.
3. `stapler validate` is a ticket presence + hash check. Verified: on an unsigned pkg it exits 66, so it fails closed — it adds "stapled", not "trusted".
4. Union of the three = {parses, org-name string present, ticket stapled} — strictly weaker than the policy the customer's machine applies. Hence CI-green / customer-reject is reachable.
5. Two independent reviewers reached this conclusion from different starting points (trust semantics vs. customer-observable behavior), which is why it is reported despite neither being able to measure the untrusted case.

## Solution

Add the assessment macOS actually performs, and make it the gate:

```bash
spctl --assess --type install -vv "${pkg}" || {
  echo "ERROR: ${pkg} is not accepted by Gatekeeper -- check the notarize-mac-artifacts job" >&2; exit 1; }
```

Verified locally on macOS 26.5.1: on an unsigned pkg this exits 3 with `source=no usable signature`, so it fails closed. On a good pkg it prints `source=Notarized Developer ID` and `origin=Developer ID Installer: Aerospike, Inc. (22224RFU67)` — covering signature, trust, notarization, and identity in one exit code. Keep `pkgutil` and `stapler` for their better diagnostics; let `spctl` be the gate, and grep its `origin=` line for the identity assertion in the companion comment.

Tradeoffs: (a) `spctl` may consult Apple for revocation/ticket lookup, so it is a network dependency — acceptable given `stapler validate` already proved the ticket is stapled, but treat a non-zero exit as a hard failure, never `|| true` it. (b) Do **not** fix this by grepping `Status: signed by a certificate trusted by macOS` — that wording has changed across releases ("Mac OS X" on older systems) and this matrix spans macos-14 through macos-26, so a literal `Status:` grep would false-fail on some runners. (c) `spctl` may surface pre-existing failures on older images; that is the gate working. Verify the exact strings on all four runner images in the same dispatch used for the positive control.

.github/workflows/build-and-release.yml
```diff
@@ -523,6 +540,26 @@ jobs:
...
+          # The shared signer rebuilds the .pkg Payload with an unprivileged
+          # `tar -xf | cpio -o` whenever it codesigns an embedded binary, which
+          # records the runner's uid instead of root and leaves the Bom stale.
+          # On a customer Mac that uid is the primary admin account, so
+          # root-installed binaries would land user-writable.
+          # Assert the tree exists first, so a changed install location can't
+          # make this check pass vacuously.
+          [[ -d /usr/local/aerospike/bin/asadm ]] || {
+            echo "ERROR: /usr/local/aerospike/bin/asadm missing after install -- did the pkg install location change?" >&2; exit 1; }
+          bad=$(find /usr/local/aerospike ! -user root -exec ls -ldn {} + 2>/dev/null | head -20 || true)
```
pvinh-spike: From Data Flow agent:

Severity: important
Problem Confidence: high
Solution Confidence: medium

## Problem

Three distinct defects in this one step. All were flagged independently by more than one reviewer.

**1. The scan misses half the payload.** Every one of these pkgs installs to two roots, not one. From `pkg/Makefile` `prep-mac`:

```make
install -d $(MAC_ROOT)/usr/local/bin
ln -sf ../aerospike/bin/<tool> $(MAC_ROOT)/usr/local/bin/<tool>
```

(aql:110-112, asbench:100-102, asconfig:101-103, asbackup+asrestore:118-122, asadm+asinfo:130-133; and the bundle inherits all of them — `build_bundle_mac.sh:57` `bsdtar -xpf "$payload" -C "$MAC_ROOT"`, plus `pkg/Makefile.mac:91-97` creating seven of them.)

Those symlinks are Payload entries, so the very repack this step guards against stamps the runner's uid onto them too — and `find /usr/local/aerospike` never looks there. `/usr/local/bin/<tool>` is the PATH-visible entry point: a non-root symlink there can be retargeted by that unprivileged account, so anything invoking the tool from PATH (including under `sudo`) runs attacker-chosen code. That is a worse outcome than a writable file under `/usr/local/aerospike`, and it is the uncovered one.

**2. uid-only, but the stated harm is writability.** The comment says files "would land user-writable". A file is user-writable via *any* of non-root owner, a writable group the user is in, or `o+w`. A cpio/tar repack carries mode and gid too, so `root:staff 0775` is a reachable outcome of the same mechanism — and every local macOS account is in `staff` (gid 20). `! -user root` reports that clean. The invariant tested is not the invariant claimed.

**3. It cannot distinguish its own failure from success.** `2>/dev/null` discards find's diagnostics; `|| true` maps every non-zero status to success; and the verdict is `[[ -n "${bad}" ]]`, so "nothing bad" and "nothing examined" are the same observable. A root-owned mode-0700 subdirectory makes `find` (running as `runner`) emit `Permission denied` and skip that subtree — stderr gone, status swallowed, `bad` empty, and the step prints the all-clear having inspected nothing inside it. `find` does not return non-zero for zero matches, so `|| true` buys nothing legitimate; it exists to absorb SIGPIPE from `head -20`, which is only needed because detection and display are fused into one pipeline.

Related: the bundle's guard is `[[ -d /usr/local/aerospike/bin ]]` and admin's is `[[ -d .../bin/asadm ]]` — both satisfied by an **empty** directory, so the anti-vacuity guard the comment promises is weakest in the two repos with the most complex payloads. The four other repos correctly assert a concrete file.

## Reasoning Chain

1. Read all five `pkg/Makefile` `prep-mac` targets and the bundle's `Makefile.mac` + `build_bundle_mac.sh` to establish the real payload path set, rather than assuming the scan root was the payload root. Two roots per pkg; the check scans one.
2. The motivating defect (whole-payload repack under the wrong uid) taints both roots uniformly, so the scan covers a strict subset of the affected files — partial detection, silently so.
3. BSD `find` defaults to `-P`, so `! -user root` on a symlink lstat()s it: symlink ownership *is* testable, it is simply untested.
4. Mapped the predicate against the comment's own threat statement: `! -user root` is a uid test, "user-writable" is a (uid, gid, mode) property.
5. For defect 3, traced each of the three constructs to what it suppresses, and confirmed the success/failure observables coincide.

## Solution

Split detection from display, widen both tree and predicate, and let `find` fail the step:

```bash
# Guard: assert the payload actually landed (per-repo entrypoint list).
for b in asadm/asadm asadm/asinfo; do
  [[ -e "/usr/local/aerospike/bin/${b}" ]] || {
    echo "ERROR: /usr/local/aerospike/bin/${b} missing after install" >&2; exit 1; }
done

bad=$(
  { find /usr/local/aerospike \( ! -user root -o -perm -0002 -o -perm -0020 \) -print
    find /usr/local/bin -maxdepth 1 -lname '*aerospike/bin/*' ! -user root -print
  }
) || { echo "ERROR: ownership scan failed (see stderr)" >&2; exit 1; }

if [[ -n "${bad}" ]]; then
  echo "ERROR: $(printf '%s\n' "${bad}" | wc -l) installed path(s) not root-owned or world/group-writable" >&2
  printf '%s\n' "${bad}" | head -20 | xargs ls -ldn >&2
  exit 1
fi
```

No `2>/dev/null`, no `|| true` on the detection path, `head` only on display, and the count makes truncation explicit.

Tradeoffs worth checking before turning it on: **do not** scan `/usr/local/bin` wholesale — on GitHub macOS runners Homebrew's entries there are owned by `runner`, so a blanket `! -user root` would false-fail; the `-lname '*aerospike/bin/*'` filter is doing real work. I deliberately omitted `! -group wheel`: confirm with `lsbom` whether any pkg intentionally ships `root:admin` before adding it. And a stronger, machine-independent alternative exists — inspect the *artifact* rather than the installed result (`pkgutil --expand`, then compare `lsbom -p MUGf Bom` against the Payload's recorded uid/gid/mode), which also covers customer macOS versions this matrix never runs.

.github/workflows/build-and-release.yml
```diff
@@ -514,6 +517,20 @@ jobs:
           name: signed-artifacts-${{ needs.get-version.outputs.VERSION }}
           path: ./packages
 
+      - name: Verify .pkg is signed and notarized
+        run: |
+          set -euo pipefail
+          pkg=$(find ./packages -name "*-${{ matrix.os }}-${{ runner.arch }}.pkg" | head -1)
```
pvinh-spike: From Architecture agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

The gate verifies *a pkg selected by name pattern on this runner*, not *the pkg set that ships*. Nothing reconciles the two, so the guarantee is narrower than the PR title claims.

**1. The subject is re-derived, not shared.** This selector and its `[[ -n ... ]]` guard now exist twice in the same job — here, and in the pre-existing "Install .pkg locally" step. Two independent evaluations, each taking `head -1` of an unordered `find`, with nothing asserting that exactly one file matches. They agree today only because the source text is identical. Change the naming scheme and update one copy and you get a green pipeline where the verified file and the installed file are different files — the exact failure class this PR exists to close, one edit away. The job already has a cross-step channel and uses it (`echo "ASTOOLS_REINSTALL=..." >> "$GITHUB_ENV"`), so the duplication is not forced by the platform.

**2. `head -1` resolves ambiguity instead of rejecting it.** If two files ever match (a re-run accumulating into an `overwrite`d artifact, a signer emitting both original and re-signed output, or — in the bundle repo — a per-tool pkg carried alongside the bundle pkg, whose names match the same pattern), exactly one is verified and the rest ship unverified. Also missing `-type f`, while this repo's own `build_bundle_mac.sh:43` selector has it.

**3. Coverage is coupled to the test matrix.** Verification of the *n*th pkg happens if and only if `test-install-mac-and-execute`'s matrix has an entry for it. Today `build-mac-artifacts` and this job both list `[macos-15-intel, macos-14, macos-15, macos-26]`, so all four are covered — by two hand-maintained literal lists that nothing cross-checks. Meanwhile `organize-signed-artifacts` ships **every** `*.pkg` it finds, with no matrix at all. Add a build target without adding the test target — exactly the edit someone makes when a new macOS ships — and that pkg is signed-unverified, uploaded to JFrog, and included in the release bundle, fully green. That is the same shape as the `**/*.pkg` bug: coverage silently shrinking with no signal.

**4. Attribution.** A notarization failure is not reported until GPG signing and ~20 Linux container jobs have run, and then it surfaces as `Test install & execute (macos-14)` — a job named for installation tests. The error strings do say "check the notarize-mac-artifacts job", which helps, but the check name in the PR's checks list is what someone triages first.

## Reasoning Chain

1. Grepped all six trees: the identical `find … | head -1` selector appears twice per mac job.
2. `head -1` is a deterministic function only if find's output order is deterministic *and* the match set is a singleton. Neither is asserted.
3. The verify step's selector is parameterised on the *test* matrix; the deploy step's is `-name '*.pkg'` with no matrix. Coverage is the accidental intersection of two independently edited lists.
4. A gate whose subject may differ from the shipped object provides no guarantee about the shipped object — so the assumption must be asserted, not assumed.
5. `pkgutil --check-signature` and `stapler validate` are pure inspection: they need no architecture match, so the per-runner coupling buys nothing.

## Solution

Move signature verification into a dedicated job between the signer and everything else, and let it check *every* pkg:

```yaml
verify-mac-signing:
  name: Verify macOS pkgs are signed & notarized
  needs: [get-version, notarize-mac-artifacts]
  runs-on: macos-15
  steps:
    - download the signed artifact
    - loop over EVERY *.pkg; assert signature + identity + staple on each
    - assert the pkg count is non-zero AND equals the expected count
```

then make `sign-artifacts` depend on it. This decouples coverage from the test matrix by construction, makes the count assertion the real anti-vacuity guard, fails within a couple of minutes of the signer rather than after the whole test tier, and names itself accurately in the checks list.

Keep the root-ownership check where it is — it genuinely needs `sudo installer`, so the install-test job is its correct home. If you prefer the minimal change instead, at least resolve once and hand it over:

```bash
pat="*-${{ matrix.os }}-${{ runner.arch }}.pkg"
n=$(find ./packages -type f -name "${pat}" | wc -l | tr -d ' ')
[[ "${n}" -eq 1 ]] || { echo "ERROR: expected exactly 1 pkg matching ${pat}, found ${n}" >&2; exit 1; }
pkg=$(find ./packages -type f -name "${pat}")
echo "VERIFIED_PKG=${pkg}" >> "$GITHUB_ENV"
```

and use `"${VERIFIED_PKG}"` in the install step, so it can only install what was verified.

Tradeoffs: the dedicated job costs one more mac runner block and serialises `sign-artifacts` behind it. The exact-count assertion needs a source for "expected" — derive it from the build matrix length rather than hardcoding. And it will hard-fail any run that legitimately produces two matching pkgs; that is the desired outcome but may block the first run, so confirm current artifact contents before merging.

.github/workflows/build-and-release.yml
pvinh-spike: Status: still open at `2622e6512`.

Confirmed, and this is the finding I am least able to fully fix — shared-workflows is off the table for this change, and the entitlements gap lives there.

Verified your reading against `entrypoint.sh@v3.7.0`: the codesign call is `codesign --deep --force --options runtime --timestamp --sign` with no `--entitlements`, and `grep -c -- --entitlements` over that file is 0.

**The real fix is available consumer-side, and it is better than an upstream input.** The signer skips binaries already signed with the expected identity:

```bash
if [[ -n $current_authority && $current_authority == *"$SIGNING_IDENTITY"* ]]; then
    echo "    Already signed with correct identity, skipping: $rel"
```

Its own error text says as much: *"Re-sign it with the correct identity during your build step, or leave it unsigned."* So pre-signing embedded binaries in the build job with `--entitlements` (as `pkg/Makefile.mac:35-43` already does) both preserves the entitlements **and** leaves `signed_any=false`, which skips the `cpio` repack entirely — removing the ownership defect at its root rather than detecting it.

I have not implemented that here: it needs the Apple Application cert and keychain setup wired into `build-mac-artifacts` across six repos, and I cannot run this workflow to test it. Landing untested keychain/codesign changes in six release pipelines is the larger risk.

**What I did instead** — make the failure fail closed and legible. Each entrypoint must now execute after install, with the diagnosis naming signing:

```bash
"${b}" --version >/dev/null 2>&1 || "${b}" --help >/dev/null 2>&1 || {
  echo "ERROR: ${b} installed but will not execute -- likely hardened runtime without entitlements" >&2
  codesign -d --entitlements - "${b}" >&2 || true
  exit 1; }
```

This does not prevent the breakage; it stops it shipping and points at the cause instead of surfacing as a bats failure. Pre-signing remains the fix and is flagged in the PR description as the blocking follow-up before a real release.

.github/workflows/build-and-release.yml
pvinh-spike: Resolved in `2622e6512` — Fixed. The assertion now uses the full configured identity including the Team ID, pinned to the leaf position, as a fixed string:

```bash
printf '%s\n' "${sig}" | grep -qF "1. ${INSTALLER_IDENTITY}" || { ... }
```

`INSTALLER_IDENTITY` is a step-level `env:` carrying the same literal as `installer-identity` on the notarize job, with a comment noting they must stay in sync — a reusable-workflow `with:` cannot read `env`, so single-sourcing it is not available. `-qF` also removes the accidental regex in `Inc.`, and the failure path now prints the captured certificate chain instead of one line.

.github/workflows/build-and-release.yml
pvinh-spike: Resolved in `2622e6512` — Fixed. Added the assessment a customer's Mac actually applies, as the gate:

```bash
spctl --assess --type install -vv "${pkg}" || { ...; exit 1; }
```

`pkgutil` and `stapler validate` are kept for their better diagnostics. I did **not** grep for `Status: signed by a certificate trusted by macOS` — as you noted, that wording has drifted across releases and this matrix spans macos-14 to macos-26, so a literal match would false-fail on some runners; `spctl`'s exit code covers signature, trust, notarization, and identity without that fragility.

Still unproven on a real runner: `spctl` may surface pre-existing failures on older images, and it is a network dependency. It is a hard failure by design, never `|| true`.

.github/workflows/build-and-release.yml
pvinh-spike: Resolved in `2622e6512` — Fixed, all three defects.

**Scope** — both payload roots are scanned now. You were right that `/usr/local/bin` was uncovered; I confirmed the symlinks in every `pkg/Makefile prep-mac` and in the bundle's `Makefile.mac:91-97`. Filtered with `-lname '*aerospike/bin/*'` rather than scanning the directory, because Homebrew owns unrelated entries there on hosted runners and a blanket scan would false-fail.

**Predicate** — widened to mode bits: `\( ! -user root -o -perm -0002 -o -perm -0020 \)`, so `root:staff 0775` and `o+w` are caught. I left `! -group wheel` out deliberately: it needs an `lsbom` check against the current pkgs first, which I cannot run here.

**Vacuity** — `2>/dev/null` and `|| true` are gone from the detection path. `head` now applies only to display, so no SIGPIPE needs absorbing and a `find` that cannot walk the tree fails the step. Added an explicit non-vacuity assertion (`scanned >= 1`) and replaced the directory-only guards with a per-entrypoint existence loop.

Exercised against a sandbox tree for pass, missing-entrypoint, vacuous-scan, and world-writable cases — the pass path reaches all three `==>` assertions and exits 0; each failure case exits 1.

.github/workflows/build-and-release.yml
pvinh-spike: Resolved in `2622e6512` — Partly fixed.

**Fixed — the subject is now single-sourced.** The pkg is resolved once, an exact-count assertion replaces `head -1`, `-type f` is added, and the path is exported as `VERIFIED_PKG`; the install step consumes it, so it can only install what was verified:

```bash
n=$(find ./packages -type f -name "${pat}" | wc -l | tr -d ' ')
[[ "${n}" -eq 1 ]] || { ...; exit 1; }
pkg=$(find ./packages -type f -name "${pat}")
echo "VERIFIED_PKG=${pkg}" >> "$GITHUB_ENV"
```

**Not fixed — the coverage half stays open.** Verification is still coupled to the test matrix, so adding a build target without a matching test target would still ship that pkg unverified. Your `verify-mac-signing` job is the right answer and I have not implemented it: it is a new job plus a `needs:` re-wiring on `sign-artifacts`, and I cannot execute this workflow to test it. Tracking it rather than landing an untested job in six release pipelines. Flagged in the PR description.

pvinh-spike:
Applying the review follow-up from @mcoberly2 on the bundle PR ([citrusleaf/aerospike-tools#440](https://github.com/citrusleaf/aerospike-tools/pull/440#issuecomment-5243227527)) here, so the six repos stay identical.

**1. One `find`, not three.** Resolving the `.pkg` walked `./packages` three separate times — the count, the error message and the result — which could in principle disagree. Now captured once into a variable.

`mapfile` was the suggested reader, but it isn't available: the macOS runners ship **bash 3.2.57**, which predates the builtin (bash 4.0).

```
$ /bin/bash -c 'mapfile -t x < <(echo hi)'
/bin/bash: mapfile: command not found
```

**2. `-perm +111` kept deliberately, now commented.** The suggested `/111` is **GNU** find syntax. macOS `/usr/bin/find` is BSD find, whose grammar is `-perm [-|+]mode`, and it rejects `/111` outright:

```
$ /usr/bin/find /tmp/permtest -type f -perm /111
find: -perm: /111: illegal mode string
$ /usr/bin/find /tmp/permtest -type f -perm +111
/tmp/permtest/a
```

These steps only ever run on macOS runners, so `+111` is correct. Commented so it doesn't get "modernized" later.

**3. The nit surfaced a real defect underneath it, which is fixed.** The scan read from a process substitution:

```bash
done < <(find /usr/local/aerospike -type f -perm +111)
```

There, a `find` that errors out yields an **empty read loop, not a failure** — so `==> every embedded Mach-O is signed` would print for a scan that walked nothing. Had `/111` been applied, that is exactly what would have happened: a green step attesting to something it never checked, the same silent-pass failure mode as the original glob bug. It now assigns under `set -e` and asserts non-empty before trusting the result.

Verified: workflow YAML parses; every `run:` block passes `bash -n` under bash 3.2; the pkg-resolution and exec-scan paths were exercised against the real `/usr/bin/find` for the populated, empty and missing-directory cases.

mcoberly2:
Minor
Patch 5 commit message cites /usr/local/aerospike/bin/asbench in the failure log — copy-paste from another repo’s run; harmless but confusing in admin history.

mcoberly2:
**non-blocking:**
ERROR: vs ::error:: — cross-repo consistency with tools #440.
Mach-O comment referencing PyInstaller strip is correct here (unlike aql/benchmark where it was misplaced).

pvinh-spike:
Follow-up from @mcoberly2's review on the bundle PR ([citrusleaf/aerospike-tools#440](https://github.com/citrusleaf/aerospike-tools/pull/440#issuecomment-5244902864)). This repo is the mixed case, so one comment changed and one deliberately did not.

**Fixed — hardened-runtime comment.** It said *"This repo's legacy mac path signed with disable-library-validation (and allow-unsigned-executable-memory for asadm's PyInstaller tree)"*. The entitlement requirement is real and asadm is exactly the binary it applies to, but the path that applies it is the **aerospike-tools** bundle build — `pkg/entitlements` and `pkg/asadm-entitlements` live there. This repo has no legacy mac signing path, no entitlements file, and no occurrence of `disable-library-validation` anywhere, so "this repo's" pointed the reader at something that isn't here. Reworded to keep the justification and name the right repo.

**Left alone — unsigned-Mach-O comment.** mcoberly2 flagged the PyInstaller reference as stale, which it is in the single-binary tool repos. Here it is accurate and I verified it before leaving it:

- `pkg/Makefile:136` runs `$(PKG_DIR)/strip-foreign-signatures.sh $(MAC_ROOT)` over the staged mac root
- `pyinstaller-build.spec:37` is `binaries = [('/usr/bin/less','.')]` — Apple-signed `/usr/bin/less`, which is the foreign signature being stripped

Comment-only; behavior unchanged. Verified: workflow YAML parses and every `run:` block still passes `bash -n` under bash 3.2.

mcoberly2 (approved):

PR #538, opened 2026-08-14
fix(pkg): name macOS pkgs for Homebrew; verify the version the tools report

pvinh-spike:
Three problems in the naming layer, plus the version checks that would have caught them.

1. **macOS `.pkg` names were not Homebrew-addressable.** `pkg/Makefile` emitted `aerospike-asadm-<mac-version>-<arch>.pkg`, which CI renamed using the GitHub runner label (`macos-15-intel`) and `runner.arch` (`ARM64`, `X64`). Neither names the target platform, both are upper case, and both change when GitHub renames a label. A cask's `arch` stanza yields `arm64` / `x86_64`, so a cask `url` could not address those files.
2. **The name was composed twice**, once in `pkg/Makefile` and once in the workflow's `mv`. Two copies of one convention is two places for it to drift.
3. **Nothing checked the version the shipped binaries report.** The smoke tests ran `asadm --help` and `asinfo --help` and asserted only the exit status.

`rcN` already folds into the package iteration here (#530), so this PR is the naming and verification half.

## What changed

**macOS `.pkg` names are emitted by `pkg/Makefile` alone** (the CI rename is gone):

```
aerospike-asadm-<version>-<iteration>-macos<major>-<arch>.pkg
```

Lower case, `-` separated, `<arch>` from `uname -m`, so a cask can interpolate it. `macos<major>` comes from `os_version.sh` — extended here to Darwin — and reports the macOS generation the artifact was built on -- nothing pins `MACOSX_DEPLOYMENT_TARGET`, so it is not a minimum, it is what keeps one release's per-runner artifacts distinct., and the field that keeps names unique when one release ships a pkg per (macOS major, arch) pair.

The workflow now moves whatever `make` produced and asserts there is exactly one `.pkg` under `pkg/target`, so a stale target directory can't ship silently. The signed-artifact selection matches on the shipping name's own platform fields rather than a runner label, which additionally asserts the artifact really was built for the platform it claims. The unused `osx-dmg` target is dropped.

**`.tar` field order** now matches the others: `<version>-<iteration>-<distro>-<arch>`.

**`get-version` emits two strings**, so a dev build's package names carry the sha and nothing else:

| output | meaning | consumers |
|---|---|---|
| `VERSION` | release identity | git tag, JFrog bundle, build number, artifact containers |
| `BUILD_VERSION` | what artifacts are **named** and asadm/asinfo are **stamped** from | `pkg/Makefile`, `docker-build.sh`, the version sed |

On master the two are the same string. On a dev branch `BUILD_VERSION` drops the branch slug and the `rcN` and keeps the sha, which leaves the iteration at 1. Before, a dev build named its packages from the whole string — branch slug and `rcN` included — when the sha alone already pins the commit. `get-version` was also missing a `Checkout`, so its `pkg_release.sh` step could not have run. Added.

**The `VERSION` file is unchanged.** It still carries the release candidate (`5.0.3-rc3`), written by the Bump Version workflow. An earlier revision of this PR moved the rc number into the tags and stripped it from the file; that is reverted, so this PR makes no version change and burns no rc number.

**asadm and asinfo report the build attempt.** Both already split the embedded string and print the core on the `Version` line, so the `Build` line is the only place the attempt shows:

| stamped | reported |
|---|---|
| `5.0.3-rc3` | `Version 5.0.3` / `Build rc3` |
| `5.0.3` | `Version 5.0.3` |
| `5.0.3-abc123def` | `Version 5.0.3` / `Build abc123def` |

**Smoke tests verify the version.** They now compare both printed lines against the `VERSION` file, per tool, and assert the *absence* of a `Build` line for a bare version, which is how a binary that fell back to `git describe` gives itself away. CI passes `EXPECTED_VERSION=BUILD_VERSION`, cross-checked against the `VERSION` file on `MAJOR.MINOR.PATCH` so a step wired to the wrong workflow output fails rather than testing against whatever it was handed. Assertions live in `version_lib.sh`, shared by the bats tests and a new non-bats `test_execute.sh`.

**A reused version fails before anything is built.** The tag-exists check also runs in `validate-inputs` now, gated on a tagging run. It already existed in the tagging job, but that runs after the packages and the release bundle are published — by then the previous build's files in the flat deb/rpm pools have been overwritten. A `concurrency` group serialises runs on a ref, with `cancel-in-progress: false` since a half-published run must finish.

## Naming, end to end

One cycle of `5.0.3`. Generated output: the real `get-version` logic run for master and dev refs, fed through `pkg/Makefile` (`make -n`).

### Release builds

| | rc1 | rc2 | rc3 → **GA** |
|---|---|---|---|
| `VERSION` file | `5.0.3-rc1` | `5.0.3-rc2` | `5.0.3-rc3` |
| git tag, JFrog bundle | `5.0.3-rc1` | `5.0.3-rc2` | `5.0.3-rc3`, + `5.0.3` at GA |
| `asadm` / `asinfo --version` | `Version 5.0.3` / `Build rc1` | … `Build rc2` | … `Build rc3` |
| deb | `aerospike-asadm_5.0.3-1ubuntu24.04_arm64.deb` | `…-2ubuntu24.04…` | `…-3ubuntu24.04…` |
| rpm | `aerospike-asadm-5.0.3-1.el9.arm64.rpm` | `…-2.el9…` | `…-3.el9…` |
| tar | `aerospike-asadm-5.0.3-1-ubuntu24.04-arm64.tar` | `…-2-…` | `…-3-…` |
| macOS pkg | `aerospike-asadm-5.0.3-1-macos15-arm64.pkg` | `…-2-…` | `…-3-…` |
| docker | `aerospike-asadm:5.0.3-1` | `:5.0.3-2` | `:5.0.3-3` |

Nothing is rebuilt or renamed at GA: the promoted `5.0.3-3` artifacts are what ship.

### Dev branches

`VERSION` `5.0.3-rc3`, branch `mybranch`, commit `abc123def`:

| | value |
|---|---|
| git tag, JFrog bundle | `5.0.3-rc3-mybranch-abc123def` |
| `asadm` / `asinfo --version` | `Version 5.0.3` / `Build abc123def` |
| deb | `aerospike-asadm_5.0.3-abc123def-1ubuntu24.04_arm64.deb` |
| rpm | `aerospike-asadm-5.0.3_abc123def-1.el9.arm64.rpm` |
| tar | `aerospike-asadm-5.0.3-abc123def-1-ubuntu24.04-arm64.tar` |
| macOS pkg | `aerospike-asadm-5.0.3-abc123def-1-macos15-arm64.pkg` |
| docker | `aerospike-asadm:5.0.3-abc123def-1` |

The sha alone makes a dev build unique. The branch stays in the git tag and bundle version. rpm mangles `-` to `_` in its version field, as it always has.

### Before / after

| | before | after |
|---|---|---|
| macOS pkg | `aerospike-asadm-5.0.3.3-arm64.pkg`, renamed to `-macos-15-ARM64.pkg` | `aerospike-asadm-5.0.3-3-macos15-arm64.pkg` |
| tar | `aerospike-asadm-ubuntu24.04-5.0.3-arm64.tar` | `aerospike-asadm-5.0.3-3-ubuntu24.04-arm64.tar` |
| deb, dev branch | `aerospike-asadm_5.0.3-mybranch.abc123def-1ubuntu24.04_arm64.deb` | `aerospike-asadm_5.0.3-abc123def-1ubuntu24.04_arm64.deb` |

## Failure modes closed

Three cases where a broken platform token would have shipped rather than failed:

- `os_version.sh` had no `set -e`, so an `sw_vers` that exits 0 but prints nothing emitted the bare token `macos`. Non-empty, so `prep-mac`'s guard waved it through and the release would ship a `.pkg` missing the one field that suffix exists to carry. It now exits non-zero.
- The macOS workflow steps called `os_version.sh` in a command substitution used as an *argument*. `set -e` only inspects the enclosing command's status, so a failing script was discarded and the leg passed, publishing an artifact container with an empty platform field. The token is assigned on its own line now, where the status does propagate.
- `RELEASE` and `MAC_PLATFORM` were recursively expanded, so `prep-mac`'s guard and the file name were separate evaluations of the same variable — the exact split the guard exists to close. Both are `:=` now, matching `PKG_VERSION`.

## Verification

- `make -n deb rpm tar osx-pkg` compared against `origin/master` for GA and rc versions — the before/after table is that output.
- `get-version` simulated for master, master + `skip_tagging`, and a dev ref; each result fed through `pkg/Makefile`.
- `get_version()` extracted verbatim from `asadm.py` and exercised over rc / GA / dev / `git describe` inputs; the reported-version table is that output.
- Tag guard: a reused version on master fails; dev branch, `skip_tagging`, and the next rc all pass.
- Smoke tests run under bats 1.14 and without bats. rc, dev, and bare versions pass; rc2-vs-rc3, bare-vs-rc, and a `git describe` binary reporting the previous tag all fail with the reason.
- `os_version.sh` verified against the failing input: an empty `sw_vers` now exits 1, and the old `set -e` pattern demonstrably swallowed a non-zero status where the new one propagates it.
- Workflow YAML parses; `bash -n` clean on all changed shell, including the `validate-inputs` script extracted from the YAML.
- No stale consumers of the old `.pkg` / `.tar` names remain: the JFrog upload and organize steps select by extension.

## Coordination

Part of a five-repo change; the bundle picks up these names. It carries a transitional fallback to the pre-rename component names, so it keeps building while these merge and cut releases — it does not need to merge first or last.

- [ ] aerospike/aerospike-tools-backup#172
- [ ] aerospike/aql#93
- [ ] aerospike/aerospike-benchmark#163
- [ ] aerospike/asconfig#136
- [ ] citrusleaf/aerospike-tools#443 (bundle)
- [ ] aerospike/aerospike-admin#538


pvinh-spike:
**PR Review (version 2026.07.31)** — updated after fixes pushed in `0599bba`

| Gate | Status |
|------|--------|
| Contract | pass |
| Impacted surface | blind spot |
| Failure paths | pass |
| Evidence | pass |
| Quality | pass |
| Backward Compat | pass |

**8 findings** were posted inline (0 Critical, 5 Important, 3 Minor). **7 fixed and resolved**; 1 partly fixed and left open.

| Finding | Status |
|---------|--------|
| Tests: `expected_version()` core-only cross-check → demonstrated false PASS on an rcN mismatch | Addressed in `9768594` |
| Control Flow: tag guard omits `inputs.release`, blocks the documented master dry run | Addressed in `3d0a3f5` |
| Data Flow: `prep-mac` guards 1 of 3 computed name fields; empty VERSION ships `aerospike-asadm---macos26-arm64.pkg` | Addressed in `b4bc48f` |
| Logic: `?=` loses the set-but-empty VERSION fallback, stamps `""` silently | Addressed in `b4bc48f` |
| Tests: none of the new assertion code runs on a PR; the git-describe branch cannot fail | **Partly** — coverage added in `05d92e0`; `test_execute.sh` still unwired |
| Architecture: `os_version.sh`'s "single source of truth for the platform field" overstates what it emits | Addressed in `0599bba` |
| Style: Homebrew example omits `<iteration>`, hardcodes `macos15` | Addressed in `0599bba` |
| Architecture: `asinfo.py` copies asadm's split without the `$$` fallback | Addressed in `0599bba` |

**The three blockers are closed.**

The false PASS is fixed by porting `aerospike-tools-backup`'s `case` form verbatim: `EXPECTED_VERSION=5.0.3-rc9` against a `5.0.3-rc3` VERSION file now returns 1, while the legitimate dev-build sha swap still returns 0. The tag guard now reads `create-tag`'s condition verbatim, verified over all eight release/ref/skip_tagging combinations against a repo with `5.0.3-rc3` present — exactly one blocks, and it is the tagging run. `prep-mac` now guards `PKG_VERSION` and `ITERATION` alongside `MAC_PLATFORM`, and the root `makefile` falls back on an empty `VERSION` instead of stamping nothing.

**The Evidence gate moves, which is the part worth reading.** `.github/bin/test/version_lib_test.sh` is ported from asconfig and run by a new `naming-scripts` job in `build-check.yml`, which already triggers on `pull_request` and `push`. It asserts against strings — no container, no package, no bats — and covers the two branches that previously could not fail, including the `git describe` case that is unreachable from either `get-version` path. 21/21 pass, and I confirmed it goes red when either fixed bug is reintroduced:

```
$ # core-only comparison restored
FAIL: a same-core different-rcN fails (wanted exit 1, got 0)
1 version_lib.sh assertion(s) behaved wrongly
```

**What I deliberately did not do**, each noted in the relevant thread rather than silently skipped:

- **`test_execute.sh` is still invoked by nothing** — the one finding left open. Wiring it needs shim binaries on `PATH`, which is a fixture-design choice; deleting it would break the "same file in all five repos" property its header claims. Restoring asconfig's dropped "assert first, then report" comment is worth doing either way.
- **`os_version.sh` still emits only `macos<major>`.** I narrowed the comment rather than consolidating, because the other four coordinated repos emit the bare token too — asadm alone would become the odd one out. Worth doing across all five, separately. The reciprocal misstatement in the bundle's `mac_platform.sh` is corrected in citrusleaf/aerospike-tools#443.
- **The cask still has to pin a build generation.** Publishing a deployment-target-pinned stable name or an alias is a change to what the release publishes, not a fix to this branch.
- **`pkg_release.bats`** — asadm has no equivalent of asbench's, and this branch adds a new `pkg_release.sh` call site. The new job is the natural home.
- **The `|| true` on the four sed lines** hides genuine substitution failures, but removing it changes failure behaviour for both build legs and deserves its own change.

Verification on the fixes: `shellcheck` clean on all four shell files, both workflow YAMLs parse, Python parses, and the emitted names are byte-identical to before for rc, GA and dev versions (`aerospike-asadm-5.0.3-3-macos26-arm64.pkg`, `…-5.0.3-1-…`, `…-5.0.3-abc123def-1-…`, tar `aerospike-asadm-5.0.3-3-ubuntu24.04-arm64.tar`). Action pins in the new job reuse the ones `installer-build` already carries in the same file, both verified against upstream.

The `Impacted surface` blind spot is unchanged and not closable from here: the four pinned `aerospike/shared-workflows@v3.7.0` reusable workflows are not in this repo, so I could not confirm their `version:` / `jf-build-id:` inputs never construct a file path that must agree with a `BUILD_VERSION`-derived name. On master the two strings are identical, so any such coupling would only misfire on a dev-branch publish.

One inaccuracy in the description rather than the code, left as-is: it says `get-version` "was also missing a `Checkout` … Added." The diff adds no such step — it was already present, and appears only as a context line.

.github/bin/test/version_lib.sh
```diff
@@ -0,0 +1,75 @@
...
+		echo "EXPECTED_VERSION is set but empty -- the step that supplies it produced nothing" >&2
+		return 1
+	fi
+	expected="${EXPECTED_VERSION:-$file_version}"
+	if [ -z "$expected" ]; then
+		echo "no VERSION file at $version_file and no EXPECTED_VERSION set" >&2
+		return 1
+	fi
+	if [ -n "${EXPECTED_VERSION:-}" ] && [ -n "$file_version" ] &&
+		[ "${EXPECTED_VERSION%%-*}" != "${file_version%%-*}" ]; then
```
pvinh-spike: From Tests agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

The EXPECTED_VERSION / VERSION-file cross-check compares only `%%-*` — the MAJOR.MINOR.PATCH core — so it cannot tell rc3 from rc9. That is the exact axis this PR makes load-bearing: `rcN` becomes the package iteration, so two artifacts of one release differ by nothing else. A run wired to the wrong workflow output that happens to share a core is validated against the wrong claim and reports green.

Reproduced side by side, same inputs, VERSION file `5.0.3-rc3`:

```
--- asadm version_lib.sh : EXPECTED_VERSION=5.0.3-rc9 ---
  5.0.3-rc9
    rc=0                      <- false PASS
--- asbackup version_lib.sh : EXPECTED_VERSION=5.0.3-rc9 ---
  EXPECTED_VERSION '5.0.3-rc9' and VERSION file '5.0.3-rc3' disagree
    rc=1
```

Both correctly still admit the one legitimate divergence, a dev build swapping the file's rcN for a sha:

```
--- both libs : EXPECTED_VERSION=5.0.3-abc123def ---
  5.0.3-abc123def
    rc=0
```

The full bats suite exits 0 under the rc9 wiring, so nothing downstream catches it either.

This is a silent weakening of a contract the coordinated siblings already implement. `aerospike-tools-backup`'s copy of this same file carries the stronger `case` form under a comment naming this case verbatim: "Compare the whole identifier, not just MAJOR.MINOR.PATCH: the rcN is what becomes the package iteration, so a core-only check cannot tell rc9 from rc1."

`.github/bin/test/README.md` states the stronger claim the code does not implement — "so a run wired to the wrong workflow output fails rather than testing against whatever it was handed". A wrong output sharing a core does not fail.

Secondary, same file, in `expected_version_lines`: a malformed expectation ending in `-` emits the literal line `Build ` (trailing space) and then blames the binary — `expected 'Build ' (from 5.0.3-)`. `asbench`'s copy guards it explicitly and asadm's dropped the guard, with `asbench`'s comment giving the reason: "Emitting the literal 'Build ' here would blame the binary for a malformed expectation." Low reachability, but it is unexplained drift from a file whose own header claims to match what the sibling repos ship.

## Reasoning Chain

1. `[ "${EXPECTED_VERSION%%-*}" != "${file_version%%-*}" ]` strips everything from the first `-`, discarding the rcN — the only field distinguishing two artifacts of one release.
2. Executed both libraries against identical inputs rather than reading them: asadm's returns 0, asbackup's returns 1. Then confirmed the false PASS is end to end by running the whole bats suite under that wiring — exit 0.
3. Confirmed the fix does not over-reject by running the dev-build case (`5.0.3-abc123def` against file `5.0.3-rc3`) through asbackup's form: rc=0, as intended.
4. `diff -u` against `asbench`'s copy showed the empty-trailing-field guard is the one behavioural (non-comment) deletion in asadm's version.

## Solution

Port asbackup's `case` form verbatim, so the copies stay one contract rather than diverging further:

```sh
if [ -n "${EXPECTED_VERSION:-}" ] && [ -n "$file_version" ]; then
	case "$EXPECTED_VERSION" in
	"$file_version") ;;
	"${file_version%%-*}"-rc*)
		echo "EXPECTED_VERSION '$EXPECTED_VERSION' and VERSION file '$file_version' disagree" >&2
		return 1
		;;
	"${file_version%%-*}"-*) ;;
	*)
		echo "EXPECTED_VERSION '$EXPECTED_VERSION' and VERSION file '$file_version' disagree" >&2
		return 1
		;;
	esac
fi
```

and restore asbench's four-line empty-build-field guard in `expected_version_lines`. Then correct the README sentence to say the whole identifier must agree except for the dev-build sha swap.

I did not check whether any caller relies on the current permissiveness — nothing in this repo passes a same-core different-rcN EXPECTED_VERSION deliberately.

.github/workflows/build-and-release.yml
```diff
@@ -96,9 +118,20 @@ jobs:
...
-            echo "::error::VERSION '$VERSION' is not valid semver (expected MAJOR.MINOR.PATCH or MAJOR.MINOR.PATCH-suffix)." >&2
+            echo "::error::VERSION '$VERSION' is not valid semver (expected MAJOR.MINOR.PATCH or MAJOR.MINOR.PATCH-suffix, e.g. 5.0.3-rc4)." >&2
+            exit 1
+          fi
+
+          # Reject a reused version before anything is built. create-tag checks
+          # this too, but only after the packages and the release bundle are
+          # published -- by then the previous build's files in the flat deb/rpm
+          # pools have already been overwritten.
+          if [[ "${GITHUB_REF}" == "refs/heads/master" && "${SKIP_TAGGING}" != "true" ]] \
```
pvinh-spike: From Control Flow agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

The guard is gated on two conjuncts; `create-tag` — the only job that consumes the tag name — is gated on three (line 1065):

```
if: inputs.release == true && github.ref == 'refs/heads/master' && inputs.skip_tagging != true
```

`inputs.release == true` is missing here, and is not smuggled in via another env var (`grep -n "inputs.release"` first hits line 1065). So a run that will never tag and never publish is aborted before anything is built.

This fires on the default invocation, not a corner case:

1. `VERSION` is `5.0.3-rc3` — unchanged by this PR, per its own promise 7.
2. `refs/tags/5.0.3-rc3` already exists on origin (`git ls-remote --tags origin` → `7956be5e refs/tags/5.0.3-rc3`), and the `Checkout` here now uses `fetch-depth: 0` / `fetch-tags: true`, so it is definitely present locally when the check runs.
3. Someone opens Actions → Run workflow on `master`, leaving both inputs at their defaults: `release=false`, `skip_tagging=false`.
4. `GITHUB_REF == refs/heads/master` and `SKIP_TAGGING` renders as `false`, so both conjuncts hold and `git rev-parse -q --verify refs/tags/5.0.3-rc3` succeeds.
5. The run fails with `Tag '5.0.3-rc3' already exists. Bump VERSION …` before a single package is built.

That is the documented dry-run: the workflow header says "Dispatch with `release=false` from any branch: builds run end-to-end, artifacts available on the GHA run page, no tag, no JFrog publish, no release bundle." And because `VERSION` is only bumped by a separate PR *after* a release is tagged, "tag exists and VERSION still points at it" is the steady state for most of the release cycle — so master dry-runs are broken most of the time. The only workaround is `skip_tagging=true`, which additionally flips `get-version` onto the branch-slug/sha path, so the dry run no longer exercises release naming.

The guard's own comment states its purpose: "by then the previous build's files in the flat deb/rpm pools have already been overwritten." A `release=false` run writes nothing to those pools — `upload-artifacts-to-jfrog` and `create-release-bundle` are both gated on `inputs.release == true`. The stated hazard is absent from the case it newly blocks.

Same omission as citrusleaf/aerospike-tools#443, which shipped the identical two-conjunct guard against a three-conjunct tagging job.

## Reasoning Chain

1. Decomposed `create-tag`'s `if:` at line 1065 into its three conjuncts, then the new guard into its two. Difference: exactly `inputs.release == true`.
2. Confirmed `inputs.release` is not referenced anywhere in `validate-inputs` — the step's `env:` passes only `SKIP_TAGGING`.
3. Checked what a `release=false` run actually writes, to see whether the guard's rationale covers it: the JFrog upload and the release bundle are both `inputs.release`-gated, so nothing reaches the pools.
4. Checked whether this is latent or immediate: read `VERSION` (`5.0.3-rc3`) and confirmed the tag exists upstream. Immediate.
5. Confirmed `skip_tagging=true` is not a clean workaround by tracing it into the `get-version` compute step.

## Solution

Add the missing conjunct so the guard's domain is exactly `create-tag`'s:

```yaml
        env:
          RELEASE: ${{ inputs.release }}
          SKIP_TAGGING: ${{ inputs.skip_tagging }}
```
```bash
          if [[ "${RELEASE}" == "true" && "${GITHUB_REF}" == "refs/heads/master" \
                && "${SKIP_TAGGING}" != "true" ]] \
             && git rev-parse -q --verify "refs/tags/${VERSION}" >/dev/null; then
```

`inputs.release` is a boolean input, so it renders as the literal `true`/`false` in `env:` — string-equal to `create-tag`'s expression-context `inputs.release == true`, and matching how `SKIP_TAGGING` is compared one line over.

This only narrows the guard, so nothing that passes today starts failing, and the real tagging run keeps both this early check and `create-tag`'s late re-check.

One edge worth a thought: the workflow is `workflow_dispatch`-only today, so `inputs.release` is always present. If a `push:` or `workflow_call:` trigger is ever added, `${{ inputs.release }}` renders empty and the guard would silently stop firing. `== "true"` keeps that on the permissive side, matching `create-tag`; if you would rather fail closed there, invert to `!= "false"`.

pkg/Makefile
```diff
@@ -124,6 +155,7 @@ prep:
 # Staging for macOS installer: matches root makefile Darwin INSTALL_ROOT and symlinks.
 .PHONY: prep-mac
 prep-mac:
+	@test -n "$(MAC_PLATFORM)" || { echo "ERROR: empty macOS platform token -- .github/bin/os_version.sh failed; refusing to build an unnamed pkg" >&2; exit 1; }
```
pvinh-spike, line 144: From Data Flow agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

`MAC_PKG` interpolates three computed fields — `$(RELEASE)`, `$(MAC_PLATFORM)`, `$(ARCH)` — and this guard checks one. Its message says "refusing to build an unnamed pkg", but two thirds of the name can be empty and it still passes.

`RELEASE` is built from two `$(shell pkg_release.sh …)` calls, and `$(shell)` swallows a non-zero exit into the empty string — precisely the hazard the diff's own `os_version.sh` comment warns about a few files over. `pkg_release.sh ''` exits 1 with a usage message, so `PKG_VERSION=`, `ITERATION=`, `RELEASE=-`.

Verified in the review tree on macOS:

```
$ make -C pkg -n osx-pkg VERSION=
... target/aerospike-asadm---macos26-arm64.pkg

$ make -C pkg -n osx-pkg VERSION=5.0.3-rc3
... target/aerospike-asadm-5.0.3-3-macos26-arm64.pkg
```

The guard passes because `MAC_PLATFORM` is `macos26` — the one field that was fine. `prep-mac` then runs, `pkgbuild` succeeds, and the new CI step moves "whatever make produced" without re-deriving or validating the name — that is the point of that step, so nothing between here and notarization inspects it. The bundle's primary glob `aerospike-asadm-*-macos26-arm64.pkg` even matches the malformed name, so it would be composed into the bundle rather than rejected.

Reachability: locally `make -C pkg osx-pkg VERSION=` hits it immediately. In CI it needs `get-version`'s `build_version` output to arrive empty, which an unresolvable `needs.<job>.outputs.<name>` does render as — see the related finding on the root `makefile`, where the same empty value silently stamps an unversioned binary. The two share a root cause: an empty version is treated as a value rather than an error.

## Reasoning Chain

1. Read `MAC_PKG` at line 71 and counted its computed components: three.
2. Read the guard: it tests one of them, while its error message speaks for the whole name.
3. Traced how `RELEASE` could go empty — `pkg_release.sh` exits non-zero on an empty argument, and `:=` with `$(shell)` captures stdout while discarding the status.
4. Ran `make -C pkg -n osx-pkg VERSION=` rather than reasoning about it, and got the malformed basename above.
5. Checked whether anything downstream would catch it: the new "move whatever make produced" step deliberately does not reconstruct the name, and the bundle's glob is permissive enough to match.

## Solution

Guard the whole computed name rather than one field:

```make
prep-mac:
	@test -n "$(PKG_VERSION)" && test -n "$(ITERATION)" || { echo "ERROR: empty version/iteration -- pkg_release.sh failed for VERSION='$(VERSION)'" >&2; exit 1; }
	@test -n "$(MAC_PLATFORM)" || { echo "ERROR: empty macOS platform token -- .github/bin/os_version.sh failed; refusing to build an unnamed pkg" >&2; exit 1; }
```

Asserting the assembled basename matches the declared five-field shape would be stronger still, and would also catch a future field reorder — but it has to tolerate the dev-build case, where `PKG_VERSION` itself carries a hyphen (`5.0.3-abc123def`), so a naive field count would misfire. The two `test -n` lines are the safe version.

Worth noting separately: the guard's comment at line 68 says `MAC_PLATFORM` is "captured once (`:=`), so prep-mac's guard below checks the very token that" lands in the name. That part is true and verified — stubbing `os_version.sh` to `exit 1` does abort at `prep-mac`. The gap is only that the other two fields have no equivalent.

makefile
```diff
@@ -37,6 +37,9 @@ VERSION_DEFAULT := $(shell git describe --tags --always --abbrev=9)
 else
 VERSION_DEFAULT := $(shell tr -d '[:space:]' < $(SOURCE_ROOT)/VERSION)
 endif
+# Embedded whole: asadm/asinfo split it, so 5.0.3-rc3 reports
+# "Version 5.0.3 / Build rc3".
+VERSION ?= $(VERSION_DEFAULT)
```
pvinh-spike, line 42: From Logic agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

Replacing the shell-level `$${VERSION:-$(VERSION_DEFAULT)}` with make's `?=` is not an equivalent rewrite. `:-` substitutes on unset **or empty**; make's `?=` assigns only when the variable is *undefined*, and an environment variable set to the empty string is defined. So a set-but-empty `VERSION` now stamps the binaries with nothing, where before it fell back to the VERSION file.

Verified both forms in isolation:

```
VERSION ?= default          VERSION_DEFAULT + shell :-
  unset -> [default]          unset -> [default]
  empty -> []                 empty -> [default]     <- the regression
  1.2.3 -> [1.2.3]            1.2.3 -> [1.2.3]
```

and against the real makefile:

```
$ make -n one-dir
  sed ... "s/[$][$]__version__[$][$]/5.0.3-rc3/g" ...
$ VERSION= make -n one-dir
  sed ... "s/[$][$]__version__[$][$]//g" ...          <- stamps ""
```

`""` does not `startswith("$$")`, so `asadm.py`'s `get_version()` returns `("", "")` and asadm prints a bare `Version ` with no `Build` line — it does **not** fall back to `DEFAULT_ASADM_VERSION`. The `|| true` on each sed means nothing surfaces at build time.

The macOS build step feeds a workflow output straight into this:

```yaml
      - name: Build asadm
        env:
          VERSION: ${{ needs.get-version.outputs.BUILD_VERSION }}
        run: |
          make one-dir
```

An unresolvable `needs.<job>.outputs.<name>` renders as the empty string, not an error — which is exactly the failure shape this PR's own `version_lib.sh` was written to defend against ("the step that supplies it produced nothing"). `.github/bin/build_package.sh` guards it correctly on the Linux leg with `if [[ -n "${PKG_VERSION:-}" ]]`; the makefile is now the one place that lost the guard.

In practice CI is self-defending today — the smoke tests read the same `BUILD_VERSION`, so an empty value trips `version_lib.sh`'s set-but-empty check two jobs later. So this is a robustness gap in the stamping path rather than a live shipping defect. But a local `VERSION= make install` ships a blank-stamped binary silently, and no test covers the makefile's stamping at all.

## Reasoning Chain

1. Compared the two constructs directly in a scratch makefile rather than from memory: `?=` and `:-` differ on exactly the empty case.
2. Confirmed the difference reaches the real recipe by running `make -n one-dir` three ways.
3. Traced an empty stamp through `asadm.py:1011-1018` — the `startswith("$$")` fallback does not catch `""`, so the default is not restored.
4. Checked the CI call sites to size the impact: both the mac build and the install test derive from the same output, so a real occurrence trips the smoke test rather than shipping — hence important, not critical.
5. Confirmed the Linux path already has the guard the makefile dropped, so this is an inconsistency introduced by the rewrite, not a pre-existing gap.

## Solution

Restore the old semantics explicitly, right after the `?=`:

```make
VERSION ?= $(VERSION_DEFAULT)
ifeq ($(strip $(VERSION)),)
VERSION := $(VERSION_DEFAULT)
endif
```

or, if an empty `VERSION` should be an error rather than a request for the default — defensible, since it means a caller passed something broken — make it one:

```make
ifeq ($(strip $(VERSION)),)
$(error VERSION is set but empty -- refusing to stamp an unversioned build)
endif
```

Either is fine; the current behaviour (stamp nothing, silently) is the one that is not. While in there, the `|| true` on the four sed lines hides genuine substitution failures and is worth dropping in the same pass.

Related: `pkg/Makefile`'s `prep-mac` guard has the same blind spot for the same empty value — see that thread.

.github/bin/test/version_lib.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
+# Version assertions shared by test_execute.bats and test_execute.sh, so the
```
pvinh-spike, line 2: From Tests agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

None of the new assertion code runs on a pull request, and there is no committed check that the assertions themselves still work. A regression in `version_lib.sh` — the one gate deciding whether a mis-stamped binary ships — would first be observed during a real release.

Every trigger in `.github/workflows/`:

```
build-and-release.yml   workflow_dispatch ONLY   <- all 3 bats invocations (548, 781, 928)
build-check.yml         pull_request + push      <- builds deb/rpm, counts files; no install, no bats
tests.yml               push + PR(dependabot)    <- make + pytest coverage
mac-artifact.yml        push + PR(master)        <- make one-dir, sudo make install, asadm -e info
                                                    never --version, never make -C pkg osx-pkg
```

`grep -rn 'bats\|test_execute' .github/workflows/` hits only inside `build-and-release.yml`. `grep -rn 'pre-commit\|shellcheck' .github/workflows/` returns nothing, so `.pre-commit-config.yaml` is wired to no workflow either.

Two consequences beyond "untested":

1. **One branch is a test that cannot fail.** The bare-version `Build`-absence check — the only assertion that catches a `git describe` fallback — fires only when `expected` has no `-` suffix. Both `get-version` paths always produce one (`5.0.3-rc3` on master, `<core>-<sha9>` elsewhere), and all three bats call sites set `EXPECTED_VERSION` to `BUILD_VERSION`. The workflow header confirms GA never produces a bare version either. I drove the branch by hand and it is correct — expected `5.0.3`, output `Version 5.0.3 / Build gabc123` returns 1 with the right message. It simply never gets driven.
2. **`test_execute.sh` is invoked by nothing.** A repo-wide grep returns three hits, all prose: its own comment, `version_lib.sh:2`, and `README.md:27,37`. It works when driven by hand (verified against shims, including the stale-build failure), but an unrun second implementation is the drift hazard this file's own header claims to prevent — and it has already drifted: asconfig's copy carries the warning "Assert first, then report: a command substitution's failure is discarded by `set -e`, so expected_version_lines must not be the thing that validates." asadm's rewrite reintroduces `$(expected_version_lines …)` inside the reporting `echo`. Safe today only because `assert_version_output` runs first — i.e. the invariant that comment protected is now undocumented, unenforced, and in a file nothing executes.

Two coordinated siblings already closed this gap, both on the string logic alone — no package, no binary, no bats needed:

```
asconfig  .github/bin/test/version_lib_test.sh   run by test.yml       (push + pull_request)
asbench   .github/bin/test/version_lib.bats      run by build-check.yml job "Naming scripts"
          + pkg_release.bats
```

asbench's `build-check.yml` header states the case exactly: "that workflow is workflow_dispatch-only, so pkg_release.sh … and version_lib.sh (which gates whether a release ships) would otherwise first run during a real release." asadm's `build-check.yml` is the same file with that paragraph and that job absent.

The port cost is near zero: I copied asconfig's `version_lib_test.sh` next to asadm's **unmodified** `version_lib.sh` and ran it — 14/14 ok, exit 0.

Note `pkg_release.sh` also gains a new call site in this PR (`BUILD_VERSION="$(.github/bin/pkg_release.sh "${BASE}" version)-${SHORT_SHA}"`), and asadm has no equivalent of asbench's `pkg_release.bats`.

## Reasoning Chain

1. Read every `on:` block rather than assuming; `build-and-release.yml` is dispatch-only, and it is the only workflow that mentions bats.
2. Traced both `get-version` branches to see which `expected` shapes are reachable, then checked which assertion branches those shapes exercise — the bare-version branch is reachable by none.
3. Grepped the whole tree for `test_execute.sh`; three hits, all prose.
4. Diffed asadm's `test_execute.sh` and `version_lib.sh` against asconfig's and asbench's to distinguish deliberate adaptation from drift.
5. Ran asconfig's self-test against asadm's library unmodified to establish the port cost empirically rather than asserting it.

## Solution

Copy `.github/bin/test/version_lib_test.sh` from asconfig, swap the version strings for asadm's and the `report()` helper for asadm's three lines ("Aerospike Administration Shell" / Version / Build), and add a step to `build-check.yml`, which already runs on `pull_request` and `push`:

```yaml
      - name: Test the version assertions
        run: bash .github/bin/test/version_lib_test.sh
```

It needs no container, no package and no bats, so it belongs in its own small job rather than riding the distro matrix. Include the two cases that make the unreachable branch reachable — asconfig's file already has them verbatim:

```sh
check 1 "a git-describe binary fails a bare version" \
    assert_version_output "$(report 5.0.3-15-gabc123def)" 5.0.3
check 0 "matching bare stamp passes" assert_version_output "$(report 5.0.3)" 5.0.3
```

For `test_execute.sh`, pick one: wire it into that same job driven against shims, or delete it and the two README references. If it stays, restore asconfig's "assert first, then report" comment so the ordering invariant survives the next edit, and say in the README which file is authoritative.

Adding asbench's `pkg_release.bats` equivalent in the same job would cover the new `BUILD_VERSION` call site.

Not checked: whether `build-check.yml`'s existing job could host this without a matrix change — I assumed a separate job is cleaner, but riding the existing one may be fewer lines.

.github/bin/os_version.sh
```diff
@@ -8,20 +8,47 @@
...
 #
-# Reads /etc/os-release. Every distro in the CI matrix and every supported
-# local-dev distro has this file; the legacy /etc/issue fallback was removed
-# along with the tools-packaging-common submodule.
+# On Linux this reads /etc/os-release. Every distro in the CI matrix and every
+# supported local-dev distro has this file; the legacy /etc/issue fallback was
+# removed along with the tools-packaging-common submodule.
+#
+# On macOS it emits "macos<major>" from the product version (26.0 -> macos26).
+# This is the single source of truth for the platform field of the .pkg file
```
pvinh-spike: From Architecture agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

The comment claims this script is "the single source of truth for the platform field of the .pkg file name". It is not — it emits `macos15`, while the platform field is `macos15-arm64`. The `-<arch>` half is hand-appended by three separate call sites:

```
pkg/Makefile:71                  MAC_PKG = ...-$(MAC_PLATFORM)-$(ARCH).pkg
build-and-release.yml:372        PKG_PLATFORM=${plat}-$(uname -m)
build-and-release.yml:602        pat="*-${plat}-$(uname -m).pkg"
```

Three copies of the same two-field rule, each pairing `os_version.sh` with its own `uname -m`. The PR's stated motivation is that "the name was composed twice … Two copies of one convention is two places for it to drift" — the `mv` rename did go away, but the two-field concatenation went from two places to three.

The bundle repo solved the same problem the other way and says so: `citrusleaf/aerospike-tools` `.github/scripts/mac_platform.sh` emits `macos${mac_version%%.*}-$(uname -m)` under the header "Single source of truth for that token", and its one caller uses it for both naming and selection. That script also asserts the two repos agree — "The per-tool repos emit the same token from their .github/bin/os_version.sh; these two must agree" — which is false as written, since the two scripts emit different tokens. The `macos<major>` sub-field does agree, so the bundle's glob still matches and nothing is broken today; the invariant is just misstated in both places.

## Reasoning Chain

1. Read what the script actually emits on the Darwin path: `echo "macos${mac_version%%.*}"` — one field.
2. Grepped this repo for every site that appends the arch: three, listed above.
3. Read the bundle's `mac_platform.sh` to see how the sibling handles it — one script emitting both fields, one caller.
4. Confirmed the mismatch is cosmetic today: the bundle globs `aerospike-asadm-*-macos<major>-<arch>.pkg`, and `make -C pkg -n osx-pkg VERSION=5.0.3-rc3` emits `aerospike-asadm-5.0.3-3-macos26-arm64.pkg`, so the primary match hits.

## Solution

Cheapest and safest: narrow the claim to what the script does — "single source of truth for the **`macos<major>`** field of the `.pkg` name; the `<arch>` half is `uname -m`, appended by the caller" — and fix the reciprocal claim in the bundle's `mac_platform.sh`.

The stronger option is to make the claim true: have the Darwin branch emit `macos${mac_version%%.*}-$(uname -m)` and drop the three appends. That matches the bundle exactly and leaves one `uname -m` in the repo. I would not do it in this PR though — the other four coordinated repos' `os_version.sh` all emit the bare `macos<major>`, so changing asadm alone makes it the odd one out, and the Linux branch's contract (distro token, no arch) would then be asymmetric with the Darwin one. If it is worth doing it is worth doing across all five, separately.

Not checked: whether anything outside these repos calls `os_version.sh` and would be affected by the stronger option.

pkg/Makefile
```diff
@@ -24,20 +24,51 @@ PKG_RELEASE_SH = $(PKG_DIR)/../.github/bin/pkg_release.sh
...
+# is a CLI installed under /usr/local, so a disk image has nothing to drag and
+# would be a second container needing its own signature, notarization and
+# staple. Homebrew consumes the .pkg directly (url + `pkg` stanza), and
+# `sudo installer -pkg ... -target /` needs no hdiutil attach.
+#
+# Every field is lower case and '-' separated so the name is machine-parsable
+# and directly interpolatable by a Homebrew cask:
+#
+#     arch arm: "arm64", intel: "x86_64"
+#     url ".../aerospike-asadm-#{version}-macos15-#{arch}.pkg"
```
pvinh-spike: From Style agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

The Homebrew example contradicts the format declared 13 lines above it, and it is the block a cask author will read first.

Declared at line 42: `<name>-<version>-<iteration>-macos<major>-<arch>.pkg`, example `aerospike-asadm-5.0.3-1-macos15-arm64.pkg`.

Rendered here: `url ".../aerospike-asadm-#{version}-macos15-#{arch}.pkg"` — the `<iteration>` field is gone. Copied into a cask with `version "5.0.3"`, that URL interpolates to `aerospike-asadm-5.0.3-macos15-arm64.pkg`, which is not a file this Makefile can produce. It only works if `#{version}` is secretly the full `RELEASE` (`5.0.3-3`), which nothing here says and which is neither the git tag nor the VERSION file value.

Second problem in the same line: `macos15` is hardcoded in a template whose stated purpose is that the macOS major varies per build. The matrix is `[macos-15-intel, macos-14, macos-15, macos-26]`, so one release publishes three arm64 pkgs (`macos14`, `macos15`, `macos26`) whose contents are interchangeable — this comment itself says the token records the build generation and that nothing pins `MACOSX_DEPLOYMENT_TARGET`. A cask therefore has to hardcode one arbitrary build major, and mapping it from the user's OS (`on_sequoia`/`on_sonoma`) would be wrong by the comment's own semantics. Whichever major is chosen, the URL 404s the day that runner label leaves the matrix.

Also, line 51's "every field is lower case and '-' separated so the name is machine-parsable" does not hold for a dev build. Verified with `make -C pkg -n osx-pkg VERSION=5.0.3-abc123def`: `RELEASE := 5.0.3-abc123def-1`, so the name is `aerospike-asadm-5.0.3-abc123def-1-macos26-arm64.pkg` — six hyphen-separated fields where the format declares five, because `PKG_VERSION` itself carries a hyphen (`pkg_release.sh` folds only `-rcN`).

## Reasoning Chain

1. Compared the declared format at line 42 against the example at line 55 field by field: the example is missing `<iteration>`.
2. Resolved the real name with `make -C pkg -n osx-pkg` for both a release and a dev version, rather than trusting either comment.
3. Read the matrix to establish how many majors one release publishes, then re-read this comment's own statement of what `macos<major>` means — the two together are what make a single-`url` cask unwritable.
4. Counted hyphen-separated fields in the dev-build name against the declared five.

## Solution

At minimum, make the example match what the code emits and stop implying a fixed major:

```make
#     arch arm: "arm64", intel: "x86_64"
#     url ".../aerospike-asadm-#{version}-macos15-#{arch}.pkg"
#   where #{version} is the full <version>-<iteration> release string, and the
#   macos<major> field must be pinned to whichever build the cask ships.
```

and soften line 51 to what is true: release names are positionally parsable; dev-build names carry a hyphenated version field and should be matched with a glob, as the bundle's selector does.

The deeper question is worth answering before a cask is actually written: shipping one pkg per build major means the cask has to name a specific runner generation. If the intent is a stable cask URL, that wants either a deployment-target-pinned `aerospike-asadm-<release>-macos-<arch>.pkg` pair with the per-major builds staying CI-internal, or a stable alias published alongside. I have not checked whether a cask exists yet or what it currently points at — that is the check that decides how much of this matters now versus at first publish.

asinfo/asinfo.py
```diff
@@ -260,7 +260,12 @@ def get_cli_args():
 
     print("Aerospike Information Tool")
     print("Version " + version)
-    print("Build " + build)
+
+    # Only when there is one, matching asadm. A bare version (a GA build) has
+    # no build field, and printing "Build " with nothing after it is both a
+    # wart and indistinguishable -- to a grep -- from a real build line.
+    if build:
```
pvinh-spike, line 276: From Architecture agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

The `if build:` fix is right, but it is applied to a hand-copy of `asadm.py`'s `get_version()`, and the copy is incomplete in a way the comment ("matching asadm") asserts it is not.

`asadm.py:1011-1018`:

```python
def get_version() -> tuple[str, str]:
    if __version__.startswith("$$"):
        return DEFAULT_ASADM_VERSION, ""
    sVersion = __version__.split("-")
    version = sVersion[0]
    build = sVersion[-1] if len(sVersion) > 1 else ""
    return version, build
```

`asinfo/asinfo.py:256-258` is lines 4-6 of that function, copied verbatim. The `startswith("$$")` fallback is **not** copied. So an unbuilt asinfo — run from source, with `__version__` still the sed placeholder — prints `Version $$__version__$$`, where asadm prints `DEFAULT_ASADM_VERSION`. The two do not match, in the one case the comment's wording most invites you to assume they do.

This is now the third encoding of the same split rule: `asadm.py`, here, and `.github/bin/test/version_lib.sh` (`${expected%%-*}` / `${expected##*-}`), plus partial copies in the bundle and four sibling repos.

Low impact — a built package stamps both files, so the divergence only shows when running from source — hence minor. But `--version` from a source checkout is a normal thing to do, and the placeholder output is confusing rather than merely wrong.

## Reasoning Chain

1. Read both implementations side by side rather than trusting "matching asadm": asinfo has lines 4-6 of a 7-line function.
2. Traced the missing branch: with `__version__ = "$$__version__$$"`, `split("-")` yields a single element, so `version` is the raw placeholder and `build` is `""` — asinfo prints `Version $$__version__$$` and no Build line.
3. Confirmed the build path stamps both files (`makefile` seds `asadm.py` and `asinfo.py`), which bounds the impact to source checkouts.
4. Counted the other encodings of the same split rule to size the duplication.

## Solution

Smallest correct fix — add the missing guard so the comment is true:

```python
if args.version:
    if __version__.startswith("$$"):
        version, build = DEFAULT_ASADM_VERSION, ""
    else:
        sVersion = __version__.split("-")
        version = sVersion[0]
        build = sVersion[-1] if len(sVersion) > 1 else ""
```

Better, since the build already copies asadm's sources next to asinfo's (`makefile`: `cp -f *.py $(BUILD_ROOT)tmp/`, `cp -rf asinfo/*`) and asinfo ships inside the asadm package: lift the parse into one helper both call.

```python
def split_version(raw: str, default: str) -> tuple[str, str]:
    if raw.startswith("$$"):
        return default, ""
    parts = raw.split("-")
    return parts[0], (parts[-1] if len(parts) > 1 else "")
```

I have not checked whether asinfo can already import from `lib/` in the frozen PyInstaller layout — that is the one thing to verify before choosing the shared-helper version over the four-line guard.

.github/bin/test/version_lib.sh
pvinh-spike: Resolved by 9768594 — asbackup's `case` form ported verbatim, and asbench's empty-trailing-field guard restored.

Verified after the change, VERSION file `5.0.3-rc3`:

```
EXPECTED_VERSION='5.0.3-rc3'        rc=0
EXPECTED_VERSION='5.0.3-rc9'        rc=1   <- was 0
EXPECTED_VERSION='5.0.3-abc123def'  rc=0   <- dev-build swap still allowed
EXPECTED_VERSION='5.0.4-rc3'        rc=1
EXPECTED_VERSION=''                 rc=1
unset (falls back to file)        rc=0
expected_version_lines '5.0.3-'      rc=1   <- was "Build " + blame the binary
```

The four `assert_version_output` cases (matching rc, stale rc2, matching bare, git-describe against a bare expectation) all keep their previous verdicts.

`05d92e0` adds a case for the rcN mismatch to the new self-test, and I confirmed it goes red when the core-only comparison is reintroduced — so this cannot regress silently again. https://github.com/aerospike/aerospike-admin/blob/0599bba/.github/bin/test/version_lib.sh

.github/workflows/build-and-release.yml
pvinh-spike: Resolved by 3d0a3f5 — `inputs.release` now passes through `env:` and the guard reads `RELEASE == "true" && master && !skip_tagging`, which is `create-tag`'s condition verbatim.

Used `== "true"` rather than `!= "false"` so an omitted input lands on the permissive side, matching `create-tag`'s `== true` — which also covers the `push:`/`workflow_call:` case raised above.

Verified over all eight release/ref/skip_tagging combinations against a repo with tag `5.0.3-rc3` present: exactly one blocks, and it is the tagging run. The `release=false` master dry run now passes against the already-tagged VERSION, and a bumped VERSION passes on the release path. https://github.com/aerospike/aerospike-admin/blob/0599bba/.github/workflows/build-and-release.yml#L136

pkg/Makefile
pvinh-spike: Resolved by b4bc48f — `prep-mac` now guards `PKG_VERSION` and `ITERATION` alongside `MAC_PLATFORM`.

```
$ make -C pkg prep-mac VERSION=
ERROR: empty version/iteration -- .github/bin/pkg_release.sh failed for VERSION=''; refusing to build an unnamed pkg
make: *** [prep-mac] Error 1
```

The existing `MAC_PLATFORM` guard still fires independently (checked by stubbing `os_version.sh` to `exit 1`), and the release name is unchanged: `aerospike-asadm-5.0.3-3-macos26-arm64.pkg`.

I did not add a whole-basename field-count assertion, as suggested — a dev build's `PKG_VERSION` legitimately carries a hyphen (`5.0.3-abc123def`, six fields where the format declares five), so a count would misfire on exactly the case the format documents. The two `test -n` lines cover the failure that was reachable.

The shared root cause in the root `makefile` is fixed in the same commit — see that thread. https://github.com/aerospike/aerospike-admin/blob/0599bba/pkg/Makefile#L172

makefile
pvinh-spike: Resolved by b4bc48f — the empty-VERSION fallback is restored:

```make
VERSION ?= $(VERSION_DEFAULT)
ifeq ($(strip $(VERSION)),)
VERSION := $(VERSION_DEFAULT)
endif
```

Took the fallback rather than `$(error …)`: it restores the exact semantics of the `$${VERSION:-…}` this replaced, so the change is a pure fix rather than a new failure mode for anyone currently passing an empty value.

Verified:

```
VERSION=           -> stamps 5.0.3-rc3   (was "")
VERSION='5.0.3-rc3'  -> stamps 5.0.3-rc3
VERSION='9.9.9'      -> stamps 9.9.9
```

I left the `|| true` on the four sed lines alone — dropping it is right, but it is unrelated to this branch's scope and would change failure behaviour for the Linux and macOS build legs at once, which deserves its own change. https://github.com/aerospike/aerospike-admin/blob/0599bba/makefile#L42

.github/bin/test/version_lib.sh
pvinh-spike: Partly resolved by 05d92e0 — the coverage gap is closed; one sub-point is not, and I left it deliberately.

**Done.** `.github/bin/test/version_lib_test.sh` is ported from asconfig, adapted to asadm's version strings and to what asadm/asinfo actually print, and run by a new `naming-scripts` job in `build-check.yml`, which already triggers on `pull_request` and `push`. It asserts against strings, so no container, no package, no bats. 21/21 pass.

It includes the two cases that make the previously-unreachable branch reachable — `a git-describe binary fails a bare version` and `matching bare stamp passes` — plus cases for both defects fixed in this round. Confirmed it genuinely fails when a bug is reintroduced:

```
$ # core-only cross-check restored
$ bash .github/bin/test/version_lib_test.sh
FAIL: a same-core different-rcN fails (wanted exit 1, got 0)
1 version_lib.sh assertion(s) behaved wrongly
exit=1
```

Action pins reuse the ones `installer-build` already carries in the same file; both verified to resolve upstream.

**Not done: `test_execute.sh`.** Still invoked by nothing. I did not wire it, because driving it needs shim binaries on `PATH` and that is a design choice about what the fixture looks like — and I did not delete it, because it is present in all five coordinated repos and removing it here alone would break the "same file everywhere" property the header claims. Restoring asconfig's "assert first, then report" comment is the one thing worth doing regardless, and I left that to you along with the decision. It is a genuine loose end, so I have not resolved this thread.

**Not done: `pkg_release.bats`.** asadm has no equivalent of asbench's, and this branch adds a new `pkg_release.sh` call site. The new job is the natural home for it. https://github.com/aerospike/aerospike-admin/blob/0599bba/.github/bin/test/version_lib_test.sh

.github/bin/os_version.sh
pvinh-spike: Resolved by 0599bba — the claim is narrowed to the field the script owns:

> This is the single source of truth for the "macos<major>" FIELD of the .pkg file name; the "<arch>" field beside it is `uname -m`, appended by the caller (pkg/Makefile's MAC_PKG, and the two workflow steps that name and select the artifact).

I did **not** take the stronger option of having the Darwin branch emit the full `macos<major>-<arch>` token. The other four coordinated repos' `os_version.sh` all emit the bare `macos<major>`, so changing asadm alone would make it the odd one out and leave its Linux branch (distro token, no arch) asymmetric with its Darwin one. If that consolidation is worth doing it is worth doing across all five, as its own change. Noted in the commit message so it is not lost.

The reciprocal claim in the bundle's `mac_platform.sh` — "The per-tool repos emit the same token from their .github/bin/os_version.sh" — was corrected in citrusleaf/aerospike-tools#443 (commit `b48b418d`) to name both halves of the composition.

Also made the new empty-`sw_vers` guard use `[[ ]]; then` like every other conditional in the file. https://github.com/aerospike/aerospike-admin/blob/0599bba/.github/bin/os_version.sh#L19

pkg/Makefile
pvinh-spike: Resolved by 0599bba — the example now states both caveats it could not express on its own:

```
#     arch arm: "arm64", intel: "x86_64"
#     url ".../aerospike-asadm-#{version}-macos15-#{arch}.pkg"
#
# with two caveats the example cannot express on its own:
#
#   * #{version} must be the full <version>-<iteration> release string
#     (5.0.3-3), not the bare product version -- the iteration is one of the
#     name's fields, and it is what distinguishes two builds of one release.
#   * macos<major> has to be pinned to whichever build the cask ships. It
#     records the runner the artifact was built on (see below), so it cannot be
#     derived from the installing machine's OS, and one release publishes a pkg
#     per (macOS major, arch) pair.
```

and the machine-parsable claim is qualified: release names are positionally parsable, a dev build's `<version>` carries its own hyphen (`5.0.3-abc123def`) so those want a glob.

I did not act on the deeper suggestion — publishing a deployment-target-pinned stable name, or an alias, so a cask need not pin a build generation. That is a real question but it is a change to what the release publishes, not a fix to this branch, and it wants the bundle and the other four repos in the room. Worth its own issue. https://github.com/aerospike/aerospike-admin/blob/0599bba/pkg/Makefile#L51

asinfo/asinfo.py
pvinh-spike: Resolved by 0599bba — asadm's `startswith("$$")` fallback is now present, so the "matching asadm" comment is true:

```
$ python3 asinfo/asinfo.py --version        # unstamped source checkout
Aerospike Information Tool
Version development
$ # stamped 5.0.3-rc3
Version 5.0.3
Build rc3
$ # stamped 5.0.3
Version 5.0.3
```

I took the four-line guard rather than the shared helper, and the check you flagged is why: `asinfo.py` imports only stdlib, and `lib/` is on its path only in the built layout (`makefile` copies `asinfo/*` and `lib` into the same `build/tmp/`). From a source checkout — the exact case this fixes — `from lib.utils.constants import …` would fail. So the literal `"development"` is inlined with a comment pointing at `DEFAULT_ASADM_VERSION` and saying why it is not imported.

That does leave the split rule encoded in two Python files. Unifying it needs the import path sorted out first, which is a bigger change than this branch. https://github.com/aerospike/aerospike-admin/blob/0599bba/asinfo/asinfo.py#L262

a-spiker (approved):

PR #541, opened 2026-08-19
fix(pkg): ship one macOS pkg per arch, with an asserted support floor

a-spiker:
Follow-up to #538, which named the macOS pkg for Homebrew but kept a `macos<major>` field in it.

## Why

`aerospike-asadm-<version>-<iteration>-macos<major>-<arch>.pkg` records the macOS generation the artifact was **built on**. That is not something a user can act on. macOS is forward compatible and never backward, so the build made on the oldest runner already serves every supported release and the per-generation builds are strictly *less* installable. One release published three arm64 artifacts and pushed the choice of which onto the user.

The floor was also never chosen. Nothing pinned `MACOSX_DEPLOYMENT_TARGET`, so `clang` stamped `minos` from whichever runner GitHub happened to schedule. Measured on the shipping 13.0.2 bundle:

| | arm64 | x86_64 |
|---|---|---|
| aql, asbackup, asrestore, asbench | 14.0 | 15.0 |
| asconfig | 12.0 | 12.0 |
| asadm, asinfo + 63 bundled deps | 11.0 | — |
| `asadm/_internal/less` | **14.8** | **15.7** |

Two different floors for the same release, both accidental, and both set by a vendored `less` rather than by anything we compile.

## What

- Name the pkg `aerospike-asadm-<version>-<iteration>-macos-<arch>.pkg`.
- Add `.github/bin/check_min_os.sh`, run before `pkgbuild`: every Mach-O in the staged payload must have `minos <= MACOS_MIN_VERSION` (14.0). It catches an unpinned compile inheriting the runner's SDK *and* prebuilt files vendored off the runner, which no compile flag reaches.
- Publish from one leg per arch (`macos-14`, `macos-15-intel`). Every leg of an arch now produces the same file name, so a second uploader would collide. The other legs still build and sanity-test — that is what the matrix is for.
- Select by arch in `test-install-mac-and-execute`. All four legs now install the same two pkgs, so the `macos-14` build is actually executed on 15 and 26. Previously each leg installed the pkg built on its own generation, so nothing ever proved a pkg runs off its build host.

## asadm-specific

`/usr/bin/less` was hardcoded in `pyinstaller-build.spec` and copied off the build runner, so it carried that runner's `minos` (14.8 above) and set the floor for the whole package — higher than any binary we compile. It is no longer bundled on macOS, where the system copy is always present; `CliView` falls back to `less` on `PATH` when the bundled copy is absent. Linux, where the bundled copy earns its place, is unchanged.

## Verified

- `pkg_release.bats` — 15/15 pass.
- `test/unit/view/test_view.py` — 59 passed, 4 subtests.
- `check_min_os.sh` validated against the real shipping 13.0.2 artifacts before wiring it in: it flags the arm64 `less` at 14.8 and the x86_64 tools at 15.0/15.7, and passes clean at floor 15.

Companion PRs follow in aql, aerospike-benchmark, aerospike-tools-backup, asconfig and aerospike-tools.

### DIFF
--- a/.github/bin/check_min_os.sh
+++ b/.github/bin/check_min_os.sh
@@ -23,14 +23,17 @@ version_le() {
 	[[ "$1" == "$2" ]] || [[ "$(printf '%s\n%s\n' "$1" "$2" | sort -V | head -1)" == "$1" ]]
 }
 
-# LC_BUILD_VERSION on current objects, LC_VERSION_MIN_MACOSX on older ones.
+# Every floor in the file, one per line: LC_BUILD_VERSION on current objects,
+# LC_VERSION_MIN_MACOSX on older ones. A universal binary carries one set of
+# load commands per slice, and `otool -l` prints them in slice order, so
+# stopping at the first would let a later slice's higher floor through.
 read_minos() {
 	otool -l "$1" 2>/dev/null | awk '
-		/LC_BUILD_VERSION/  { in_bv = 1; next }
-		/LC_VERSION_MIN_MAC/ { in_vm = 1; next }
-		in_bv && $1 == "minos"   { print $2; exit }
-		in_vm && $1 == "version" { print $2; exit }
-		/^Load command/ { in_bv = 0; in_vm = 0 }
+		/^Load command/      { bv = 0; vm = 0; next }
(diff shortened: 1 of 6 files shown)


pvinh-spike (approved):
LGTM

PR #548, opened 2026-09-02
feat: TOOLS-4213 create secondary indexes from AEL source 

a-spiker:
## Summary

Adds AEL (Aerospike Expression Language) support to `manage sindex create`, and shows the AEL source back in `show sindex` / `info sindex`.

Before, an expression secondary index meant leaving asadm: write Java against the compiled-expression builder, print `Expression.getBase64()`, paste the blob into `manage sindex create ... exp_base64 <blob>`. Server 8.1.3 compiles AEL source itself, so asadm can take the source directly.

Jira: TOOLS-4213

## What's in it

Two new modifiers on `manage sindex create`, both gated on server >= 8.1.3:

```
manage sindex create integer sum-idx ns test set donor ael '$.a:INT + $.b:INT'
manage sindex create integer sum-idx ns test set donor ael_b64 JC5hOklOVCArICQuYjpJTlQ=
```

- `ael` — plain source. Must arrive as one quoted argument: asadm's lexer splits on spaces and ends the command at a bare `;`, and the modifier parser drops duplicate words, so an unquoted expression would be silently mangled. Multi-word input is rejected up front with a message naming the fix.
- `ael_b64` — the same source base64 encoded, for scripts and for anywhere quoting is awkward.

Both are mutually exclusive with each other and with `bin`, `ctx`, `ctx_base64`, `exp_base64`.

**Wire format.** The server has no `ael` info parameter. AEL travels inside `exp=` as base64 of msgpack `[128, "<source>"]` (`EXP_AEL_COMPILE`), which `build_internal` sniffs before parsing a compiled tree (`as/src/exp/exp.c:1192-1210`). asadm does the packing in `pack_ael_expression`.

**Display.** The default `sindex-list` renders the compiled form; only `b64=true` echoes back what the client sent. On 8.1.3+ `info_sindex` issues both calls and joins them per `(ns, indexname)`, putting the decoded source in a new `ael` field and clearing `exp`, so an index shows either an expression or AEL, never both. Decoding never raises: anything that is not a well-formed envelope falls back to the server's rendering, and a failed `b64=true` call leaves the default response untouched.

```
Index Name |Set  |Bin |Bin Type|State|Expression                          |AEL
ael-idx    |donor|null|integer |RW   |--                                  |$.a:INT + $.b:INT
cmp-idx    |donor|null|integer |RW   |add(bin_int("a"), bin_int("b"))     |--
```

Both columns hide themselves when nothing populates them, so clusters without expression indexes see no change.

This second call is interim, and goes away with SERVER-1401 / TOOLS-4227.

## Deviations from the ticket

- The ticket asks for the AEL source in the **Expression** column. This puts it in its own **AEL** column instead: AEL is not a rendering of a compiled expression, and mixing the two in one column leaves no way to tell which language a cell holds. Ticket text should be updated to match.
- The ticket only asks for `ael_b64`. `ael` was added alongside it so operators do not have to base64 by hand.
- "Invalid AEL surfaces the server's positioned diagnostic" **cannot pass today**. The server builds `"<msg> at line N col M"` and then discards it: `parse_exp` calls `as_exp_build_err_reset()` (`as/src/sindex/sindex.c:1662`) and `as_sindex_validate_exp` replies with a flat `bad 'exp'`. Confirmed against 8.1.3.0 — the useful text only reaches the server log:

  ```
  WARNING (exp): (exp_ael.c:219) build_internal_ael - unresolved bin type,
                 use $.bin:STRING, $.bin:INT, or $.bin:FLOAT @ 6
  ```

  asadm surfaces whatever the server sends, unmodified, so the better message flows through once the server keeps it. Needs a SERVER ticket.

## Testing

- Unit: 2087 pass. New coverage for the envelope encode/decode and its fallbacks, the `b64=true` join, the version gate, base64 and empty-source validation, unquoted `ael`, and every new conflict pair.
- e2e: `ManageSindexTest` passes against an 8.1.3 cluster, including four new AEL tests (quoted `ael`, `ael_b64`, unquoted rejection, invalid AEL).
- Manual, against a live 8.1.3.0 server: create from `ael` and from `ael_b64`, source round-trips through `show sindex` and `info sindex`, `exp_base64` indexes still render the compiled form unchanged, and pre-8.1.3 clusters error before any info call.

## Notes for reviewers

Two server-side behaviours worth knowing, neither introduced here:

- An AEL index and an equivalent compiled-expression index are not deduped — `as_si_by_defn` compares raw wire bytes, and the envelope differs from the compiled tree. Both get created and both populate.
- Bin types are resolved statically by the AEL compiler; `:T` is only needed where inference cannot pin a type. `$.a + 1` compiles, `$.a + $.b` does not.


lib/live_cluster/manage_controller.py
```diff
@@ -1532,7 +1643,7 @@ async def _do_create_set(self, line):
 
         unsupported = [
             m
-            for m in ("bin", "in", "ctx", "exp_base64", "ctx_base64")
+            for m in ("bin", "in", "ctx", "exp_base64", "ctx_base64", "ael", "ael_b64")
```
colton-aerospike, line 1661: Severity: Important

 When ael or ael_b64 is supplied without a value, parse_modifiers() represents it as an empty list. Since this check uses truthiness (self.mods.get(m)), the modifier is not added to unsupported.
As a result, commands such as:
`text
 manage sindex create idx ns test set demo ael
 `
silently create a plain set index, even though the user requested an AEL modifier. The same issue occurs for a bare ael on the typed bin-index path, where it can create a normal bin index.

Customers should receive a clear validation error for an incomplete command rather than have the tool silently perform a different operation. Please track modifier presence separately from its value and reject valueless ael/ael_b64.

test/e2e/live_cluster/test_manage.py
```diff
@@ -883,6 +884,141 @@ def test_fails_to_create_sindex_in_invalid(self):
 
         self.assertStdErrEqual(exp_stderr, actual.stderr)
 
+    def test_can_create_sindex_from_ael(self):
+        ael_src = "$.a:INT + $.b:INT"
+        ael_b64 = base64.b64encode(bytes(ael_src, "utf-8")).decode("utf-8")
+
+        actual = test_util.run_asadm(
+            self.get_args(
+                "manage sindex create integer {} ns {} set {} ael_b64 {}".format(
```
colton-aerospike, line 893: Severity: Important

The newly added AEL E2E tests run unconditionally, but AEL sindex support is explicitly gated to server version 8.1.3 and later in the implementation. This test suite supports older server versions, so these tests will fail on pre-8.1.3 servers instead of being skipped.

Please add the existing-style server-version guard around all AEL-dependent E2E tests, including creation, invalid-AEL, rendering, and set-index rejection cases. This preserves coverage on supported servers without breaking the suite on older ones.

colton-aerospike (changes requested):

lib/live_cluster/manage_controller.py
a-spiker: Good catch! Done.

test/e2e/live_cluster/test_manage.py
a-spiker: The e2e suite runs against a pinned server image (SERVER_TAG in test/e2e/lib.py, currently 8.1.3.0), so these tests always execute against a version that supports AEL. Adding skip guards would let a tag bump silently drop AEL coverage rather than fail loudly. Leaving them unconditional, matching the other sindex tests in this class.

colton-aerospike (approved):
LGTM

PR #549, opened 2026-09-03
fix(docker): restore the C.UTF-8 locale in the image

a-spiker:
## Problem

Every image up to 13.0.2 ran through the `/aerospike/wrapper` entrypoint, whose first act was `export LC_CTYPE=C.UTF-8`. The 13.0.3 rebuild dropped the wrapper and did not replace the locale setting, so 13.0.3 containers start under the POSIX locale:

```
$ docker run --rm --entrypoint sh <13.0.3 image> -c 'locale charmap'
ANSI_X3.4-1968
```

Per the wrapper's own comment that is the setting that let `asadm` render the µ in micro-benchmark output.

## Fix

```dockerfile
ENV LANG=C.UTF-8 \
    LC_CTYPE=C.UTF-8
```

- **`ENV`, not a new entrypoint script.** It applies however the container is invoked: `docker exec`, `--entrypoint` overrides and Kubernetes pod specs, none of which the wrapper covered. It also survives whatever TOOLS-4360 decides about `ENTRYPOINT`.
- **Placed with the base setup, before the install step**, so the build itself runs UTF-8. Setting `LANG` there follows `python`/`postgres`; pinning `LC_CTYPE` alongside it is a deliberate deviation, since those images ship or generate real locales and this one has only `C.utf8`.
- **Both vars, not just `LANG`.** C.UTF-8 is the only locale in the image, and its non-ctype categories are identical to C, so `LANG` alone changes nothing but the charset. But with `LANG` alone a user passing `-e LANG=en_US.UTF-8` silently falls back to ASCII, because that locale is not generated here. Measured on `debian:trixie-slim`:

  | image `ENV` | default `locale charmap` | with `-e LANG=en_US.UTF-8` |
  | --- | --- | --- |
  | none (13.0.3 today) | `ANSI_X3.4-1968` | `ANSI_X3.4-1968` |
  | `LANG` only | `UTF-8` | `ANSI_X3.4-1968` |
  | `LANG` + `LC_CTYPE` | `UTF-8` | `UTF-8` |

  Pinning `LC_CTYPE` holds UTF-8 through the override, which is the regression this ticket is about.
- `C.UTF-8` is built into glibc on trixie, so no `locales` package and no image-size cost.

## Test

The docker smoke test asserts the built image resolves a UTF-8 charmap, per arch:

```bash
CHARMAP=$(docker run --rm --entrypoint sh "${IMAGE}" -c 'locale charmap')
[[ "${CHARMAP}" == "UTF-8" ]] || { ... exit 1; }
```

`locale charmap` is the assertion rather than a grep of the env var: it fails on an unset locale *and* on a locale name glibc cannot resolve, so it proves the charset the tools actually get. It runs in the existing per-arch matrix on native runners, covering `linux/amd64` and `linux/arm64`.

Verified locally against the pinned base digest (table above), `hadolint` clean, `actionlint` clean.

Companion PRs carry the identical change in the other five repos; the other wrapper-removal gaps (`ENTRYPOINT`/`CMD`, missing `ca-certificates`, PID 1 signal handling, tag policy, build attestations) are tracked separately.

TOOLS-4361

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -943,6 +943,12 @@ jobs:
           docker run --rm --entrypoint sh "${IMAGE}" -c 'test -f /etc/aerospike/astools.conf' \
             || { echo "ERROR: /etc/aerospike/astools.conf missing in image" >&2; exit 1; }
 
+          # The pre-13.0.3 wrapper exported LC_CTYPE=C.UTF-8; the Dockerfile now
+          # sets it as ENV. Assert glibc resolves a UTF-8 charmap in the image.
+          CHARMAP=$(docker run --rm --entrypoint sh "${IMAGE}" -c 'locale charmap')
+          [[ "${CHARMAP}" == "UTF-8" ]] \
+            || { echo "ERROR: image locale charmap is ${CHARMAP}, expected UTF-8" >&2; exit 1; }
+
           if [[ -n "${SNYK_TOKEN:-}" ]]; then
             snyk container test "${IMAGE}" \
               --platform=linux/${{ matrix.arch }} \
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -945,9 +945,18 @@ jobs:
 
           # The pre-13.0.3 wrapper exported LC_CTYPE=C.UTF-8; the Dockerfile now
           # sets it as ENV. Assert glibc resolves a UTF-8 charmap in the image.
-          CHARMAP=$(docker run --rm --entrypoint sh "${IMAGE}" -c 'locale charmap')
+          CHARMAP=$(docker run --rm --entrypoint sh "${IMAGE}" -c 'locale charmap') \
+            || { echo "ERROR: could not read the locale charmap from ${IMAGE}" >&2; exit 1; }
           [[ "${CHARMAP}" == "UTF-8" ]] \
-            || { echo "ERROR: image locale charmap is ${CHARMAP}, expected UTF-8" >&2; exit 1; }
+            || { echo "ERROR: image locale charmap is '${CHARMAP//$'\n'/ }', expected UTF-8" >&2; exit 1; }
+
+          # LC_CTYPE is pinned so a LANG override to an ungenerated locale cannot
+          # drop the ctype back to ASCII. Without this the pin could be trimmed to
+          # LANG alone and the assertion above would stay green.
+          CHARMAP=$(docker run --rm -e LANG=en_US.UTF-8 --entrypoint sh "${IMAGE}" -c 'locale charmap') \
+            || { echo "ERROR: could not read the locale charmap from ${IMAGE} under a LANG override" >&2; exit 1; }
+          [[ "${CHARMAP}" == "UTF-8" ]] \
+            || { echo "ERROR: LANG override drops the charmap to '${CHARMAP//$'\n'/ }', expected UTF-8 (LC_CTYPE pin lost?)" >&2; exit 1; }
 
           if [[ -n "${SNYK_TOKEN:-}" ]]; then
             snyk container test "${IMAGE}" \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -40,6 +40,15 @@ LABEL org.opencontainers.image.title="Aerospike asadm" \
       org.opencontainers.image.licenses="Apache-2.0" \
       org.opencontainers.image.vendor="Aerospike"
 
+# Restores the UTF-8 ctype the pre-13.0.3 entrypoint wrapper exported, so asadm
+# renders µ in micro-benchmark output. ENV also covers docker exec and
+# --entrypoint overrides, which the wrapper never did. C.UTF-8 is built into
+# glibc, so no locales package is needed; LC_CTYPE is pinned alongside LANG
+# because it is the only locale the image has, so a LANG override to an
+# ungenerated one would otherwise drop back to ASCII.
+ENV LANG=C.UTF-8 \
+    LC_CTYPE=C.UTF-8
+
 SHELL ["/bin/bash", "-Eeuo", "pipefail", "-c"]
 
 # Fetch (or copy) → verify SHA → apt-install the .deb. apt resolves any
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -40,12 +40,14 @@ LABEL org.opencontainers.image.title="Aerospike asadm" \
       org.opencontainers.image.licenses="Apache-2.0" \
       org.opencontainers.image.vendor="Aerospike"
 
-# Restores the UTF-8 ctype the pre-13.0.3 entrypoint wrapper exported, so asadm
-# renders µ in micro-benchmark output. ENV also covers docker exec and
-# --entrypoint overrides, which the wrapper never did. C.UTF-8 is built into
-# glibc, so no locales package is needed; LC_CTYPE is pinned alongside LANG
-# because it is the only locale the image has, so a LANG override to an
-# ungenerated one would otherwise drop back to ASCII.
+# glibc otherwise resolves the POSIX locale, whose ANSI_X3.4-1968 charmap strips
+# non-ASCII from tool output. ENV rather than an entrypoint export, so docker exec,
+# --entrypoint overrides and K8s pod specs inherit it; the pre-13.0.3 wrapper only
+# covered its own process. C.utf8 is glibc-builtin on trixie: no locales package,
+# no size cost. LC_CTYPE is pinned alongside LANG deliberately -- C.utf8 is the
+# image's only UTF-8 locale, so -e LANG=<ungenerated> would fall back to ASCII.
+# Precedence is LC_ALL > LC_CTYPE > LANG, so override the ctype via LC_ALL or
+# LC_CTYPE. Deviates from python/postgres, which set LANG only and ship locales.
 ENV LANG=C.UTF-8 \
     LC_CTYPE=C.UTF-8
 


a-spiker:
Coordinated change, one PR per image. Merge together:

- citrusleaf/aerospike-tools#450 (bundle)
- aerospike/aerospike-admin#549
- aerospike/aql#98
- aerospike/aerospike-tools-backup#175
- aerospike/aerospike-benchmark#167
- aerospike/asconfig#140

pvinh-spike:

**Promises:**
1. The image resolves a UTF-8 charmap by default, restoring the pre-13.0.3 wrapper's `LC_CTYPE=C.UTF-8`.
2. Non-ASCII output is not mangled under the POSIX locale.
3. The locale applies however the container is invoked (`docker run`, `docker exec`, `--entrypoint` overrides, K8s pod specs).
4. `-e LANG=en_US.UTF-8` still yields UTF-8 rather than silently falling back to ASCII.
5. No image-size or dependency cost.
6. The docker smoke test fails the build when the charmap is not UTF-8, on both arches.

| Gate | Status |
|------|--------|
| Contract | pass |
| Impacted surface | fail |
| Failure paths | fail |
| Evidence | **fail** |
| Quality | fail |

**3 findings** posted inline (0 Critical, 1 Important, 2 Minor).

Reviewed as one of six coordinated PRs; citrusleaf/aerospike-tools#450 is the canonical one. I diffed all six and the `ENV` is byte-identical and identically placed (right after `LABEL`, before `SHELL`) in every repo, and the `ERROR:` vs `::error::` prefix correctly follows each repo's dominant local convention. No per-repo divergence defects.

The change is right and I verified it rather than taking it on trust — reproduced the PR's measured table exactly against the pinned base digest on both arches, confirmed `C.utf8` is glibc-builtin, that `locale charmap` prints exactly `UTF-8` so the comparison is correct, and that `C.UTF-8` collates byte-identically to `C` so the `ENV` placement before the install `RUN` perturbs nothing.

The Evidence gate is what fails: `build-and-release.yml` is `workflow_dispatch`-only and is the only docker-touching workflow, so no check on this PR built the image or ran the new assertion. Promise 6 rests on a local claim. Please run a `release=false` dispatch on this branch and link the run.

The Dockerfile comment here says "so asadm renders µ in micro-benchmark output". Two reviewers independently froze a PyInstaller CPython and ran it with `LANG`/`LC_ALL`/`LC_CTYPE` all unset — the 13.0.3 state — and it printed µ correctly (`sys.flags.utf8_mode 1`, `stdout.encoding utf-8`), because PEP 540 auto-enables UTF-8 mode on a C/POSIX `LC_CTYPE`. So that stated mechanism is not the real one. The fix is still correct; the reasoning is what invites a later maintainer to delete the `ENV`. See citrusleaf/aerospike-tools#450 for the full trace. The other four companion PRs use accurate generic wording ("so non-ASCII output is not mangled under the POSIX locale") — consider adopting it here.

docker/Dockerfile
```diff
@@ -40,6 +40,15 @@ LABEL org.opencontainers.image.title="Aerospike asadm" \
...
       org.opencontainers.image.vendor="Aerospike"
 
+# Restores the UTF-8 ctype the pre-13.0.3 entrypoint wrapper exported, so asadm
+# renders µ in micro-benchmark output. ENV also covers docker exec and
+# --entrypoint overrides, which the wrapper never did. C.UTF-8 is built into
+# glibc, so no locales package is needed; LC_CTYPE is pinned alongside LANG
+# because it is the only locale the image has, so a LANG override to an
+# ungenerated one would otherwise drop back to ASCII.
+ENV LANG=C.UTF-8 \
+    LC_CTYPE=C.UTF-8
```
pvinh-spike, line 52: From Tests agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

The override case is the entire justification for this second `ENV` line and for four of the six comment lines above it — and it is the one case the smoke test does not assert.

The assertion at the charmap assertion runs a bare `docker run` with no `-e`, so it exercises only the image default. `LANG=C.UTF-8` alone satisfies it. The test therefore cannot distinguish the two-variable version from a one-variable version, and if someone later trims this to just `LANG` — the shape every official Debian-derived image uses — the smoke test stays green while promise 4 regresses silently.

The behaviour is real and worth pinning. Verified on the pinned base digest:

- `-e LANG=en_US.UTF-8` alone → `ANSI_X3.4-1968` (`locale -a` lists only `C`, `C.utf8`, `POSIX`)
- `-e LANG=en_US.UTF-8` with `LC_CTYPE=C.UTF-8` → `UTF-8`

Compounding it: nothing in CI runs this assertion at all. `build-and-release.yml` is `workflow_dispatch`-only and is the only workflow that mentions docker, so no check on this PR built the image. Promise 6 rests on the PR body's local verification.

## Reasoning Chain

1. Read the assertion at 900-902 — one `docker run`, no `-e`, so it measures the default only.
2. `LANG=C.UTF-8` alone makes that pass, so the assertion is blind to the presence or absence of the `LC_CTYPE` line directly beneath it.
3. Read the Dockerfile comment: the sole stated reason for pinning `LC_CTYPE` is the ungenerated-`LANG` fallback. That is the claim under test, and it is untested.
4. Reproduced the fallback and the fix on `debian:trixie-slim` at the pinned digest, both arches.
5. Enumerated every workflow's `on:` block. Only `build-and-release.yml` matches `docker`, and it is dispatch-only — so an untested deliberate deviation with a six-line rationale is exactly the thing that gets "simplified" away later, with no gate to catch it.

## Solution

Add the override case beside the existing check, in the same step:

```bash
CHARMAP=$(docker run --rm -e LANG=en_US.UTF-8 --entrypoint sh "${IMAGE}" -c 'locale charmap')
[[ "${CHARMAP}" == "UTF-8" ]] \
  || { echo "::error::LANG override drops charmap to ${CHARMAP}, expected UTF-8 (LC_CTYPE pin lost?)" >&2; exit 1; }
```

One extra container start in a step that already runs `apt-get install bats`, a `docker run` per bats assertion, and a Snyk scan — immaterial. It passes against the Dockerfile as written; I verified the expected value.

Separately, and this is the part that blocks the Evidence gate rather than the code: run a `release=false` dispatch of `build-and-release.yml` on this branch and link the run here, with the `docker-build-arch` log for both arches. The assertion is per-arch and only a real dispatch exercises the arm64 runner. A `pull_request` trigger is not a realistic alternative — the Dockerfile needs either OIDC-gated JFrog URLs or a signed `.deb` from `organize-signed-artifacts`.

.github/workflows/build-and-release.yml
```diff
@@ -943,6 +943,12 @@ jobs:
           docker run --rm --entrypoint sh "${IMAGE}" -c 'test -f /etc/aerospike/astools.conf' \
             || { echo "ERROR: /etc/aerospike/astools.conf missing in image" >&2; exit 1; }
 
+          # The pre-13.0.3 wrapper exported LC_CTYPE=C.UTF-8; the Dockerfile now
+          # sets it as ENV. Assert glibc resolves a UTF-8 charmap in the image.
+          CHARMAP=$(docker run --rm --entrypoint sh "${IMAGE}" -c 'locale charmap')
```
pvinh-spike: From Logic agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

Two robustness problems in these three lines, both departing from the idiom used by the check directly above them.

**The `::error::` is unreachable on the failure path that matters.** Line 900 is a simple assignment whose exit status *is* the command substitution's. Under the step's `set -euo pipefail` (line 864), a non-zero `docker run` kills the step at line 900 — lines 901-902 are dead code on that path. I reproduced this with the diff's own three lines against a non-existent image: exit 125, no `::error::` emitted, execution never reached the comparison.

So if the image is not loaded, the daemon errors, or `locale` is missing from the image (`sh -c` → 127), the job fails with raw docker stderr and no GitHub annotation. Compare lines 895-896, which hang `|| { echo "::error::..."; exit 1; }` off the `docker run` itself and therefore annotate on every failure.

**The value is interpolated unquoted.** `locale` writes its "Cannot set LC_* to default locale" diagnostics to stderr, which `$()` discards, so an exit-0-but-empty-stdout case renders as `image locale charmap is , expected UTF-8` — a degenerate message for exactly the case a reader most needs to tell apart from "wrong charmap". Quoting also contains a newline-bearing value, which would otherwise split the `echo` into a second line that the runner parses as a fresh workflow command. That last part is hardening, not a live vulnerability — the value comes from glibc inside an image built from a digest-pinned base and a signed `.deb`, and the bats shims above already pipe image-controlled output into the log — but it costs one character.

## Reasoning Chain

1. Read the step at 859-908: `set -euo pipefail` at 864, workflow-level `defaults.run.shell: bash`.
2. Bash semantics: for a command consisting only of an assignment with a command substitution, the exit status is the substitution's, so `set -e` acts on it. Nothing forms an `||` list with line 900.
3. Ran the exact three lines against a bogus image and confirmed exit 125 with no annotation and no reached comparison.
4. Checked whether `locale` itself could fail misleadingly: on this base it exits 0 in every unresolvable-locale case I could construct, writing warnings to stderr and a clean value to stdout. So the risk is the unguarded `docker run`, not `locale`.
5. Compared with lines 895-896 — the only other image assertion in the step — and found the new code is the sole departure from the file's own pattern.

## Solution

```bash
CHARMAP=$(docker run --rm --entrypoint sh "${IMAGE}" -c 'locale charmap') \
  || { echo "::error::could not read the locale charmap from ${IMAGE}" >&2; exit 1; }
[[ "${CHARMAP}" == "UTF-8" ]] \
  || { echo "::error::image locale charmap is '${CHARMAP//$'\n'/ }', expected UTF-8" >&2; exit 1; }
```

Appending `|| { ...; }` makes the assignment part of a compound command, so `set -e` no longer pre-empts it — verified. The success path is unchanged and the wrong-charmap path still exits 1 with the same text plus quotes.

docker/Dockerfile
```diff
@@ -40,6 +40,15 @@ LABEL org.opencontainers.image.title="Aerospike asadm" \
       org.opencontainers.image.licenses="Apache-2.0" \
       org.opencontainers.image.vendor="Aerospike"
 
+# Restores the UTF-8 ctype the pre-13.0.3 entrypoint wrapper exported, so asadm
+# renders µ in micro-benchmark output. ENV also covers docker exec and
+# --entrypoint overrides, which the wrapper never did. C.UTF-8 is built into
+# glibc, so no locales package is needed; LC_CTYPE is pinned alongside LANG
+# because it is the only locale the image has, so a LANG override to an
+# ungenerated one would otherwise drop back to ASCII.
+ENV LANG=C.UTF-8 \
```
pvinh-spike, line 51: From Architecture agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

`ENV` in a published image is public API, and pinning `LC_CTYPE` has two consequences the PR does not record.

**It creates a split locale downstream, with an undocumented escape hatch.** Unlike the wrapper's `export` — which lived inside the entrypoint process — image config `Env` is inherited by `docker run`, `docker exec`, K8s pod specs, *and* any image built `FROM aerospike-tools`. Since glibc precedence is `LC_ALL` > `LC_CTYPE` > `LANG`, a derived image that installs `locales`, generates `de_DE.UTF-8` and sets `LANG=de_DE.UTF-8` gets:

```
LC_CTYPE=C.UTF-8          <- pinned, wins
LC_NUMERIC="de_DE.UTF-8"  <- follows LANG
LC_TIME="de_DE.UTF-8"     <- follows LANG
```

verified on `debian:trixie-slim` + `locale-gen`. `LANG` — the variable users reach for — silently does not move ctype, with no diagnostic. `LC_ALL` does override cleanly, and so does an explicitly empty `-e LC_CTYPE=`; both verified. Neither is mentioned in the comment or in `docs/bundle-release-pipeline.md:193`, which is the only doc in the repo showing `docker run` invocations and therefore where a user would look.

In practice the damage is small — `C.UTF-8` classifies full Unicode and I found no case-mapping difference against `de_DE.UTF-8` — which is why this is minor rather than important. The case that bites is a consumer who wants a non-UTF-8 ctype and can no longer get one via `LANG`. That is a defensible deliberate choice for a tools image; it just needs to be written down, because you are overriding a user's expressed intent.

**The conformance claim in the PR body is false for the half that is contentious.** The body says this is "where the official Debian-derived images put `ENV LANG`". Inspected `.Config.Env`:

```
python:3.12-slim    LANG=C.UTF-8      (LANG only)
python:3-slim       (neither)
postgres:17         LANG=en_US.utf8   (LANG only)
debian:trixie-slim  (neither)
```

None pins `LC_CTYPE`. The `LANG` half is conventional; the `LC_CTYPE` half is a deliberate deviation — defensible on this repo's facts, since those images ship or generate real locales and this one does not, but it should be named as a deviation. Otherwise the next person "aligns with upstream" by deleting the line, which is precisely the regression promise 4 exists to prevent.

Backward compat is clean, for the record: I diffed POSIX vs `C.UTF-8` on `sort`, `grep '[a-z]'` ranges, `awk printf %.2f` and `date` — byte-identical. Nothing that parsed 13.0.3 tool output through the shell changes.

## Reasoning Chain

1. Enumerated the four consumer classes an image `Env` reaches; the `FROM` case appears nowhere in the PR.
2. Built the derived-image scenario for real (`locales` + `locale-gen de_DE.UTF-8`, then `LANG=de_DE.UTF-8`) and observed the split locale.
3. Verified both escape hatches (`LC_ALL`, and an explicitly empty `LC_CTYPE`) work, making this a documentation gap rather than a dead end.
4. Bounded the severity by measuring the actual ctype delta between `C.UTF-8` and `de_DE.UTF-8` — none for case mapping — so the residual risk is confined to consumers wanting a non-UTF-8 ctype.
5. Tested the conformance claim rather than accepting it, via `docker inspect` on three official Debian-derived images. All set `LANG` only or nothing.

## Solution

Extend `docs/bundle-release-pipeline.md:193`:

> The image sets both `LANG` and `LC_CTYPE` to `C.UTF-8`. `C.utf8` is the only UTF-8 locale glibc ships on trixie, so `LC_CTYPE` is pinned deliberately: `-e LANG=<other>` will not move the character type (precedence is `LC_ALL` > `LC_CTYPE` > `LANG`). To run under a different locale — including in an image built `FROM` this one that generates its own — set `LC_ALL`, or override `LC_CTYPE` explicitly. This deviates from `python`/`postgres`, which set `LANG` only; those images ship generated locales and this one does not.

and add a clause to the Dockerfile comment marking the `LC_CTYPE` line as an intentional deviation with that reason. Correct the PR body's conformance sentence too, since it will be read as justification.

Keep the pin — dropping it to match upstream would trade this documentation gap for a promise-4 violation. The alternative, installing `locales` so `LANG` behaves conventionally, contradicts promise 5 and is not worth it for this image.

docker/Dockerfile
a-spiker: Fixed. Added the override case in the same step, right below the existing check:

```bash
CHARMAP=$(docker run --rm -e LANG=en_US.UTF-8 --entrypoint sh "${IMAGE}" -c 'locale charmap') \
  || { echo "ERROR: could not read the locale charmap from ${IMAGE} under a LANG override" >&2; exit 1; }
[[ "${CHARMAP}" == "UTF-8" ]] \
  || { echo "ERROR: LANG override drops the charmap to '${CHARMAP//$'\n'/ }', expected UTF-8 (LC_CTYPE pin lost?)" >&2; exit 1; }
```

Confirmed it discriminates. Built two images off `debian:trixie-slim`, one with the Dockerfile's exact `ENV LANG` + `LC_CTYPE`, one with `LANG` only, and measured on both arches:

| image ENV | default | `-e LANG=en_US.UTF-8` |
| --- | --- | --- |
| `LANG` + `LC_CTYPE` | UTF-8 | UTF-8 |
| `LANG` only | UTF-8 | `ANSI_X3.4-1968` |

Identical on `linux/amd64` and `linux/arm64`. The old assertion passes in all four cells; the new one fails exactly the trim-to-`LANG` case, which is the regression you described.

On the Evidence gate, see my top-level comment: a `release=false` dispatch cannot reach `docker-build-arch` on any branch right now, and the reason is unrelated to this PR.

.github/workflows/build-and-release.yml
a-spiker: Fixed, as suggested. The annotation now hangs off the assignment, and the value is quoted with newlines folded:

```bash
CHARMAP=$(docker run --rm --entrypoint sh "${IMAGE}" -c 'locale charmap') \
  || { echo "ERROR: could not read the locale charmap from ${IMAGE}" >&2; exit 1; }
[[ "${CHARMAP}" == "UTF-8" ]] \
  || { echo "ERROR: image locale charmap is '${CHARMAP//$'\n'/ }', expected UTF-8" >&2; exit 1; }
```

Verified your diagnosis and the fix. Ran the original three lines against a nonexistent image under `set -euo pipefail`: the step died at the assignment with raw docker stderr and no annotation, exactly as you said. With the `|| { ...; }` appended, the same input reaches the annotation and exits 1. Also checked `${CHARMAP//$'\n'/ }` renders inside the double-quoted `echo` (it does, and an empty value now shows as `''` rather than vanishing).

The same guard is applied to the new override assertion.

docker/Dockerfile
a-spiker: Fixed. The comment block now records the precedence, the escape hatch and the deviation rather than just the immediate reason:

```dockerfile
# glibc otherwise resolves the POSIX locale, whose ANSI_X3.4-1968 charmap strips
# non-ASCII from tool output. ENV rather than an entrypoint export, so docker exec,
# --entrypoint overrides and K8s pod specs inherit it; the pre-13.0.3 wrapper only
# covered its own process. C.utf8 is glibc-builtin on trixie: no locales package,
# no size cost. LC_CTYPE is pinned alongside LANG deliberately -- C.utf8 is the
# image's only UTF-8 locale, so -e LANG=<ungenerated> would fall back to ASCII.
# Precedence is LC_ALL > LC_CTYPE > LANG, so override the ctype via LC_ALL or
# LC_CTYPE. Deviates from python/postgres, which set LANG only and ship locales.
```

Your point about the `FROM` case is the one that changed my mind on wording: the split-locale behaviour is a deliberate trade, so it needs to read as deliberate.

The PR body's conformance sentence is corrected too, since you were right that it would be read as justification. It now says setting `LANG` there follows `python`/`postgres` while pinning `LC_CTYPE` is the deviation, with the reason.

Keeping the pin, and not installing `locales`, per your recommendation.

a-spiker:
### On the Evidence gate

You asked for a `release=false` dispatch on this branch. **That cannot reach `docker-build-arch` today**, and the reason is unrelated to this PR.

Debian 11 went LTS end-of-life on 2026-08-31, and `bullseye-security` expired on 2026-09-07. Every `debian11` job in this workflow now dies at *Install prereqs*, before checkout:

```
E: Release file for http://deb.debian.org/debian-security/dists/bullseye-security/InRelease is expired
/__w/_temp/....sh: line 12: git: command not found
##[error]Process completed with exit code 127.
```

The chain to the image is:

```
build-artifacts -> organize-built-artifacts -> notarize-mac-artifacts
  -> sign-artifacts -> test-install-linux-and-execute
  -> organize-signed-artifacts -> docker-build-arch
```

`debian11` sits in **both** `build-artifacts` and `test-install-linux-and-execute`, so a dispatch fails long before the image is built. The gate is unreachable, not skipped.

TOOLS-3137 removes `debian11`: aerospike/aerospike-admin#555. Once that merges and this branch rebases onto it, a `release=false` dispatch becomes possible and I will link the run with the `docker-build-arch` log for both arches.

### Local evidence in the meantime

Not the product image, since that needs a signed `.deb`. This is `debian:trixie-slim`, the image's own base, with the Dockerfile's exact `ENV`:

| image ENV | default | `-e LANG=en_US.UTF-8` |
| --- | --- | --- |
| `LANG` + `LC_CTYPE` (as shipped) | UTF-8 | UTF-8 |
| `LANG` only | UTF-8 | `ANSI_X3.4-1968` |

Identical on `linux/amd64` and `linux/arm64`, matching your reproduction. Running the step's assertions as written: the pinned image passes both, the `LANG`-only image fails the new override assertion, and a nonexistent image now reaches the annotation instead of dying on raw docker stderr.

That covers promise 4 and the discriminating power of promise 6's assertion. What it does not cover is the assertion running in the real job on the real image, which is exactly what the dispatch is for.

pvinh-spike (approved):
LGTM

PR #522, opened 2026-07-24
feat: TOOLS-4134 info memory Capacity vs Allocated view (cgroup-aware, RAM-only) + per-backing index alloc in info namespace usage

a-spiker:
## What changed

`info memory` answers one question: how much of a node's RAM is spoken for. It is a memory command, so only RAM-backed allocation is counted, `index-type`/`sindex-type` pmem and flash arenas are device-backed and belong to `info namespace usage`, not here.

### `info memory` (default) - one table

`Memory Information`: `Node`, `Build`, `Capacity`, `Allocated` (`Total` / `Shmem` / `Heap` / `Heap%`), `Alloc%`, `Free%`, `Stop%`.

- `Capacity` is the cgroup limit, and only when `cgroup-mem-tracking` is enabled. There is **no host-total fallback**: the server publishes no total-memory stat (SERVER-1546 tracks exposing one) and asadm does not estimate one from the free stats, because `system_free_mem_pct` is a whole number that nets out `inactive_file` and clamps by host availability (SERVER-742), so a division by it is not a figure to divide `Alloc%` by. A node without a tracked limit renders a blank `Capacity` and `Alloc%` and is named in a warning instead.
- `Allocated Total` = shmem index + sindex arenas (+ in-memory data on Enterprise and Federal, where it is reserved up front) + process heap.
- `Alloc%` is a reservation ratio against `Capacity` and carries **no** alerts: shmem arenas are reserved and demand-faulted, so a high value is not itself an OOM signal.
- `Free%` is `system_free_mem_pct` straight from the server and carries the alerts. The tier is the node's own stop-writes floor, not a fixed number: red below `100 - Stop%`, yellow within 10 points above it, where `Stop%` is the lowest `stop-writes-sys-memory-pct` across the node's namespaces (server default 90 when namespace stats are missing).

### `info memory --verbose` (`-v`) - three more tables

- **Host and CGroup Memory**: `Free` (`System` / `Sys%` / `Host` / `Host%`), `Stop%`, `CGroup` (`Tracking` / `Used` / `Limit` / `Used%`), `THP`. The `Host` pair is dropped only when it repeats the `System` pair in both bytes and percent, which is what an untracked cgroup reports; under tracking the two are measured against different capacities and both are kept.
- **Index and Data Memory**: paired `Alloc` + `Used` for `Total`, `Primary Index`, `Secondary Index`, `Set Index` and `Data`.
[...]

PR #550, opened 2026-09-03
feat: TOOLS-4134 read host_total_mem_kbytes for exact info memory Capacity

a-spiker:
Follow-up to #522, closing the gap that ticket's "Design decision: no estimated capacity" section left open.

SERVER-1546 ([server#1620](https://github.com/citrusleaf/aerospike-server/pull/1620), [enterprise#680](https://github.com/citrusleaf/aerospike-server-enterprise/pull/680)) is merged: the server now emits `host_total_mem_kbytes`, the `MemTotal` it already read from `/proc/meminfo` and previously discarded. `info memory` reads it.

## What changes

`Capacity` is the tracked cgroup limit when the kernel measured one, else the host total when the server reports one. A node with no cgroup limit gets a `Capacity` and an `Alloc%` again, where before both cells were blank. A node whose cgroup limit exists but is not tracked gets `Capacity` from the host total and no `Alloc%`: allocation over the host total would read as headroom the cgroup does not have.

Before (master), against a container with `cgroup-mem-tracking` off that reported a cgroup limit:

```
WARNING: 1.0.0.127.in-addr.arpa:3000: the cgroup limit is not tracked, so Free% and Stop% are measured against the host rather than the cgroup. Enable cgroup-mem-tracking so asadm can also report Capacity and Alloc%.
                        Node|  Build|~~~~~~~~~~~~~Allocated~~~~~~~~~~~~~~| Free%| Stop%
                            |       |   Total|   Shmem|      Heap|  Heap%|      |
*1.0.0.127.in-addr.arpa:3000|8.1.3.0|6.263 GB|5.500 GB|781.453 MB|12.18 %|80.0 %|90.0 %
```

After, against a live server reporting `host_total_mem_kbytes=24569628` with `cgroup-mem-tracking` off and no cgroup limit reported. No warning: nothing is host-scoped that should not be.

```
                        Node|  Build| Capacity|~~~~~~~~~~~~~Allocated~~~~~~~~~~~~~~| Alloc%| Free%| Stop%
                            |       |         |   Total|   Shmem|      Heap|  Heap%|       |      |
*1.0.0.127.in-addr.arpa:3000|8.1.3.0|23.431 GB|4.226 GB|3.500 GB|742.960 MB|17.17 %|18.03 %|81.0 %|90.0 %
```

The untracked-limit case (the Before node's shape) now renders `Capacity` from the host total, leaves `Alloc%` blank, and warns with the first message under Details.

## Nothing is estimated

The removed workaround was a *derived* host total, back-solved from `host_free_mem_bytes` and a rounded `host_free_mem_pct`; #522 deleted it because the rounding error made `Alloc%` untrustworthy on exactly the nodes an operator inspects. This does not bring the derivation back. `host_total_mem_bytes` is the server's own reading, converted from kbytes and nothing more, and a node reporting neither a limit nor a total still renders a blank `Capacity`. `host_total_mem_kbytes=0` (unreadable `/proc/meminfo`) is treated as absent.

## Details

* `util.derive_memory_stats` converts `host_total_mem_kbytes` to `host_total_mem_bytes` alongside the other kbyte stats.
* `util.derive_memory_headline`: `capacity = tracked cgroup limit if > 0, else host total`. `alloc_pct` is withheld when a cgroup limit exists but is not tracked. The third return bucket, previously "nodes with no cgroup limit at all", is now `no_capacity`: nodes with no usable cgroup limit (absent or the uncapped sentinel) and no host total. Classification is by the untracked predicate first, so an untracked node is always in `untracked_limits` and never in `no_capacity`, whether or not a host total arrived. A pre-8.1.3 build is in neither: it cannot report either stat, and the build warning already names it.
* Gating is on the stat being present, not on a version constant, so a server that predates SERVER-1546 collapses `Capacity` and `Alloc%` exactly as it does today. No new entry in `constants.py`.
* Warnings. The view splits `untracked_limits` by whether the node's row rendered `Capacity`, so each message is true of every node it names:
  * `%s: cgroup-mem-tracking is off, so Capacity, Free% and Stop% are host-wide, not cgroup-scoped, and Alloc% is withheld.` for untracked nodes whose Capacity is the host total.
  * `%s: cgroup-mem-tracking is off, so Free% and Stop% are host-wide, not cgroup-scoped. Enable it to report Capacity and Alloc%.` for untracked nodes with no host total. Not gated on the column, so a homogeneous cluster of such nodes still gets the diagnostic.
  * `Capacity is blank for %s: the server reported no cgroup limit and no host total.` for `no_capacity` nodes, still gated on `Capacity` having rendered somewhere.
  * All keep the node list out of subject position, so `summarize_nodes`' singular and `all N nodes` forms both read correctly.
* Command help in both controllers: "Capacity (the tracked cgroup limit, else the host total when the server reports one) ... Alloc% is withheld when a cgroup limit exists but is not tracked."
* `cgroup_limit_or_zero` docstring no longer implies the limit is cross-checked against the host total; it is not.
* `-v` Host and CGroup Memory sheet regains a `Host Total` column, projecting `host_total_mem_bytes`. The round-1 `Host Total*` column was removed with the estimate it displayed; this one is the server's `MemTotal` and collapses on a server that does not report it. With the cgroup limit and host total side by side, the operator can see which source produced the headline `Capacity`.
* `-v` Free `Host` / `Host%` always render when the server reports them. #522 collapsed the pair when it duplicated the System pair (`_collapse_duplicate_host_free`); that made an untracked node look like it was missing a stat, with nothing on screen saying why. The repetition is the signal that Free is host-scoped, so it stays.

## Testing

* `pytest test/unit` - 2073 passed, 1 skipped, 293 subtests. `black --check` clean.
* Unit coverage at the `derive_memory_headline` level: tracked limit wins over a present host total; uncapped cgroup sentinel falls back to the host total, or to `no_capacity` without one; untracked limit with a host total renders Capacity but no Alloc%; untracked limit with no host total, or `host_total_mem_kbytes=0`, is `untracked_limits` only; a build reporting neither stat is `no_capacity`; a pre-8.1.3 build is in neither list. At the `CliView.info_memory` level, every warning-bearing shape: host-wide warning with Alloc% absent, enable-tracking warning in single-node and mixed clusters, blank-Capacity warning in a mixed cluster. The stdout-leak test anchors on the fragments of both warnings it triggers and on every substituted warning.
* Verified against a live 8.1.3.0 server carrying the new stat (`host_total_mem_kbytes=24569628`), walking all four states by toggling `cgroup-mem-tracking` with `manage config` and the container limit with `docker update`:

  | cgroup limit | tracking | Capacity | Alloc% | Free% | warning |
  |---|---|---|---|---|---|
  | none | off | 23.431 GB (host) | 18.05 % | 81 % (host) | none |
  | none | on | 23.431 GB (host) | 18.05 % | 81 % (host) | none |
  | 8 GiB | on | 8.000 GB (cgroup) | 52.88 % | 42 % (cgroup) | none |
  | 8 GiB | off | 23.431 GB (host) | blank | 81 % (host) | host-wide, Alloc% withheld |

  With no limit the server emits no `cgroup_memory_*` stats at all, on or off. With a limit and tracking off the `-v` CGroup columns still show Used 7.5 GB of 8 GB, so the pressure the headline cannot scope is one flag away. The After block above is the first row.
* The e2e fixture moves from `database-docker-test-local:8.1.3.0`, which reports the arena stats but not `host_total_mem_kbytes` (verified: a Capacity assertion fails against it), to `database-docker-dev-local/aerospike-server-enterprise:8.1.3.0-20260904184619`, which carries SERVER-1546. `test_memory` now asserts `Capacity` renders and exceeds `Allocated Total`; `test_memory_verbose` asserts the `Host Total` column. Full e2e suite against the new image: 350 passed, 7 skipped. The CI identity for `tests.yml` is scoped to `database-docker-test-local` per the #522 security thread, so the workflow may need read access to `database-docker-dev-local` before this passes in CI.

## Ticket

TOOLS-4134, whose Follow-up section scopes this: "once `host_total_mem_kbytes` ships, the next asadm release reads it to restore Capacity and Alloc% for nodes without a tracked cgroup limit."

The ticket body still describes the interim behaviour as final ("Capacity renders only where the kernel actually measured a limit") and needs that section updated to match; not done in this PR.


AerospikeNate-L:
**PR Review (version 2026.09.01)** — round 2 at `b371aad5`

**Promises:**
1. Capacity is the tracked cgroup limit when the kernel measured one, else the host total the server reports — never derived or estimated; a node reporting neither renders a blank Capacity.
2. Alloc% is withheld when a cgroup limit exists but is not tracked, so allocation is never divided by a capacity the cgroup does not have.
3. Classification is by the untracked predicate first — an untracked node is always in `untracked_limits`, never in `no_capacity`, independent of whether a host total arrived — and each warning is true of every node it names.
4. Gating is on the stat being present, not a version constant, so a server predating SERVER-1546 collapses Capacity and Alloc% exactly as it does today.
5. Command help in both controllers and the touched docstrings describe the shipped semantics, with no claim of a diagnostic that does not exist.
6. `-v` regains a `Host Total` column projecting the server's MemTotal and collapsing without it; the Free `Host`/`Host%` pair always renders when the server reports it.
7. Unit tests cover the new host-total cases at both the `derive_memory_headline` and `CliView.info_memory` levels; e2e asserts Capacity and Host Total against an image carrying SERVER-1546.

| Gate | Status |
|------|--------|
| Contract | pass |
| Impacted surface | pass |
| Failure paths | pass |
| Evidence | pass |
| Quality | pass |
| Backward Compat | pass |
| Remediation | pass |

Your robot overlords approve this PR.

All eight round-1 findings are addressed, and the three items round 1 could not anchor to a diff line are closed: the description's "After" block and verification table now agree with the code, the "every released server" premise is gone from the test docstrings, and the e2e Capacity exclusion is settled by moving the fixture to an image that carries SERVER-1546.

Verification this round went past reading the diff:

- **Promise 4, by exercising the function.** `derive_memory_headline` across three server generations (neither stat / `host_total_mem_kbytes=0` / both) crossed with three cgroup states, plus the uncapped sentinel and a missing tracking config. A server predating SERVER-1546 produces cells identical to base in every cgroup state, and the two buckets stay mutually exclusive throughout. `Alloc%` withholds nothing that used to render — base already gated it on `capacity > 0`, which was unconditionally 0 for an untracked node.
- **Promise 3's warning text, against the server.** `system_free_mem_pct` really is re-scoped server-side: `get_mem_info_with_cgroup_stats` assigns it from the cgroup figure when `cgroup-mem-tracking` is on and a usable cgroup exists, and stop-writes is evaluated against that same value. So "Free% and Stop% are host-wide, not cgroup-scoped" and the enable-tracking remedy are both true of the nodes each message names.
- **Promise 5's `host_total_mem_kbytes=0` claim, at the source.** `sys_mem_info` zero-initialises the total and returns early when `/proc/meminfo` cannot be read, and the stat is emitted unconditionally — so treating a non-positive total as absent is correct, and the docstring saying so is accurate.
- **Promise 7.** CI at this HEAD: unit 2077 passed / 1 skipped, e2e 350 passed / 7 skipped in 22m39s against `8.1.3.0-20260904184619` — the new Capacity and Host Total assertions ran against a real SERVER-1546 server. All checks green.

The two `CommandHelp` blocks are byte-identical apart from the `usage` line, which differs legitimately.

**Round history:**

| Round | HEAD | Verdict |
|-------|------|---------|
| 1 | `aba77a4a` | request changes — 8 findings (4 Important, 4 Minor) |
| 2 | `b371aad5` | approve — 0 findings |

**Findings to date:**

| Finding (round) | Status |
|-----------------|--------|
| Control Flow: `untracked_limits` reachable only with a host total (`lib/utils/util.py`) (r1) | Addressed in `fa92b2c` |
| Style: untracked warning omitted that Alloc% is blank (`lib/view/view.py`) (r1) | Addressed in `fa92b2c` — split into two messages; Stop% wording kept, author's call |
| Style: comment-only deletions (`lib/live_cluster/collectinfo_controller.py`) (r1) | Addressed in `cf3b42e` |
| Data Flow: docstring implied a cgroup-vs-host cross-check (`lib/utils/util.py`) (r1) | Addressed in `cf3b42e` — wording fixed, no check added |
| Style: `derive_memory_headline` docstring lost `builds`/`nodes` (`lib/utils/util.py`) (r1) | Addressed in `cf3b42e` — restored in prose; gating paragraph left out, author's call |
| Tests: stdout-leak test lost its fragment anchors (`test/unit/view/test_view.py`) (r1) | Addressed in `b371aad` |
| Tests: e2e docstring contradicted the pinned image (`test/e2e/live_cluster/test_info.py`) (r1) | Addressed in `cf3b42e` |
| Style: command help over-claimed the host-total fallback (both controllers) (r1) | Addressed in `fa92b2c` |

lib/utils/util.py
```diff
@@ -765,13 +730,16 @@ def derive_memory_headline(stats, configs, ns_agg, builds=None, nodes=None):
...
+        cgroup_untracked = cgroup_limit > 0 and not cgroup_tracked
+        host_total = max(0, int_or_zero(node_stats.get("host_total_mem_bytes")))
+        capacity = cgroup_limit if cgroup_tracked and cgroup_limit > 0 else host_total
+
+        if node_stats:
+            if capacity <= 0:
+                if alloc_known:
+                    no_capacity.append(node)
+            elif cgroup_untracked:
                 untracked_limits.append(node)
```
AerospikeNate-L, line 740: From Control Flow agent:

`untracked_limits` is reachable only through this `elif`, which requires `capacity > 0` and therefore requires `host_total_mem_bytes` — so whether an untracked cgroup limit gets diagnosed depends on an unrelated stat's presence. On a build reporting `cgroup_memory_limit_bytes` but not yet `host_total_mem_kbytes`, a node with a real cgroup limit and `cgroup-mem-tracking` off has `capacity == 0`, is classified `no_capacity` instead, and that warning is gated on `capacity_rendered` — so a homogeneous cluster, or `info memory with <node>` narrowed to the affected node, emits nothing at all. Base always put that node in `untracked_limits` and always warned "Free% and Stop% are measured against the host rather than the cgroup. Enable cgroup-mem-tracking".

Two populations reach it. The two server stats landed 47 commits apart on master (`018915e81` Jul 31, `d36d281bd` Sep 3) and neither is in the `8.1.3.0` tag, so any build cut from that window is in the gap — including whatever `8.1.3.0` image the e2e fixture pulls. That one closes on its own. The one that does not: `sys_mem_info` zero-initialises `total_mem_kbytes` and returns early when `/proc/meminfo` is unreadable, and the server emits the stat unconditionally, so `host_total_mem_kbytes=0` reproduces the same misclassification on any build carrying both commits.

Classify by the untracked predicate first, so membership stops depending on whether a host-total fallback happened to exist:

```suggestion
        if node_stats:
            if cgroup_untracked:
                untracked_limits.append(node)
            elif capacity <= 0 and alloc_known:
                no_capacity.append(node)
```

Buckets stay mutually exclusive and the headline rows are unchanged — `capacity_bytes` and `alloc_pct` are computed above this block, never from list membership. Dropping `alloc_known` from the untracked branch is safe: `cgroup_memory_limit_bytes` shares the 8.1.3 gate with the alloc stats (`constants.py:233`), so a build that cannot report alloc stats cannot produce `cgroup_limit > 0` either, and the remaining overlap is a node whose build string failed to parse while its stats arrived — the warning's content comes from that node's own stats and config, not from the build.

Three tests then need assertions, docstrings and names updated, since all three encode the old classification: `test_untracked_cgroup_limit_without_host_total_is_only_no_capacity`, `test_info_memory_untracked_limit_without_host_total_is_silent` (its `== []` assertion, not just its wording), and `test_info_memory_untracked_limit_without_host_total_in_a_mixed_cluster`, where the surviving warning becomes the untracked one naming node2. The silent test's docstring premise needs correcting too — "Every released server reports `cgroup_memory_limit_bytes` without `host_total_mem_kbytes`" is not right, since `018915e81` is not in `8.1.3.0` and so no tagged release reports that stat at all.

lib/view/view.py
```diff
@@ -199,21 +199,19 @@ def _warn_memory_gaps(
 
         if untracked_limits:
             logger.warning(
-                "%s: the cgroup limit is not tracked, so Free%% and Stop%% are "
-                "measured against the host rather than the cgroup. Enable "
-                "cgroup-mem-tracking so asadm can also report Capacity and "
-                "Alloc%%.",
+                "%s: cgroup-mem-tracking is off, so Capacity, Free%% and Stop%% "
+                "are host-wide, not cgroup-scoped.",
```
AerospikeNate-L: From Style agent:

Alloc% is blank for every node this warning names — `alloc_pct` is gated on `not cgroup_untracked` at `util.py:765`, and a node reaches `untracked_limits` only when `cgroup_untracked` is true — but the message lists only Capacity, Free% and Stop%. The wording it replaced did account for it ("Enable cgroup-mem-tracking so asadm can also report Capacity and Alloc%"). Separately, Stop% is the configured `stop-writes-sys-memory-pct` minimum (`util.py:557-574`), not a measurement, so it is not itself "host-wide"; what changes with tracking is the memory pct it gets evaluated against.

Once the bucket fix on `util.py:741` lands this list also contains nodes whose Capacity cell is blank, so one rewording should carry all of it. Apply that fix first, then:

```suggestion
                "%s: cgroup-mem-tracking is off, so Free%% and Stop%% are "
                "host-wide and Alloc%% is withheld; Capacity shows the host "
                "total where the server reports one, else blank.",
```

Note the `%%` on `Alloc%%` — a bare `Alloc%` makes `logging` read `% i` as a conversion specifier and raise `TypeError: not enough arguments for format string`, replacing the warning with a logging-error traceback.

lib/live_cluster/info_controller.py
```diff
@@ -97,13 +97,14 @@ async def do_network(self, line):
...
-        "cgroup-mem-tracking is enabled) vs Allocated (the shmem index and",
-        "sindex arenas plus the process heap). Memory-engine data is in the arenas",
-        "on Enterprise and Federal, and in the heap on Community. The verbose",
-        "tables are not additive with each other: set index, and memory-engine",
-        "data on Community, are heap-allocated and appear in both Index and Data",
-        "Memory and Process Heap. Allocation figures require server",
+        "Displays node memory: Capacity (the tracked cgroup limit, else the host",
+        "total the server reports) vs Allocated (the shmem index and sindex arenas",
+        "plus the process heap). When a cgroup limit exists but is not tracked,",
+        "Capacity is the host total and Alloc% is withheld. Memory-engine data is",
```
AerospikeNate-L: From Style agent:

"When a cgroup limit exists but is not tracked, Capacity is the host total and Alloc% is withheld" is false on every released server: none report `host_total_mem_kbytes` yet, so Capacity is blank rather than the host total — which is what `test_untracked_cgroup_limit_without_host_total_is_only_no_capacity` asserts. Qualify it, e.g. "Capacity falls back to the host total when the server reports one, else it is blank; Alloc% is withheld either way." The same sentence is duplicated at `lib/collectinfo_analyzer/info_controller.py:97` and needs the same change.

lib/live_cluster/collectinfo_controller.py
```diff
@@ -2017,8 +2017,6 @@ async def _dump_collectinfo_ascollectinfo(
         logger.info(f"Capturing pretty print output for {file} . . .")
 
         try:
-            # memory is verbose here so the bundle carries the host/index/heap
-            # breakdown behind the headline sheet, not just the headline.
             dignostic_info_params = [
```
AerospikeNate-L, line 2022: From Style agent:

Both hunks in this file are comment-only deletions with no code change, and nothing in this PR's scope accounts for them. The comment above `dignostic_info_params` was the only thing explaining why this list uses `memory -v` while `summary_info_params` uses plain `memory`; without it the asymmetry reads as a slip. The one deleted above the `InfoController()` try block was the rationale for the six repeated `try/except: write_to_file` blocks in this function — the isolation a #522 round asked for, and exactly the shape a later duplication pass would try to collapse into a shared `try`. Restore both.

lib/utils/util.py
```diff
@@ -481,9 +481,9 @@ def cgroup_limit_or_zero(value):
     An uncapped cgroup reports either 'max' (v2), a negative value, or a
     near-int64 sentinel (v1), and CGROUP_MEMORY_NO_LIMIT_THRESHOLD catches every
     one of those without reference to anything else. Any other value the kernel
-    reports is a measurement and is reported as given, even when it exceeds host
-    RAM: that is a real misconfiguration the operator needs to see, and asadm
-    has no total-memory figure of its own to check it against.
+    reports is a measurement and is reported as given, even when it exceeds the
```
AerospikeNate-L: From Data Flow agent:

The reworded docstring calls a cgroup limit exceeding "the host total the server reports" a misconfiguration "the operator needs to see", but nothing surfaces it: no code compares `cgroup_memory_limit_effective_bytes` against `host_total_mem_bytes`, and `Capacity` has no formatter in `templates.py` (unlike `Free%`, which has `sys_memory_free_pct_formatters`). The clause this replaced was accurate precisely because it disclaimed the comparison — "asadm has no total-memory figure of its own to check it against" — and now that asadm does have one, the sentence reads as a diagnostic that exists. Either state that the two figures are not cross-checked, or add the check.

lib/utils/util.py
```diff
@@ -690,51 +690,16 @@ def nodes_missing_memory_alloc_stats(builds):
 
 def derive_memory_headline(stats, configs, ns_agg, builds=None, nodes=None):
     """
-    Build the per-node rows behind the 'info memory' headline table.
+    Build the per-node rows behind the 'info memory' headline table: Capacity is
```
AerospikeNate-L, line 692: From Style agent:

Deleting the paragraph that claimed the server publishes no total-memory stat is right — it is now false. The same trim also took two things the code still does: the `Args:` block documenting that `builds=None` treats every node as reporting alloc stats and `nodes=None` defaults to every node in `stats` (both still live, at lines 723 and 705), and the paragraph explaining why the allocation total and Shmem are gated as they are — lines 755-766, which this PR touches only for one condition. `aggregate_ns_memory_stats` in this same file keeps a full `Args:`/`Returns:` block for comparable complexity. Worth restoring those two pieces.

test/unit/view/test_view.py
```diff
@@ -2865,11 +2948,14 @@ def test_info_memory_warnings_never_reach_stdout(self):
...
+        warnings = self.warnings()
+        self.assertTrue(any("cgroup-mem-tracking is off" in w for w in warnings))
+        self.assertTrue(any("No namespace statistics" in w for w in warnings))
         self.assertIn("Memory Information", printed)
         self.assertNotIn("WARNING", printed)
-        self.assertNotIn("cgroup limit is not tracked", printed)
-        self.assertNotIn("No namespace statistics", printed)
+
+        for warning in warnings:
+            self.assertNotIn(warning, printed)
```
AerospikeNate-L, line 2974: From Tests agent:

Replacing the two fixed-substring checks with `assertNotIn(warning, printed)` is weaker rather than a superset: the loop only catches a leak of the complete substituted string. Confirmed by mutation — printing the first 40 characters of the real substituted warning next to the `logger.warning` call leaves this test passing, while a check for `"cgroup-mem-tracking is off"` catches it. The same gap covers a leak of the unsubstituted `%s` template, which the loop can never match because it only ever compares against already-substituted strings. Keep the loop, which does generalize to all four warnings, and restore the fragment anchors alongside it:

```suggestion
        self.assertNotIn("WARNING", printed)
        self.assertNotIn("cgroup-mem-tracking is off", printed)
        self.assertNotIn("No namespace statistics", printed)

        for warning in warnings:
            self.assertNotIn(warning, printed)
```

test/e2e/live_cluster/test_info.py
```diff
@@ -164,8 +164,9 @@ async def test_memory(self):
         This test asserts <b> info memory </b> output heading, the headline
         columns, and the allocation values. The fixture pins server 8.1.3+
         (lib.SERVER_TAG), so the arena stats and every allocation column must
-        render; Capacity still needs the container to report a cgroup limit
-        with cgroup-mem-tracking enabled, so it is not asserted here.
+        render; Capacity still needs the container's server to report
```
AerospikeNate-L: From Tests agent:

This docstring says Capacity is not asserted because the pinned image may or may not carry `host_total_mem_kbytes`, while the PR description says the change was verified against a live 8.1.3.0 server that does report it (`host_total_mem_kbytes=24569628`). `lib.SERVER_TAG` pins `8.1.3.0` from `database-docker-test-local`, a test registry where a tag can be rebuilt in place, so both statements can be true of different pulls — but only one describes the image this test actually runs against. Worth settling with `asinfo -v statistics` against the pinned image: if it reports the stat, Capacity is assertable here now and the exclusion is stale.

lib/live_cluster/info_controller.py
AerospikeNate-L: From Style agent:

"When a cgroup limit exists but is not tracked, Capacity is the host total and Alloc% is withheld" holds only when the server also reports `host_total_mem_kbytes`. On a build that reports the cgroup limit but not the host total, Capacity is blank rather than the host total — which is what `test_untracked_cgroup_limit_without_host_total_is_only_no_capacity` asserts, and what a node with an unreadable `/proc/meminfo` gets on any build. Qualify it, e.g. "Capacity falls back to the host total when the server reports one, else it is blank; Alloc% is withheld either way." The same sentence is duplicated at `lib/collectinfo_analyzer/info_controller.py:97` and needs the same change.

lib/utils/util.py
a-spiker: Done, took the suggestion as is: untracked is classified first, `no_capacity` only when nothing else applies. The three tests are renamed and re-asserted, and the "every released server" premise is gone from the docstrings. Added a `host_total_mem_kbytes=0` subtest for the unreadable `/proc/meminfo` case you raised.

lib/view/view.py
a-spiker: Went a different way. Kept the short message and appended "and Alloc% is withheld". Untracked nodes with no host total get their own line instead: "cgroup-mem-tracking is off, so Free% and Stop% are host-wide, not cgroup-scoped. Enable it to report Capacity and Alloc%." Each message is now true of every node it names. Left the Stop% wording alone.

lib/live_cluster/collectinfo_controller.py
a-spiker: Restored both. Out of scope here, agreed.

lib/utils/util.py
a-spiker: Docstring now says the limit is reported as given and asadm does not cross-check it against the host total. No check added.

lib/utils/util.py
a-spiker: Put the `builds=None` and `nodes=None` lines back. Left the gating paragraph out on purpose, keeping docstrings short; the rationale lives in the test docstrings.

test/unit/view/test_view.py
a-spiker: Both anchors restored next to the loop, as suggested.

test/e2e/live_cluster/test_info.py
a-spiker: Settled it by running against the pinned image: the `database-docker-test-local` 8.1.3.0 tag reports the arena stats but not `host_total_mem_kbytes`. Moved the fixture to `database-docker-dev-local/aerospike-server-enterprise:8.1.3.0-20260904184619`, which carries SERVER-1546. `test_memory` now asserts Capacity and `test_memory_verbose` asserts Host Total, both passing against it.

a-spiker:
Decisions from this round, so they are in one place:

- Capacity is the tracked cgroup limit, else the host total when the server reports one. Nothing derived.
- On a node with a cgroup limit that is not tracked, Capacity still shows the host total but Alloc% is withheld. Same reasoning as the untracked-cgroup thread in #522: allocation over the host total reads as headroom the cgroup does not have.
- Untracked nodes are classified first and never land in `no_capacity`. Two warning variants, depending on whether that node rendered a Capacity, so a homogeneous cluster on a build without `host_total_mem_kbytes` still gets the enable-tracking hint.
- Kept the short warning wording. Only addition is "and Alloc% is withheld".
- `-v` gets the Host Total column back. It was removed because it was an estimate; it is the server's MemTotal now.
- `-v` Free Host / Host% always render. The duplicate collapse from #522 made an untracked node look like it was missing a stat.
- Help text stays inline in both controllers. Docstrings kept short.
- e2e fixture moved to `database-docker-dev-local` 8.1.3.0-20260904184619, which carries SERVER-1546. Capacity and Host Total are asserted.
- Not changed: Stop% wording in the warning. TOOLS-4134 follow-up text still says "nodes without a tracked cgroup limit"; shipped behaviour restores Alloc% only for nodes with no limit. Will fix on the ticket.

lib/live_cluster/info_controller.py
a-spiker: Reworded in both controllers: "Capacity (the tracked cgroup limit, else the host total when the server reports one)", with "Alloc% is withheld when a cgroup limit exists but is not tracked" as its own sentence.

test/e2e/lib.py
```diff
@@ -50,9 +50,9 @@
 
 IMAGE_REPO = os.environ.get(
     "ASADM_E2E_IMAGE_REPO",
-    "artifact.aerospike.io/database-docker-test-local/aerospike-server-enterprise",
+    "artifact.aerospike.io/database-docker-dev-local/aerospike-server-enterprise",
```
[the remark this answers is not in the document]
a-spiker: I will take care of this once we have official RC image. for now we are pulling this change from the dev stage.

lib/utils/util.py
AerospikeNate-L: Resolved by fa92b2c — untracked is classified first, so bucket membership no longer depends on whether a host total arrived. See https://github.com/aerospike/aerospike-admin/blob/b371aad5615fed0e68e92895452e2083707d5351/lib/utils/util.py#L739

lib/view/view.py
AerospikeNate-L: Resolved by fa92b2c — split into two messages, each naming Alloc%: the host-wide one for untracked nodes whose Capacity rendered, the enable-tracking one for those with no host total. Stop% wording left as-is per your reply. See https://github.com/aerospike/aerospike-admin/blob/b371aad5615fed0e68e92895452e2083707d5351/lib/view/view.py#L205-L224

lib/live_cluster/collectinfo_controller.py
AerospikeNate-L: Resolved by cf3b42e — both comments restored. See https://github.com/aerospike/aerospike-admin/blob/b371aad5615fed0e68e92895452e2083707d5351/lib/live_cluster/collectinfo_controller.py#L2020 and https://github.com/aerospike/aerospike-admin/blob/b371aad5615fed0e68e92895452e2083707d5351/lib/live_cluster/collectinfo_controller.py#L2121

lib/utils/util.py
AerospikeNate-L: Resolved by cf3b42e — the docstring now says the limit is reported as given and that asadm does not cross-check it against the host total. See https://github.com/aerospike/aerospike-admin/blob/b371aad5615fed0e68e92895452e2083707d5351/lib/utils/util.py#L484-L485

lib/utils/util.py
AerospikeNate-L: Resolved by cf3b42e — builds=None and nodes=None are documented again. The gating paragraph stays out per your reply. See https://github.com/aerospike/aerospike-admin/blob/b371aad5615fed0e68e92895452e2083707d5351/lib/utils/util.py#L696-L697

test/unit/view/test_view.py
AerospikeNate-L: Resolved by b371aad — both fragment anchors are back alongside the loop, plus positive assertions that each warning actually fired. See https://github.com/aerospike/aerospike-admin/blob/b371aad5615fed0e68e92895452e2083707d5351/test/unit/view/test_view.py#L2962-L2975

test/e2e/live_cluster/test_info.py
AerospikeNate-L: Resolved by cf3b42e — settled against the pinned image; the fixture moved to one carrying SERVER-1546 and Capacity is now asserted. CI at b371aad ran the e2e suite green against it (350 passed, 7 skipped). See https://github.com/aerospike/aerospike-admin/blob/b371aad5615fed0e68e92895452e2083707d5351/test/e2e/live_cluster/test_info.py#L162-L206

lib/live_cluster/info_controller.py
AerospikeNate-L: Resolved by fa92b2c — reworded to "else the host total when the server reports one" in both controllers. See https://github.com/aerospike/aerospike-admin/blob/b371aad5615fed0e68e92895452e2083707d5351/lib/live_cluster/info_controller.py#L101 and https://github.com/aerospike/aerospike-admin/blob/b371aad5615fed0e68e92895452e2083707d5351/lib/collectinfo_analyzer/info_controller.py#L95

test/e2e/lib.py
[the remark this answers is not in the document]
AerospikeNate-L: Still open at b371aad — line 53 still points at database-docker-dev-local.

The access half of this is settled empirically, though: the Tests workflow runs the full e2e suite on every branch push, and it ran green against this image both at cf3b42e (the commit that switched it) and at b371aad — 350 passed, 7 skipped. So the OIDC identity does have read access to dev-local, and the PR description’s "may need read access to database-docker-dev-local before this passes in CI" caveat is stale.

What is left is the re-pin to an official RC tag, already committed to above. Worth keeping open until that lands: a timestamped dev-local tag is subject to artifact retention, so the e2e suite breaks whenever it is pruned.

AerospikeNate-L (approved):

PR #551, opened 2026-09-03
feat(docker): TOOLS-4360 help entrypoint that lists the tools the image ships

a-spiker:
Rescoped to TOOLS-4360 only. The earlier revision of this PR bundled nine changes; this one is the help entrypoint, its tests, and nothing else. Branch restarted from `master`, so it carries the ubuntu:24.04 base, the C.UTF-8 locale (TOOLS-4361) and the shared-workflows image build (TOOLS-4364) as-is.

## What changes

`docker/docker-entrypoint.sh` becomes the image's `ENTRYPOINT`; `CMD` is removed. The script is a docker-library-style passthrough: it `exec`s whatever the caller names, and only speaks up when nothing runnable was named.

| invocation | before | now |
| --- | --- | --- |
| `docker run --rm IMAGE` | ran `asadm` with no arguments | usage on stdout listing `asadm` and `asinfo`, exit 0 |
| `echo '...' \| docker run -i IMAGE` or `docker run -i IMAGE < file` (piped stdin, no arguments) | fed the commands to `asadm` (its interactive prompt) | usage on stdout, exit 0, stdin never read: name the binary, `docker run -i IMAGE asadm` |
| `docker run --rm IMAGE --version` | runc: `exec: "--version": executable file not found in $PATH`, exit 127 | usage on stderr naming the argument, exit 127 |
| `docker run --rm IMAGE no-such-tool` | runc error, exit 127 | usage on stderr, exit 127 |
| `docker run --rm IMAGE asadm --version` | ran | unchanged |
| `docker run --rm -it IMAGE bash` | ran | unchanged (passthrough) |
| `docker run --rm --entrypoint bash IMAGE -c '...'` | ran | unchanged |
| pod spec with `command:` and/or `args:` | ran | unchanged (`command:` replaces the entrypoint; `args:` alone is exec'd as given) |

Only the first four rows change, and the first is the point of the ticket. The piped-stdin row is the one break a caller cannot see from the exit code, so it belongs in the release note: several of the tools images ship more than one binary, so no default is correct and forgetting the name should say so.

Why this shape:
- `docker-entrypoint.sh` ending in `exec "$@"` is the Docker Official Images convention (postgres, redis, mysql, docker:cli): the entrypoint adds behaviour and hands off to whatever was named, without narrowing what the image can run. An allow-list would add no security (`--entrypoint`, `docker exec` and a pod `command:` all bypass it) and would only break callers.
- A multi-command CLI run with no subcommand prints its usage and exits 0 (`kubectl`, `docker`, `gh`); this image is a multi-command CLI.
- 127 is the shell's "command not found" status and what runc already returned, so callers testing for it see the same code, now with an explanation.

Mechanics:
- The tool list is generated at build time from `dpkg-query -L` on the installed package into `/usr/local/share/aerospike/tools`, so it follows the package. It only feeds the usage text; exec never reads it.
- The usage refusal covers names and paths that resolve to nothing. A path that exists but is not executable passes through to exec and fails there with exit 126 and dash's own message, exactly as under runc; a bats case pins that boundary.
- `COPY --chmod=0755` pins the script's mode instead of relying on the git index.
- `.dockerignore` re-includes `docker/docker-entrypoint.sh`, since the shared-workflows build context is the repo root; `docker-build.sh` overrides `ENTRYPOINT_SRC` for its `docker/` context, so both build paths use one `COPY` line.

## Removed from this PR

Obsolete since TOOLS-4364 moved image build and publish to `aerospike/shared-workflows` (`reusable_docker-build-deploy.yaml@v4.1.0`), which owns tagging, provenance and attestation of the published index: SLSA provenance and SBOM attestations in `docker-build.sh`; dynamic OCI `version`/`revision`/`created` labels; `--latest` rolling tags and the `build-check` tag-policy assertions (rolling tags belong to the Docker Hub path, TOOLS-4177); the `--dry-run` change that existed to feed those assertions.

Deferred, each a behaviour change worth its own ticket and review: tini as PID 1 (signal handling for tools without a SIGTERM handler); `ca-certificates`; an unprivileged `aerospike` user; a hadolint/shellcheck `docker-lint` job.

## Acceptance criteria (TOOLS-4360)

- [x] `docker run --rm IMAGE` prints usage listing the tools the image ships, exit 0
- [x] `docker run --rm IMAGE --version` and `docker run --rm IMAGE no-such-tool` print usage on stderr, nothing on stdout, exit 127
- [x] `docker run --rm IMAGE <tool> [args]` behaves exactly as before
- [x] `docker run --rm -it IMAGE bash` and `--entrypoint bash` still give a shell
- [x] the tool list is generated at build time, not hardcoded
- [x] the script is unit-tested on every pull request; the contract is asserted against the built image on amd64 and arm64

## Verification

- `bats .github/bin/test/test_entrypoint.bats`: 13 cases, green on macOS and under dash in `ubuntu:24.04`; wired into `build-check` (its own job, "Docker entrypoint (bats)", distro bats).
- `shellcheck docker/docker-entrypoint.sh` clean; `docker build --check` clean; `actionlint` reports nothing on the edited regions.
- Built the image locally on arm64 from the repo root (the CI context) using the ubuntu24.04 `.deb` from run 34879616668, and ran the new smoke-test block against it verbatim: passes. Also built through `docker/docker-build.sh -t` (the `docker/` context): same entrypoint, same usage.
- Build and Release dry run (`release=false`) on this branch: https://github.com/aerospike/aerospike-admin/actions/runs/34898875010 (green: `Docker build & smoke test` passed on amd64 and arm64 with the new entrypoint assertions, and `Docker build & publish` (push disabled) built from the shared-workflows context, which exercises the `.dockerignore` re-include)
- Usage from the local build:

```console
$ docker run --rm IMAGE
Aerospike asadm image (aerospike-asadm 5.0.3-5bf7abff0-1ubuntu24.04)

This image ships more than one tool, so name the one you want:

  docker run --rm <image> asadm  [args]   Aerospike Admin
  docker run --rm <image> asinfo [args]   Aerospike info command-line tool

Examples:
  docker run --rm <image> asadm --help
  docker run --rm -ti <image> asadm [args]   (add -ti for interactive tools)

For a shell inside the image:
  docker run --rm -ti --entrypoint bash <image>

$ docker run --rm IMAGE --version; echo rc=$?
Error: '--version' is an option, not a binary. Name the binary first: docker run --rm <image> <binary> --version

Aerospike asadm image (aerospike-asadm 5.0.3-5bf7abff0-1ubuntu24.04)
...
rc=127
```

## Coordinated change, one PR per image

- aerospike/aerospike-admin#551
- aerospike/aql#99
- aerospike/aerospike-tools-backup#176
- aerospike/aerospike-benchmark#168
- aerospike/asconfig#141
- citrusleaf/aerospike-tools#451 (bundle)

`docker-entrypoint.sh`, `test_entrypoint.bats` and `.dockerignore` are byte-identical across the six; the Dockerfile, workflow and build-script hunks differ only in the tool names. Independent of each other; merge in any order.

TOOLS-4360


a-spiker:
Coordinated change, one PR per image. Each is stacked on that repo's TOOLS-4361 locale PR, so merge the 4361 set first.

- citrusleaf/aerospike-tools#451 (bundle) — on #450
- aerospike/aerospike-admin#551 — on #549
- aerospike/aql#99 — on #98
- aerospike/aerospike-tools-backup#176 — on #175
- aerospike/aerospike-benchmark#168 — on #167
- aerospike/asconfig#141 — on #140

a-spiker:
Rebased onto the default branch: this PR no longer depends on the TOOLS-4361 locale PR and can merge in either order. Whichever of the two lands second may need a trivial conflict fix, since both add assertions to the same docker smoke-test step.

Coordinated change, one PR per image:

- citrusleaf/aerospike-tools#451 (bundle)
- aerospike/aerospike-admin#551
- aerospike/aql#99
- aerospike/aerospike-tools-backup#176
- aerospike/aerospike-benchmark#168
- aerospike/asconfig#141

pvinh-spike:
**PR Review (version 2026.07.31)** — round 3, at head 4be0faa3 (one linear follow-up commit; applied identically across the six coordinated PRs, shared files re-verified byte-identical by md5). Set-wide analysis lives on citrusleaf/aerospike-tools#451.

| Gate | Status |
|------|--------|
| Contract | pass |
| Impacted surface | pass |
| Failure paths | pass — the new -126 bats case verified empirically this round (runs under dash, skips under bash, fails a mutated entrypoint) |
| Evidence | pass — "Docker entrypoint (bats)" and "Naming scripts" both green on this head |
| Quality | pass |
| Backward Compat | pass — compat table now carries the piped-stdin row with this image's asadm-specific before-behaviour |

Your robot overlords approve this PR.

**Round-2 finding:**
| Finding | Status |
|---------|--------|
| Architecture: piped-stdin break (commands fed to asadm's prompt) undocumented | Addressed — compat table row + release-note flag in 4be0faa3 |

(Round-1 findings from the pre-restart branch remain resolved as before.)

pvinh-spike:
From Cross-Cutting agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

The PR body says "Stacked on the TOOLS-4361 locale PR, which is this PR's base." That is not the current state of this branch:

- `.base.ref` is this repo's default branch, not `TOOLS-4361`.
- `git merge-base --is-ancestor <TOOLS-4361 head> <this head>` returns false — the locale PR's commit is not in this branch, and grepping this tree's Dockerfile for `LC_CTYPE` finds nothing.
- `git merge-tree` of the two heads reports a content conflict in `.github/workflows/build-and-release.yml`.

Both PRs insert into the same region of the smoke-test step (right after the `astools.conf` check), so whichever merges second conflicts.

I checked all six repo pairs in this coordinated set and the pattern is identical in every one: the TOOLS-4360 PR targets the default branch, does not contain its TOOLS-4361 counterpart, and conflicts with it in `build-and-release.yml`. In the bundle repo (citrusleaf/aerospike-tools#451) `docs/bundle-release-pipeline.md` conflicts too, and there the two versions of the same sentence each drop the other's assertion — so a naive resolution silently loses content.

There is also a merge-order consequence: this PR alone does not set the locale `ENV`. If it merges first, the published image has no UTF-8 locale until the TOOLS-4361 PR lands — and this PR's verification section claims `locale charmap = UTF-8` was confirmed, which cannot have been against this tree alone.

## Reasoning Chain

1. Read the PR body's stacking claim, then checked `.base.ref` via the API.
2. Fetched both PR heads into a local clone and ran `git merge-base --is-ancestor`; confirmed the locale change is absent from this tree by grepping the Dockerfile.
3. Ran `git merge-tree --write-tree` on the two heads and observed the workflow conflict.
4. Repeated all three checks across the other five repo pairs — same result everywhere, so this is a property of how the set was branched rather than a one-off.

## Solution

Land the TOOLS-4361 locale PR first in each repo — it is the smaller, lower-risk change and the one this PR's verification assumes — then rebase this branch onto it, so the conflict is resolved once and deliberately by the author rather than at merge time.

Either way, update the body's "which is this PR's base" sentence to match reality, or actually retarget the base so GitHub shows the stack. Right now the body and the branch disagree, and a reviewer following the body will look for the locale change in this diff and not find it.

docker/docker-build.sh
```diff
@@ -587,6 +642,23 @@ function main() {
     exit 1
   fi
 
+  # `latest` and the rolling major/minor tags mean "the current GA release", so
+  # a pre-release must never move them.
+  if [[ "${MOVING_TAGS}" == true ]]; then
+    if [[ "${VERSION}" == *-rc* ]]; then
```
pvinh-spike: From Logic agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

This guard tests the substring `-rc`, but `rc` is not the only pre-release identifier this repo's own version grammar accepts, and the tags are derived from a different variable than the one being tested.

`.github/scripts/pkg_release.sh` documents and implements the distinction explicitly:

```
#   5.0.3-beta1              -> version 5.0.3-beta1           iteration 1
# Only "rcN" is folded; any other pre-release identifier is left in the version
```

so `-beta1`, `-alpha2` and `-master.abc123` are all first-class accepted inputs that survive into `PKG_VERSION`. `moving_tags` then strips everything from the first `.` onward, so the suffix — which lives in the patch component — is silently discarded and the rolling tags come out looking GA.

Trigger, which I ran in the review tree:

1. `docker/docker-build.sh -p -v 9.9.9-beta1 -N -L`
2. Line 651: `"9.9.9-beta1" == *-rc*` is false → passes.
3. `moving_tags`: `major=9`, `minor=0.3-beta1` → `%%.*` → `9`.
4. Emitted tags:

```
"aerospike/aerospike-tools:9.9.9-beta1-1",
"aerospike/aerospike-tools:latest",
"aerospike/aerospike-tools:9",
"aerospike/aerospike-tools:9.9"
```

`-v 9.9.9-rc1 -N -L` is correctly refused, so the guard works for the one spelling it knows.

Four places assert the property this does not have: this help text ("GA versions only"), the error message on the next line, the `moving_tags` comment ("refuses an rcN version, so a pre-release can never move `latest`"), and `build-check.yml`'s "OK: latest / major / major.minor are GA-only". The same guard is the only thing protecting `-M -L`, which pushes those tags for real via `imagetools create`.

Severity is important rather than critical only because nothing in the pipeline passes `-L` yet.

## Reasoning Chain

1. Read `moving_tags` — both `major` and `minor` derive from `PKG_VERSION`. Read the guard — it tests `VERSION`, for one literal substring. Two different strings, two different predicates.
2. "GA" and "not rc" are the same set only if rc is the only pre-release form. `pkg_release.sh`'s header and its accept regex `^[0-9]+\.[0-9]+\.[0-9]+([-.][A-Za-z0-9.]+)*$` say otherwise.
3. Traced the parameter expansions on `9.9.9-beta1`: the suffix sits in the patch component, which both `%%.*` and `#*.` discard — so the tags are spelling-independent while the guard is spelling-specific.
4. Ran both invocations and captured the output above. The gap is real, not theoretical.

## Solution

Guard on shape, against the value the tags are actually derived from:

```bash
if [[ ! "${PKG_VERSION}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
  log_error "--latest refuses a non-GA version (${VERSION} -> ${PKG_VERSION}); rolling tags are GA-only."
  exit 1
fi
```

`PKG_VERSION` is assigned at line 597, well before this block, so no reordering is needed. This also eliminates by construction the latent `13` → `13.13` case in `moving_tags`.

Add the case to the CI gate so it cannot regress:

```bash
if docker/docker-build.sh -p -v 9.9.9-beta1 -N -L >/dev/null 2>&1; then
  echo "::error::--latest accepted a pre-release version" >&2
  exit 1
fi
```

This will now also refuse a legitimate non-pre-release suffix like `13.0.3-master.abc123` together with `-L` — which is the correct refusal; a dev build should not be `latest` either.

.github/workflows/build-check.yml
```diff
@@ -179,3 +179,47 @@ jobs:
...
+      # during a release.
+      - name: Assert the rolling-tag policy
+        run: |
+          set -euo pipefail
+          if docker/docker-build.sh -p -v 9.9.9-rc1 -N -L >/dev/null 2>&1; then
+            echo "::error::--latest accepted a pre-release version" >&2
+            exit 1
+          fi
+          ga=$(docker/docker-build.sh -p -v 9.9.9 -N -L)
+          for want in "aerospike-asadm:latest" "aerospike-asadm:9" "aerospike-asadm:9.9"; do
```
pvinh-spike: From Logic agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

Two of these three assertions are vacuous — they pass even when `--latest` emits no rolling tags at all.

`grep -qF` is an unanchored substring match, and the bake file always contains the version tag `"aerospike/<image>:9.9.9-1"`. That string contains both `<image>:9` and `<image>:9.9` as substrings. So two of the three `want` values are satisfied by the *version* tag, not by the rolling tags they are meant to check.

I generated the bake file for the bundle repo with `-p -v 9.9.9 -N` (no `-L`, i.e. rolling tags entirely absent) and ran the equivalent three greps against it — the same shape applies verbatim here, since the image name is a common prefix of all the tags:

```
FAIL (correct)  : <image>:latest
PASS (vacuous!) : <image>:9
PASS (vacuous!) : <image>:9.9
```

So if a future change breaks just the `<major>`/`<major>.<minor>` tags — `moving_tags` emitting only `latest`, say — this gate stays green. The step's discriminating power is one third of what it reads as, and it is masked today only because `latest` happens not to be a substring of any other tag.

This matters more than a normal weak-test finding because it is the gate that would otherwise catch the `-beta1` guard gap flagged separately on `docker-build.sh:651`.

## Reasoning Chain

1. Ran `-p -v 9.9.9 -N -L` and captured the push target's tags: `9.9.9-1`, `9.9.9-1-<timestamp>`, `latest`, `9`, `9.9`.
2. Observed that `9.9.9-1` lexically contains `9` and `9.9` immediately after the `<image>:` prefix, so the two shorter needles are prefixes of the version tag line.
3. Ran the same command *without* `-L` and applied the CI's exact `grep -qF -- "${want}" <<<"${ga}"` for each want. Two passed against output that contains no rolling tags — which is the definition of an assertion that cannot distinguish the policy holding from the policy being absent.

## Solution

Match whole tag values. The bake file emits one quoted tag per line, so a whole-line ERE is exact and still readable:

```bash
for want in latest 9 9.9; do
  grep -qE "^ *\"aerospike/<image>:${want}\",?$" <<<"${ga}" \
    || { echo "::error::GA bake file is missing tag ${want}" >&2; exit 1; }
done
```

`${want}` is literal digits or `latest`, so no ERE escaping is needed. The negative assertion below it is fine as-is (`latest` is not a substring of any version tag), but applying the same anchoring there is harmless and consistent.

docker/Dockerfile
```diff
@@ -68,11 +68,55 @@ RUN --mount=type=bind,source=.,target=/build-ctx \
...
+# dispositions apply: `docker stop` on a tool that installs no SIGTERM handler
+# (asbench, aql, asadm) exits at once instead of waiting out the timeout and
+# being SIGKILLed. It also reaps orphans.
+#
+# The entrypoint script is the help entrypoint: it runs the binary named as the
+# first argument, and prints usage listing the binaries this image ships when
+# the caller names none or names something else. `--entrypoint bash` still gives
+# a shell. No CMD: a default binary would hide the usage message, and images
+# that ship several binaries have no correct default.
+ENTRYPOINT ["/usr/bin/tini", "--", "/usr/local/bin/docker-entrypoint.sh"]
```
pvinh-spike, line 108: From Architecture agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

The image's invocation surface is published public API, and this converts it from "anything on PATH" to a build-time allow-list while deleting `CMD ["bash"]`. The decision is fine and deliberate — I am not re-litigating it. What is missing is the enumeration of what stops working. Each row below was measured against the script with a fixture tools file:

| invocation | before | after |
|---|---|---|
| `docker run -it IMAGE` | interactive bash shell | usage on stdout, **exit 0**, no shell |
| `docker run IMAGE bash -c '…'` | worked | **exit 127** |
| `docker run IMAGE sh -c '…'` | worked | **exit 127** |
| `docker run IMAGE /usr/bin/<tool> --version` | worked | **exit 127** (`grep -qxF` matches basenames only) |
| `docker run IMAGE env FOO=1 <tool> …` | worked | **exit 127** |
| `docker run IMAGE ls /etc/aerospike` (any diagnostic: `cat`, `dpkg`, `openssl`) | worked | **exit 127** |
| K8s pod with `args:` only and no `command:` | worked | **exit 127** |

`bash`/`sh` can never be rescued by the generated list: it comes from `dpkg-query -L "${PKG_NAME}"` on the tools package, and in `debian:trixie-slim` `/usr/bin/bash` belongs to the `bash` package — `dpkg -S /bin/bash` returns no match at all. Confirmed in the pinned base. So `--entrypoint bash` is the only route to a shell, permanently.

Two corrections so the note is accurate: `docker exec -it <ctr> bash` is **unaffected** (exec bypasses ENTRYPOINT), and a K8s `command:` override is **unaffected** (it replaces ENTRYPOINT). Only `args:`-only specs break.

Where this is currently documented: Unlike the bundle repo, this repo has no `docs/` page describing the image at all, and no `CHANGELOG` or release-notes file — so the break is currently documented nowhere. That makes writing it down more important here, not less. The bundle repo (citrusleaf/aerospike-tools#451) at least gained one sentence about `--entrypoint bash`.

## Reasoning Chain

1. `ENTRYPOINT ["/usr/bin/tini","--","/usr/local/bin/docker-entrypoint.sh"]` plus no `CMD` means `docker run IMAGE` passes zero args → `$# -eq 0` → usage, exit 0. No shell. Verified.
2. With args, `exec "$@"` is gated on `is_tool "$1"` = `tools | grep -qxF -- "$1"` — exact whole-line match against the generated file, so basenames only. Any absolute path, wrapper prefix, or base-image binary falls through to stderr + `exit 127`. Ran all seven rows against a fixture.
3. Traced the list's provenance to `dpkg-query -L` filtered by `^/usr/(local/)?bin/.`, and confirmed with `dpkg -S` in the pinned base that `bash`/`dash` belong to other packages, so they are structurally unreachable.
4. Searched the repo for any changelog or release-notes convention: none exists. Read the two changed docs paragraphs — they describe the new behaviour and the escape hatch but frame nothing as removed.

## Solution

Put an explicit breaking-change block in the PR body and the TOOLS-4360 release note, and in this repo's README if it shows `docker run` examples:

```md
#### Breaking change: the image now has an entrypoint and no `CMD`

Previously `ENTRYPOINT` was empty and `CMD` was `["bash"]`, so `docker run IMAGE <anything>`
ran `<anything>`. Now the first argument must be one of the binaries the package installs;
anything else prints usage and exits 127.

| was | now | instead |
|---|---|---|
| `docker run -it IMAGE` → shell | usage, exit 0 | `docker run -it --entrypoint bash IMAGE` |
| `docker run IMAGE bash -c '…'` | exit 127 | `docker run --entrypoint bash IMAGE -c '…'` |
| `docker run IMAGE /usr/bin/asadm …` | exit 127 | `docker run IMAGE <tool> …` (name, not path) |
| `docker run IMAGE ls /etc/aerospike` | exit 127 | `docker run --entrypoint ls IMAGE /etc/aerospike` |
| pod spec with only `args: ["bash", …]` | exit 127 | set `command:` explicitly |

Unaffected: `docker exec` into a running container, and pod specs that set `command:`.
```

Optionally soften the sharpest edge by letting `is_tool` accept a path whose basename is in the list — three lines, removes one row, and does not weaken the allow-list since only already-allowed basenames would pass. The `bash`/`sh` rows should stay broken by design; they just need writing down.

.github/workflows/build-check.yml
```diff
@@ -179,3 +179,47 @@ jobs:
...
+        with:
+          fetch-depth: 1
+
+      - name: Lint the Dockerfile
+        run: |
+          set -euo pipefail
+          docker run --rm -i hadolint/hadolint:v2.14.0@sha256:27086352fd5e1907ea2b934eb1023f217c5ae087992eb59fde121dce9c9ff21e hadolint - < docker/Dockerfile
+
+      - name: Lint the docker shell scripts
+        run: shellcheck docker/docker-entrypoint.sh docker/docker-build.sh
```
pvinh-spike, line 203: From Tests agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

`docker/docker-entrypoint.sh` is 106 lines of `sh` that becomes tini's child on every container start of every published tools image, and no automated test executes it anywhere in this repo. `grep -rn "docker-entrypoint" .` returns four hits: `Dockerfile:98` (COPY), `Dockerfile:110` (ENTRYPOINT), `docker/Dockerfile` (COPY/ENTRYPOINT), and this line — which lints it without running it.

The only place its behaviour is exercised is the smoke-test step in `build-and-release.yml`, and that workflow is `workflow_dispatch`-only. So on this PR the riskiest new artifact is shellcheck-clean and nothing more. The failure mode is total: the script gates `exec` behind `is_tool`, so if that misbehaves, *nothing* runs.

It is already trivially unit-testable without building an image, because line 22 exposes `AEROSPIKE_METADATA_DIR`. A reviewer did exactly that — a two-file fixture directory plus `sh docker/docker-entrypoint.sh` reproduces the entire promise-1 contract in under a second. That the seam already exists and goes unused is what makes this hard to defend.

What the release smoke test would still miss even when it does run:
- that the exit code is exactly **0** for no-args (it only captures stdout);
- that it is exactly **127** for an unknown arg — the assertion says merely "non-zero", so a regression to 126 (lost exec bit, i.e. the entrypoint never ran) is indistinguishable from a correct refusal;
- that usage goes to **stdout** with no args and **stderr** on error — a swap would pass CI and break `docker run IMAGE > file`;
- the `[ -r "${TOOLS_FILE}" ] || return 0` fallback and the `tool_count -le 1` branch, which are dead in the bundle image but live in the five single-binary repos this script is copied into verbatim.

## Reasoning Chain

1. Grepped the tree for `docker-entrypoint`: four hits, none an execution outside a built image.
2. Read `build-check.yml` — `on: pull_request`/`push`, jobs `naming-scripts` and `docker-lint`. Neither starts a container from the image.
3. Read the smoke-test step — the only behavioural coverage — and confirmed its workflow is dispatch-only.
4. Ran the script against a fixture via `AEROSPIKE_METADATA_DIR` and got the full contract in one command, proving the test is cheap, needs no Docker, and could run in this same `docker-lint` job on `pull_request`.
5. The smoke-test step already installs bats, so there is a place for this to live.

## Solution

Add `.github/bin/test/test_entrypoint.bats` and a bats step to `docker-lint`:

```bash
setup() {
  ENTRY="${BATS_TEST_DIRNAME}/../../../docker/docker-entrypoint.sh"
  export AEROSPIKE_METADATA_DIR="${BATS_TEST_TMPDIR}/md"
  mkdir -p "${AEROSPIKE_METADATA_DIR}"
  printf 'asadm\nasinfo\n' > "${AEROSPIKE_METADATA_DIR}/tools"
  printf 'aerospike-tools image (aerospike-tools 0.0.0)\n' > "${AEROSPIKE_METADATA_DIR}/image-info"
}

@test "no args: usage on stdout, exit 0" {
  run -0 sh "$ENTRY"
  [[ "$output" == *"docker run --rm <image> asadm"* ]]
}

@test "unknown first argument: exit 127, message on stderr, nothing on stdout" {
  run --separate-stderr sh "$ENTRY" --version
  [ "$status" -eq 127 ]
  [ -z "$output" ]
  [[ "$stderr" == *"is not a tool shipped in this image"* ]]
}

@test "a listed tool is exec'd with its arguments intact" {
  printf 'faketool\n' > "${AEROSPIKE_METADATA_DIR}/tools"
  mkdir -p "${BATS_TEST_TMPDIR}/bin"
  printf '#!/bin/sh\nprintf "%%s\\n" "$@"\nexit 42\n' > "${BATS_TEST_TMPDIR}/bin/faketool"
  chmod +x "${BATS_TEST_TMPDIR}/bin/faketool"
  PATH="${BATS_TEST_TMPDIR}/bin:$PATH" run sh "$ENTRY" faketool -a "b c"
  [ "$status" -eq 42 ]
  [ "${lines[1]}" = "b c" ]
}
```

plus the single-tool lede and the missing-metadata case. Additive, runs in about a second, and turns an unexercised knob into the mechanism a PR-gating test depends on. Document `AEROSPIKE_METADATA_DIR` in the script header as the test seam while you are there.

docker/docker-build.sh
```diff
@@ -613,7 +685,9 @@ function main() {
   fi
 
   if [[ "${dry_run}" == true ]]; then
-    log_info "Dry run: skipping bake generation and Docker build."
+    log_info "Dry run: generating ${BAKE_FILE}, skipping the build."
```
pvinh-spike, line 689: From Style agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

This line changes `--dry-run` to generate and print the bake file, but the help text at `docker/docker-build.sh:90` was not updated and now says the opposite:

```
-N, --dry-run           Resolve URLs/SHAs and print, then exit (no bake, no build).
```

"no bake" is false. `-N` now writes `docker/docker-bake.hcl` and prints it — I ran `docker/docker-build.sh -p -v 9.9.9 -N -L` and the file is on disk afterwards. The filesystem side effect itself is benign (`.gitignore:75` covers it, and `git status --porcelain` was empty after), but `--help` is what a person reads before deciding whether a flag is safe to run, and it currently describes the inverse of the code in the same file.

This is not cosmetic here. The new "Assert the rolling-tag policy" step in `build-check.yml` has exactly one input: the bake file this prints. A maintainer who trusts the help text and removes the `generate_bake` call as stray debug output breaks that blocking gate — and the failure would be confusing rather than obvious, since the GA assertion would error out while the rc assertion still passes.

The `-L` help entry was added three lines above this one, so the text was being edited at the time.

## Reasoning Chain

1. Read the dry-run branch at HEAD: `generate_bake` then `cat "${BAKE_FILE}"`, replacing the previous `log_info "Dry run: skipping bake generation and Docker build."`.
2. Read line 90 at HEAD: unchanged from before the PR, still "(no bake, no build)".
3. Ran `-N` and confirmed `docker/docker-bake.hcl` exists afterwards; confirmed `.gitignore:75` covers it so the side effect does not dirty the tree.
4. Read `build-check.yml:116` — `ga=$(docker/docker-build.sh -p -v 9.9.9 -N -L)` — so the new gate depends on precisely the behaviour the help text denies.

## Solution

```
    -N, --dry-run           Resolve URLs/SHAs, write and print the bake file, then
                            exit (no build). Writes ${BAKE_FILE} in docker/.
```

If the old "nothing touches disk" semantics are wanted back for interactive use, that is a separate change — but then the tag-policy step needs another input. Do not change both.

docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,106 @@
...
+    asbackup) echo "Back up an Aerospike namespace or cluster" ;;
+    asrestore) echo "Restore an Aerospike backup" ;;
+    asbench)  echo "Benchmark an Aerospike cluster" ;;
+    asconfig) echo "Manage Aerospike configuration files" ;;
+    *)        echo "" ;;
+    esac
+}
+
+tools() {
+    [ -r "${TOOLS_FILE}" ] || return 0
```
pvinh-spike: From Control Flow agent:

Severity: important
Problem Confidence: high
Solution Confidence: medium

## Problem

`tools()` converts "the metadata file is missing or unreadable" into "the file is empty", with no diagnostic. Because `is_tool` is the sole gate in front of `exec "$@"`, an absent `${TOOLS_FILE}` makes *every* legitimate invocation fail closed — the image refuses every binary it actually ships, with a message that blames the caller.

Measured against a real built image:

1. `docker run --rm -e AEROSPIKE_METADATA_DIR=/nope IMAGE asadm`
   → `Error: 'asadm' is not a tool shipped in this image.`, then a usage block listing **no** tools, offering `docker run --rm <image> <binary> --help`. Exit 127.
2. `docker run --rm -e AEROSPIKE_METADATA_DIR=/nope IMAGE`
   → exit 0, "Name the tool to run:" followed by an empty list.

The failure is total, silent, and actively misleading: the one message the operator sees asserts the opposite of the truth.

Reachable paths: a `--tmpfs` or volume mount over `/usr/local/share`; a future refactor that moves the metadata directory; or — most relevant — any of the five other repos adopting this script verbatim without the matching Dockerfile generation. The Dockerfile's `! -s` guard protects this repo's build only, not the runtime.

## Reasoning Chain

1. `tools()` returns 0 with no output when `[ -r ]` fails. Under `set -eu` that is deliberate — it stops `set -e` killing the script — but it discards the information that the file was missing at the same moment.
2. Traced all four callers: `tool_count` → `"0"` → singular branch; `width` → `0`; `for tool in $(tools)` → zero iterations so `first` falls back to `<binary>`; `is_tool "$1"` → grep over empty input → exit 1 for *every* argument.
3. So no path reaches `exec "$@"`. The only reachable outcome for a named tool is the error block and `exit 127`.
4. Built the image and ran both cases; output and exit codes as above. Confirmed, not inferred.

## Solution

Two defensible directions — pick one deliberately.

(a) Fail loud, treating the metadata file as a build invariant. Note the check belongs once at the top of the script, not inside `tools()`: `tool_count`'s `|| true` would swallow it, and `for x in $(cat /nonexistent)` does **not** trip `set -e` under dash (verified).

```sh
if [ ! -r "${TOOLS_FILE}" ]; then
    echo "${0##*/}: ${TOOLS_FILE} is missing or unreadable; this image was built wrong" >&2
    exit 1
fi
```

(b) Fail open on the exec path only, so a real binary is never blocked:

```sh
if [ ! -r "${TOOLS_FILE}" ] || is_tool "$1"; then
    exec "$@"
fi
```

This restores the pre-PR behaviour (runc's "executable file not found") when the metadata is gone, at the cost of that one degraded case printing the runc error instead of usage.

(a) fits a build-time invariant better; (b) fits better given the script is being copied into five repos whose Dockerfiles you do not control here. Either way, the diagnostic should name the file rather than blaming the argument.

docker/Dockerfile
```diff
@@ -68,11 +68,55 @@ RUN --mount=type=bind,source=.,target=/build-ctx \
...
+      --home-dir /home/aerospike --shell /bin/bash aerospike && \
     rm -f /tmp/asadm.deb && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
 # .deb installs /usr/bin/asadm and /usr/bin/asinfo symlinks → the bundle.
 # Both are on PATH; default to asadm, swap with `docker run IMAGE asinfo …`.
-CMD ["asadm"]
+
+COPY docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
```
pvinh-spike: From Security agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

This `COPY` relies on the git index to carry the exec bit of the file that is now PID 1's child, and nothing in the repo enforces it.

Today it is correct — `git ls-files -s docker/docker-entrypoint.sh` is `100755` in all six repos (identical blob), and `COPY` preserves it: in a throwaway image the file lands `-rwxr-xr-x root root`, neither group- nor world-writable. No finding on the mode as shipped.

The gap is the absence of enforcement. If the source mode ever regresses to `100644` — a stray `chmod`, a patch applied with `git apply` losing the bit, `core.fileMode=false`, a `git archive` export, a non-POSIX checkout round-trip — then `ENTRYPOINT ["/usr/bin/tini", "--", ".../docker-entrypoint.sh"]` is exec-form, so tini `execve`s the path directly and gets EACCES with no fallback to `sh`. Every invocation of the image fails with exit 126, including `docker run IMAGE asadm`. The image is completely unusable.

And that regression merges green:

1. `docker-lint` runs hadolint and shellcheck; neither inspects file modes (confirmed — hadolint at the pinned digest exits 0 and has no mode rule).
2. The only thing that would catch it is the smoke test's `usage=$(docker run --rm "${IMAGE}")`.
3. `build-and-release.yml` is `workflow_dispatch`-only, so nothing on a PR builds the image.

So it is invisible until a release run — precisely when it is most expensive. Build-time silence, runtime totality.

## Reasoning Chain

1. Confirmed the index mode with `git ls-files -s` in all six review trees rather than trusting the diff header.
2. Built a throwaway image FROM the pinned base with the same `COPY` and inspected the result: `-rwxr-xr-x`. So `COPY` preserves the bit and the shipped permissions are correct.
3. `grep -n 'chmod' docker/Dockerfile` → no match. The mode has exactly one source of truth: the git index.
4. Ran the new blocking gate and read its scope — Dockerfile instructions only, no filesystem mode checks. Same for shellcheck.
5. Read the release workflow's trigger to confirm the one check that would catch a regression never runs pre-merge.

## Solution

```dockerfile
COPY --chmod=0755 docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
```

`--chmod` requires BuildKit, which this Dockerfile already mandates (`RUN --mount=type=bind`), so there is no compatibility cost, and it pins the mode independent of the checkout. I re-ran the pinned hadolint with it to confirm the clean result is undisturbed.

docker/docker-build.sh
a-spiker: Fixed, but **not with the suggested guard** -- it would have regressed the one case the old code got right. Flagging that explicitly since it is the substantive part of this reply.

You proposed testing `PKG_VERSION`. But `pkg_release.sh` folds `rcN` into the *iteration*, so an rc has a clean three-component `PKG_VERSION`:

| VERSION | PKG_VERSION | guard on PKG_VERSION | guard on VERSION |
| --- | --- | --- | --- |
| `9.9.9` | `9.9.9` | ALLOW | ALLOW |
| `9.9.9-rc1` | `9.9.9` | **ALLOW** | REFUSE |
| `9.9.9-rc12` | `9.9.9` | **ALLOW** | REFUSE |
| `9.9.9-beta1` | `9.9.9-beta1` | REFUSE | REFUSE |
| `9.9.9-alpha2` | `9.9.9-alpha2` | REFUSE | REFUSE |
| `13.0.3-master.abc123` | `13.0.3-master.abc123` | REFUSE | REFUSE |

Measured by running `pkg_release.sh <v> version` for each. So `PKG_VERSION` would newly let `-rc1` move `latest`.

Guarding the shape of `VERSION` gets every row right:

```bash
if [[ ! "${VERSION}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
  log_error "--latest refuses a non-GA version (${VERSION}); rolling tags are GA-only."
  exit 1
fi
```

Your diagnosis was right and I reproduced it exactly before changing anything: `-p -v 9.9.9-beta1 -N -L` emitted `latest`, `9` and `9.9`. It is now refused, along with `-alpha2` and `-master.abc123`, and `9.9.9` still passes. Your point about the latent `13` -> `13.13` case is handled too: a two-component version can no longer reach `moving_tags`.

The `-beta1` case is now in the CI gate.

.github/workflows/build-check.yml
a-spiker: Fixed, and reproduced first. Generated the bake file with `-p -v 9.9.9 -N` (no `-L`, so no rolling tags at all) and ran the three greps as written:

```
FAIL (correct)  : aerospike-asadm:latest
PASS (vacuous!) : aerospike-asadm:9
PASS (vacuous!) : aerospike-asadm:9.9
```

Exactly as you described. Now a whole-line match, tolerant of the registry prefix so it does not break if that changes:

```bash
for want in latest 9 9.9; do
  grep -qE "^ *\"[^\"]*aerospike-asadm:${want}\",?$" <<<"${ga}" \
    || { echo "::error::GA bake file is missing tag ${want}" >&2; exit 1; }
done
```

Against the same no-rolling-tags output all three now fail; against a real `-L` run all three pass.

Also added the `-beta1` refusal to the same step, so the guard gap from the `docker-build.sh` finding cannot come back, and guarded both `$(docker/docker-build.sh ...)` captures -- previously a script failure was indistinguishable from a genuine tag-policy failure, as you noted.

docker/Dockerfile
a-spiker: Fixed. Your seven-row table is now written down rather than left as trivia, and the two corrections you supplied (`docker exec` unaffected, K8s `command:` unaffected, only `args:`-only specs break) are in it.

It is in the PR body for every repo in the set, and in `docs/bundle-release-pipeline.md` for the bundle, under an explicit "Breaking change: the image now has an entrypoint and no `CMD`" heading, placed with the `docker run` examples rather than buried in the pipeline description.

I did **not** take the optional `is_tool` basename relaxation. It removes one row, but it makes the allow-list's rule ("the name of a binary this package installs") into two rules, and the row it removes is the one most likely to be a copy-paste of a path from other docs -- where the clear 127 plus usage is more useful than silently working. Happy to add it if you disagree; it is three lines.

The `bash`/`sh` rows stay broken by design, as you said, and are now documented as such.

.github/workflows/build-check.yml
a-spiker: Fixed. Added `.github/bin/test/test_entrypoint.bats` and wired it into `docker-lint`, so the entrypoint contract is now a `pull_request` gate rather than something only a dispatch would exercise.

Seven cases, all passing locally against the real script:

```
ok 1 no args: usage on stdout, exit 0
ok 2 several tools: usage says to name one
ok 3 one tool: singular lede, and that tool in the examples
ok 4 unknown first argument: exit 127, message on stderr, nothing on stdout
ok 5 a listed tool is exec'd with its arguments intact
ok 6 missing tool list: fails loudly and names the file
ok 7 missing image-info: usage still renders under the generic title
```

That covers all four gaps you listed: exact exit 0, exact exit 127, the stdout/stderr split, and both the missing-metadata path and the `tool_count -le 1` branch that is dead here but live in the single-binary repos. Case 5 also pins argument passthrough and exit-code propagation through `exec`, which is what the separate `docker stop` finding depends on.

Used your structure. Two adjustments: `bats_require_minimum_version 1.5.0` at the top, since `run` flags need it and the job installs 1.11.0 from source, and `run -127` / `run -42` / `run -1` instead of separate `status` assertions, which silences bats' BW01/BW02 warnings. Runs clean with zero warnings in all six repos.

The smoke test's own "non-zero" assertion is pinned to 127 separately -- see the Control Flow thread.

docker/docker-build.sh
a-spiker: Fixed, with your wording:

```
    -N, --dry-run           Resolve URLs/SHAs, write and print the bake file,
                            then exit (no build).
```

Keeping the new behaviour rather than restoring "nothing touches disk", precisely because the tag-policy gate's only input is this output -- so the two must not be changed independently, as you said.

docker/docker-entrypoint.sh
a-spiker: Fixed, taking **(a) fail loud**.

```sh
if [ ! -r "${TOOLS_FILE}" ]; then
    echo "${0##*/}: ${TOOLS_FILE} is missing or unreadable; this image was built wrong" >&2
    exit 1
fi
```

At the top of the script, not inside `tools()`, for the reason you gave: `tool_count`'s `|| true` would swallow it. `tools()` is now just `cat "${TOOLS_FILE}"`, since the invariant is enforced above it.

Chose (a) over (b) deliberately. (b) restores the pre-PR runc error, but it does it by making a missing metadata file silently disable the allow-list -- so the one situation where we know the image is broken is also the one where it stops checking anything. A build invariant that announces itself is easier to act on. The five single-binary repos get the identical script, so they are covered by the same check rather than depending on each Dockerfile.

The new bats case pins it: exit 1, and the message names the file rather than the argument.

docker/Dockerfile
a-spiker: Fixed:

```dockerfile
COPY --chmod=0755 docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
```

Agreed on the reasoning -- the mode is correct today, but its only source of truth is the git index, and the failure mode is a completely unusable image that no pre-merge check would catch. Re-ran the pinned hadolint on all six Dockerfiles after the change: clean.

a-spiker:
### Status of this round

All 7 findings on this PR are addressed and pushed, in three commits. The bundle PR carries 15; the extra ones that also applied here (the `useradd --system` warning, the duplicated package name in `image-info`, the unguarded `usage=$(...)` capture, the loose non-zero assertion) are fixed here too, so the six images stay byte-identical where they were.

One finding was fixed **differently from the suggestion**, and it matters: the `--latest` guard. Testing `PKG_VERSION` would have newly accepted `-rc1`, because `pkg_release.sh` folds `rcN` into the iteration and leaves `PKG_VERSION` a clean `X.Y.Z`. Guarding the shape of `VERSION` handles rc, beta, alpha and dev suffixes alike. Full table in that thread.

Three findings are **deliberately deferred**, each with its reasoning in-thread: the `mode=max` provenance URL, the `docker stop` behavioural assertion, and the per-arch `IMAGE_CREATED`. All three share one blocker.

### On the Evidence gate

A `release=false` dispatch cannot reach `docker-build-arch` on any branch right now, for a reason unrelated to this PR.

Debian 11 went LTS end-of-life on 2026-08-31 and `bullseye-security` expired on 2026-09-07, so every `debian11` job now dies at *Install prereqs*, before checkout:

```
E: Release file for http://deb.debian.org/debian-security/dists/bullseye-security/InRelease is expired
/__w/_temp/....sh: line 12: git: command not found
##[error]Process completed with exit code 127.
```

The chain is:

```
build-artifacts -> organize-built-artifacts -> notarize-mac-artifacts
  -> sign-artifacts -> test-install-linux-and-execute
  -> organize-signed-artifacts -> docker-build-arch
```

`debian11` sits in **both** matrix jobs, so the dispatch fails long before the image is built. TOOLS-3137 removes it: aerospike/aerospike-admin#555. Once that lands, a dispatch becomes possible and the three deferred findings go in with it.

### What was verified locally in the meantime

- The `-beta1` gap, reproduced before the fix (`latest`, `9`, `9.9` emitted) and refused after; `9.9.9` still accepted.
- The vacuous greps, reproduced against a no-rolling-tags bake file (2 of 3 passed), and failing correctly with the anchored match.
- `test_entrypoint.bats`: 7/7 passing, zero bats warnings, in all six repos.
- `shellcheck` clean on both docker scripts, pinned `hadolint` clean on the Dockerfile, both workflows parse, in all six repos.
- `useradd` with and without `--system` on `debian:trixie-slim`: identical account, warning gone.

What none of that covers is the smoke test running in the real job against the real image, which is what the dispatch is for.

a-spiker:
Rescoped this PR to TOOLS-4360 only and restarted the branch from `master` (previous head: d20b5097). The allow-list entrypoint is replaced by a passthrough one (`exec "$@"`), so `docker run IMAGE bash` and every other invocation that worked before still works; only "no arguments", "leading flag" and "unknown name" change, and all three now print usage. Where each review finding landed:

1. `--latest` guard only tests `-rc` (docker-build.sh): dropped with the rolling-tags feature; tags are the shared workflow's since TOOLS-4364, rolling tags belong to TOOLS-4177.
2. Two vacuous tag-policy greps (build-check.yml): dropped with the same feature; the `docker-lint` job is gone.
3. Enumerate what stops working (Dockerfile): kept. The table is in the PR body; with passthrough it has three changed rows, not seven.
4. No automated test executes the entrypoint (build-check.yml): kept. `test_entrypoint.bats` runs on every pull request in the `Naming scripts` job; the release smoke test pins exit 0 / 127 and the stdout / stderr split against the built image.
5. `--dry-run` help text contradiction (docker-build.sh): dropped with the feature; that change is not in this PR.
6. Missing tools file fails closed silently (docker-entrypoint.sh): moot. exec no longer consults the list; a missing list only degrades the usage text, and a bats case asserts exec still works without it.
7. `COPY` relies on the git index exec bit (Dockerfile): kept, `COPY --chmod=0755`.

From the follow-up round: the unguarded `usage=$(...)` capture and the non-zero-only assertion are fixed (guarded capture, 127 exactly, stdout asserted empty, stderr asserted to name the argument and carry usage); the duplicated package name in `image-info` is fixed and the smoke test asserts the title format. The "stacked on TOOLS-4361" claim is moot: 4361 merged and this branch is rebuilt on top of it.

docker/Dockerfile
```diff
@@ -86,10 +94,32 @@ RUN --mount=type=bind,source=.,target=/build-ctx \
...
+# The .deb installs /usr/bin/asadm and /usr/bin/asinfo symlinks → the bundle;
+# both are on PATH.
+# Convention: `docker run --rm IMAGE <binary> [args]`.
+# The entrypoint execs whatever is named (a passthrough, so `IMAGE bash` still
+# works) and prints usage built from /usr/local/share/aerospike/tools when
+# nothing, or an option, is named. No CMD: a default binary would hide that
+# usage and a multi-binary image has no right default. Exec form so the script
+# -- and, after exec, the tool -- is PID 1 and gets `docker stop`'s SIGTERM.
+COPY --chmod=0755 ${ENTRYPOINT_SRC} /usr/local/bin/docker-entrypoint.sh
+ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]
```
pvinh-spike, line 126: From Architecture agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

This is the only image in the set whose old default (`CMD ["asadm"]`) was an interactive, stdin-consuming tool, and the PR body's compat table understates the one break a caller cannot see. `echo 'summary' | docker run -i IMAGE` (or `docker run -i IMAGE < commands.txt`) previously executed those commands through asadm; it now prints usage and exits 0 — a success code with nothing executed. The table's row 1 ("ran `asadm` with no arguments" → usage, exit 0) is accurate but describes only the interactive no-arg case.

## Reasoning Chain

The flag/unknown-tool rows fail loudly with 127 — a caller notices. The piped-stdin case is the opposite: the pipeline keeps returning 0 while executing nothing, and its stdout (usage text) even resembles output. A cron job or wrapper doing `docker run -i <image> <<EOF` breaks silently on the next image pull. Nothing in-tree relies on it (verified — the e2e code runs the server image, not this one), but external users of the published asadm image are exactly who release notes exist for. The entrypoint design is right; the description should name the class so the break is searchable when someone hits it.

## Solution

No code change. Add one row (or a note) to the PR body table and carry it into the image's release notes: "`echo '…' | docker run -i IMAGE` previously fed asadm's prompt; it now prints usage and exits 0 — add `asadm` as the first argument."

docker/docker-build.sh
pvinh-spike: Resolved at bf71c2f0 — the `--latest` rolling-tag change containing this guard was rescoped out when the branch restarted from the default branch; tagging is owned by shared-workflows, and rolling tags belong to TOOLS-4177. The flagged code is no longer part of this PR.

.github/workflows/build-check.yml
pvinh-spike: Resolved at bf71c2f0 — the tag-policy assertions were removed together with the rolling-tag change in the rescope; the concern no longer applies to this PR.

docker/Dockerfile
pvinh-spike: Resolved at bf71c2f0 — the allow-list was replaced by a passthrough (`exec "$@"`), and the invocation-surface table is in the PR body with this thread's corrections. Residual doc gaps for the piped-stdin class are raised fresh this round (inline here or on the coordinated aql/asadm PRs).

.github/workflows/build-check.yml
pvinh-spike: Resolved at bf71c2f0 — .github/bin/test/test_entrypoint.bats (12 cases) now gates every pull request via build-check.yml, and the release smoke test asserts the same contract against the built image on both arches.

docker/docker-build.sh
pvinh-spike: Resolved at bf71c2f0 — the `--dry-run` behavior change was rescoped out; docker-build.sh's help text matches its behavior again.

docker/docker-entrypoint.sh
pvinh-spike: Resolved at bf71c2f0 — redesigned rather than the fail-loud fix from the previous round: the guard moved to build time (the Dockerfile RUN fails the image build if the generated tools list is empty), and the runtime deliberately tolerates a missing file (generic usage, exec unaffected), pinned by the 'missing tool list' bats case. The silent-misbuild concern is now caught before an image exists.

docker/Dockerfile
pvinh-spike: Resolved at bf71c2f0 — `COPY --chmod=0755 ${ENTRYPOINT_SRC} /usr/local/bin/docker-entrypoint.sh`; the mode no longer depends on the git index.

a-spiker:
Follow-up pushed as 4be0faa, applied identically in all six repos (same generator; shared files still byte-identical). Where the re-review findings landed:

- Piped stdin to the old default: added as its own row in the compat table above, with the per-image "before" behaviour, and flagged as the release-note line. No code change; exit 0 on no arguments is the set-wide decision.
- Self-refuting "multi-binary image has no right default" comment: reworded in all six to the real rationale, one convention across the tools images (TOOLS-4360) even where one binary ships.
- Refusal contract wording: bats header and the bundle docs now say "resolves to nothing"; a new bats case pins the exists-but-not-executable path falling through to exec with 126 and no usage (skipped where `sh` is bash, so it runs on the dash runner). Code unchanged, as recommended.
- Bats step in the naming job: moved to its own job, "Docker entrypoint (bats)", with the file's existing harden-runner and checkout pins; the naming job is back to what it was on the default branch (the bundle's added tarball install is gone with it).
- Guard message: now "under /usr/bin or /usr/local/bin", matching the grep.
- "never hand-maintained" vs the OCI description: the comment no longer claims that; it says the usage text follows the package. The description label is left as is, since changing registry-facing metadata is a separate decision.
- Install bats via tarball under sudo: the new job uses the distro package (`apt-get install bats`, 1.10 on noble, floor is 1.5). The pre-existing tarball install in the naming job and in build-and-release.yml is untouched.

docker/Dockerfile
pvinh-spike: Resolved by 4be0faa3 — the compat table gained the piped-stdin row with this image's 'before' behaviour (commands fed to asadm's prompt) and the rewrite (`docker run -i IMAGE asadm`), flagged as the release-note line. No code change, matching the recommendation.

pvinh-spike (approved):
LGTM

PR #552, opened 2026-09-04
feat(ci): TOOLS-4364 per-leg build-info and SLSA provenance for packages and images

pvinh-spike:
## Problem

Two gaps in what a release records about itself:

1. **No per-leg build-info.** The release build-info named the artifacts but not the environment or git revision each package was built from. Twenty build legs (ten distros × two arches) each produce a package in its own container, and none of that context survived into Artifactory.
2. **No provenance.** Neither the packages nor the docker images carried a SLSA attestation, so there was nothing for a consumer to verify a downloaded `.deb`/`.rpm`/`.pkg` or a promoted image against.

## Change

`.github/workflows/build-and-release.yml` only.

**Per-leg build-info** (`build-artifacts`, gated on `inputs.release`)

- Install the JFrog CLI in each leg and publish build-info under `<VERSION>-metadata-<distro>-<arch>`. `upload-artifacts-to-jfrog` already globs that `jf-metadata-build-id` prefix, so the legs land as child builds of the release build-info with no change on the aggregation side.
- Every step is best-effort — `continue-on-error` on the install, `|| echo "::warning::"` on each `jf` call, and a `command -v jf` guard because the install is allowed to fail. This is deliberately unlike the hard-failing publish in shared-workflows' `execute-build` entrypoint: metadata about a package that built, signed and tested clean is not worth failing a release over. A leg that fails here still ships; it just goes missing from the aggregated build-info and says so.

**Package provenance** (new `attest-artifacts` job)

- Runs after `organize-signed-artifacts`, not after the build. Signing rewrites the deb/rpm in place, so an attestation of the unsigned digest would never verify against the file a customer downloads.
- Stages only installable packages (`.deb .rpm .tgz .tar .pkg .exe`); detached signatures and checksums describe a package rather than being build outputs of their own. Hard-fails if the staging set comes out empty.
- `create-release-bundle` now `needs: attest-artifacts`, so an unattested package cannot reach the promotion unit.

**Image provenance** (`docker-manifest`) — ⚠️ superseded by the shared-workflow move below; `docker-manifest` no longer exists

- The multi-arch index and both per-arch images are all consumable refs, so all three get provenance. Digests are captured as step outputs by the build-info step that already resolves them, rather than inspecting the registry a second time.
- Recorded with GitHub rather than pushed as OCI referrers: `gh attestation verify oci://` resolves them either way, and push-to-registry would make a release depend on Artifactory's referrers support.
- `subject-name` is the `dev-local` path the images are pushed to. Promotion moves images without rebuilding, so a promoted image still verifies on the digest recorded here.

## Verification

Full release run on this branch: [33811267997](https://github.com/aerospike/aerospike-admin/actions/runs/33811267997) — green end to end, including `upload-artifacts-to-jfrog`, `docker-manifest` and `create-release-bundle`.

**Build-info:** 23 build-infos deployed — all 20 `…-metadata-<distro>-<arch>` legs, plus `-artifacts`, the docker build-info, and the aggregated release build-info. Zero `::warning::` from the new steps in any leg, so no leg fell back to best-effort.

**Package attestation:** `Attestation created for 22 subjects` — 12 deb + 8 rpm + 2 macOS pkg, i.e. every installable package the release ships, attested post-signing.

**Image attestation:** three distinct digests attested against `artifact.aerospike.io/database-docker-dev-local/aerospike-asadm`:

| subject | digest |
|---|---|
| multi-arch index | `sha256:ddf7a4db…` |
| amd64 | `sha256:e07c29b3…` |
| arm64 | `sha256:8391c84b…` |

**Non-release runs** are unaffected: every new step and the `attest-artifacts` job are gated on `inputs.release == true`.

## Follow-up: `setup-jfrog-cli` v4.9.1 → v5.1.0

Adding the action to twenty build legs made an existing deprecation loud: v4.9.1 declares `node20`, which the runners now force onto Node 24 and warn about once per job. v5 declares `node24` outright.

Bumped at all five call sites (four in this workflow, one in `tests.yml`) so the pin does not fragment. Every input we pass — `version`, `oidc-provider-name`, `oidc-audience` — is unchanged across the major bump; v5 only adds `enable-package-alias` / `package-alias-tools`, both defaulting off. The action's default CLI version moved 2.88.0 → 2.91.0, which does not reach us because all five sites pin `2.78.8`.

Run 33811267997 already executed this action on Node 24 in all ten distro containers on both arches (that is what the warning says it did), so declaring `node24` is behaviourally a no-op — it just stops relying on a shim that will eventually be removed.

## Follow-up: build images via `reusable_docker-build-deploy`

`docker-build-arch` and `docker-manifest` are replaced by a call to shared-workflows' `reusable_docker-build-deploy.yaml@v4.1.0`, plus a new `docker-test` job. That workflow already does everything those two jobs did — native per-platform legs, push by digest, one stitched manifest list, docker build-info, provenance attestation — so this removes ~210 lines and takes CI off `docker/docker-build.sh` (which stays as the local and manual path).

**This supersedes the docker half of a24ff4a.** The three per-arch attestation steps that commit added are deleted here; the shared workflow attests the manifest index under an `oci://` subject instead. The package attestation in `attest-artifacts` is untouched.

### Pin

v4.1.0, not the v3.7.0 used elsewhere in this file. `gh-context-artifacts-json` — the input that places the signed `.deb` inside the build context — does not exist before v4.1.0. Without it the image could only install a package fetched from a URL, never the artifact the run just built and signed. Refs are per-`uses:`, so the other calls stay on v3.7.0.

### Behaviour changes

| | before | after |
|---|---|---|
| moving tag | `<RELEASE>` | `<RELEASE>` (via `versions-override`) |
| immutable tag | — | `<BUILD_VERSION>_<timestamp>` (always added, cannot be disabled) |
| per-arch tags | `<RELEASE>-amd64` / `-arm64` | **gone** — legs push by digest |
| `latest` / bare major | never published | still never published (suppressed by the override) |
| image attestation | index + both per-arch manifests | index only, `oci://` subject |
| smoke test | before push, on every run | after publish, release runs only |
| build context | `docker/` (28K) | repo root + `.dockerignore` (446B) |

`create-release-bundle` now asks for the image build-info by `github.run_number`: the shared workflow records it under the JFrog CLI default build number, not the `VERSION` string `upload-artifacts-to-jfrog` uses for the packages. Leaving it as `VERSION` would have silently failed to resolve.

`create-release-bundle` also `needs: docker-test`, so test-before-promote survives the move to test-after-publish — a failing image stays in `dev-local` and is never promoted to TEST.

### Verification

The run's artifacts had expired (1-day retention), so the `.deb` was reproduced synthetically — named exactly as `pkg/Makefile` emits — and the real `docker/Dockerfile` was built with the exact context and build-args the workflow passes:

- amd64 leg installed the amd64 package, arm64 leg the arm64 one — the `dpkg --print-architecture` branch and both `ASADM_LOCAL_PKG_*` args resolve correctly
- build log shows `Get:1 /tmp/asadm.deb`, i.e. the in-context package rather than a URL download — the reason v4.1.0 is required
- context transfer measured at **446B** with `.dockerignore` against **15.94MB** without

`actionlint` is clean on the workflow apart from one pre-existing `SC2086` at line 365, untouched by this PR.

### Known risk

Non-release runs pass `push: false`. The shared workflow documents dry-run as supported, but its own test suite never exercises that path, and each leg uploads its digest with `if-no-files-found: error`. If buildx emits no digest without a push, non-release runs will fail there. If that happens, the fix is to gate the whole call on `inputs.release`.


pvinh-spike:
**Review response — 4 findings, all verified as accurate. 3 fixed in 01d4adb, 1 accepted as a known gap.**

| # | Finding | Status |
|---|---|---|
| 1 | `app-version` reintroduces rcN into the immutable image tag | **Fixed** |
| 2 | `docker-test` gated on `release==true`, no dry-run image validation | **Accepted** — deliberate; doc corrected |
| 3 | Missing in-context `.deb` falls through to `curl ""` | **Fixed** |
| 4 | No `ASADM_*_SHA256`, so the digest check is skipped | **Accepted** — see below |

### 1 — rcN in the image tag (fixed)

Confirmed and the most serious of the four. On the master release path `get-version` sets `BUILD_VERSION="${BASE}"` — the raw `VERSION` file value, rcN included — and `app-version` feeds the immutable `<app-version>_<timestamp>` tag the shared workflow always emits. A release would have published `aerospike-asadm:5.0.3-rc3_<ts>`.

That contradicts an invariant stated in two places: this file's header ("rcN lands in the deb revision / rpm Release and never in a file name or image tag") and `pkg_release.sh` ("must not leak into package file names or image tags"). `build-check.yml` enforces it for package names only, so nothing would have caught it.

Worth noting why it survived local testing: on a *branch* build `BUILD_VERSION` is `pkg_release.sh … version` + sha, which strips rcN. The bug is only reachable on the master release path.

Now passes `RELEASE`, the same value `versions-override` already used, so the moving and immutable tags are both `<RELEASE>`-based.

### 2 — no image validation on dry runs (accepted, doc fixed)

Accurate. This was a deliberate tradeoff taken when the smoke test moved after publish: the shared workflow pushes by digest and exposes no test hook, so with `push=false` there is no image to exercise. The alternative considered was keeping a second, local build-and-test path for non-release runs; it was rejected as two build paths to maintain.

The reviewer's stronger point is that the header still advertised dry runs as running "end-to-end", which now overstates. Corrected to say the image is built for both platforms but not pushed and therefore not smoke-tested.

Happy to reopen the coverage question if the gap matters more than the duplicate build path.

### 3 — silent fall-through to `curl ""` (fixed)

Confirmed. `[ -n "$LOCAL_PKG" ] && [ -f … ]` treated "named but absent" as a cue to download, and with no URL args in CI that is `curl -fsSL "" -o /tmp/asadm.deb` — an opaque curl usage error instead of the name that was wrong.

Now names the missing file, lists what the context actually holds, and exits 1 — restoring the diagnostic `docker-build.sh:423-427` gave before CI stopped calling it. Verified both directions: a misnamed package fails with that message and no curl fallback, a correctly named one still installs.

The `.dockerignore` comment describing a COPY failure mode this Dockerfile does not have was also corrected.

### 4 — checksum no longer verified (accepted)

Accurate: no `ASADM_*_SHA256` is passed, so `PKG_SHA` is empty and the `sha256sum -c` at `docker/Dockerfile:74-76` is skipped. `docker-build.sh` computed it from the local file.

Not restored, because the cheap route does not exist. `deploy-artifacts` ships 22 `.asc` detached signatures and **no** `.sha256` files, so there is nothing in the context to verify against. Computing it would need a prep job that downloads the entire artifact set — every `.deb`, `.rpm` and `.pkg` — to derive two checksums, on every release run.

What still covers the realistic failure modes: a wrong or missing name now hard-fails (finding 3); GitHub artifact upload/download is itself integrity-checked; and `apt-get install` rejects a structurally corrupt `.deb`. What is genuinely lost is end-to-end verification that the bytes installed are the bytes that were signed.

If that gap matters, the better fix is verifying the `.asc` signature already present in the context rather than reintroducing a sha256 — that would need the Aerospike public key available at build time. Say the word and I'll open a follow-up.

a-spiker (approved):
LGTM

PR #553, opened 2026-09-05
feat: copy aerospike-asadm images to Docker Hub from VERSION

mcoberly2:
## Summary
- Add a dispatchable workflow that copies already-published `aerospike-asadm` tags from JFrog to Docker Hub (same pattern as [aerospike-tools](https://github.com/citrusleaf/aerospike-tools/blob/master/.github/workflows/copy-tools-images-to-dockerhub.yml)). Nothing is built or pushed back to Artifactory.
- Tags are derived from `VERSION` (e.g. `5.0.3-rc5` → image tag `5.0.3-5`, Hub aliases `5.0.3` and `latest`). Derived tags require git tag `VERSION` so an unreleased bump cannot become `:latest`. Non-rc pre-releases cannot default aliases to `<semver>,latest`. `alias-tags=none` copies tags only.
- Dry-run is on by default. Explicit tag inputs override the derived defaults.

## Test plan
- [ ] After merge (workflow_dispatch is not registered until the file is on the default branch): run **Copy aerospike-asadm images to Docker Hub** with dry-run checked and empty tag inputs; confirm it resolves `VERSION` `5.0.3-rc5` to `5.0.3-5` / `5.0.3-5-amd64` / `5.0.3-5-arm64` and aliases `5.0.3` and `latest` from `artifact.aerospike.io/docker/aerospike-asadm`.
- [ ] Run again with dry-run unchecked and confirm Docker Hub has `aerospike/aerospike-asadm:latest` (and `:5.0.3`) pointing at the same digest as `:5.0.3-5`.

### DIFF
--- a/.github/workflows/copy-tools-images-to-dockerhub.yml
+++ b/.github/workflows/copy-tools-images-to-dockerhub.yml
@@ -1,0 +1,161 @@
+name: Copy aerospike-asadm images to Docker Hub
+
+# Copies already-published aerospike-asadm tags from JFrog to Docker Hub.
+# Does not build images and does not write anything back to Artifactory.
+# Tags: 5.0.3-5 (multi-arch index), 5.0.3-5-amd64 (linux/amd64),
+# 5.0.3-5-arm64 (linux/arm64).
+
+on:
+  workflow_dispatch:
+    inputs:
+      source-image:
+        description: Source image without tag
+        required: true
+        default: artifact.aerospike.io/docker/aerospike-asadm
+      dest-image:
+        description: Destination Docker Hub image without tag
+        required: true
+        default: aerospike/aerospike-asadm
+      tags:
+        description: Comma-separated tags to copy unchanged (index, linux/amd64, linux/arm64)
+        required: true
+        default: 5.0.3-5,5.0.3-5-amd64,5.0.3-5-arm64
+      dry-run:
+        description: Inspect JFrog tags only; do not push to Docker Hub
+        type: boolean
+        required: false
+        default: true
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  copy:
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
+        with:
+          egress-policy: audit
+
+      - name: Install JFrog CLI
+        id: jf
+        uses: step-security/setup-jfrog-cli@893a92f96c7727242adf0e603c77b1f2a22390e1 # v4.9.1
+        env:
+          JF_URL: https://artifact.aerospike.io
+          JF_PROJECT: database
+        with:
+          version: 2.78.8
+          oidc-provider-name: database-gh-aerospike
+          oidc-audience: database-gh-aerospike
+
+      - name: Docker login to Artifactory (pull only)
+        uses: docker/login-action@b45d80f862d83dbcd57f89517bcf500b2ab88fb2 # v4.0.0
+        with:
+          registry: artifact.aerospike.io
+          username: ${{ steps.jf.outputs.oidc-user }}
+          password: ${{ steps.jf.outputs.oidc-token }}
+
+      - name: Docker login to Docker Hub
+        if: ${{ !inputs.dry-run }}
+        uses: docker/login-action@b45d80f862d83dbcd57f89517bcf500b2ab88fb2 # v4.0.0
+        with:
+          username: ${{ vars.AEROSPIKE_DOCKER_ORG_NAME || 'aerospike' }}
+          password: ${{ secrets.AEROSPIKE_DOCKER_ORG_RW_TOKEN }}
+
+      - uses: step-security/setup-buildx-action@f931205d68723ad9589fd2a7e2ece238bf9de341 # v4.0.0
+
+      - name: Copy existing tags to Docker Hub
+        id: copy
+        shell: bash
+        env:
+          SOURCE_IMAGE: ${{ inputs.source-image }}
+          DEST_IMAGE: ${{ inputs.dest-image }}
+          TAGS: ${{ inputs.tags }}
+          DRY_RUN: ${{ inputs.dry-run }}
+        run: |
+          set -euo pipefail
+
+          # Registry-to-registry copy of already-published tags.
+          # Do not docker build, do not docker push to Artifactory.
+          copy_tag() {
+            local src="$1"
+            local dest="$2"
+            docker buildx imagetools create --tag "$dest" "$src"
+          }
+
+          inspect_digest() {
+            docker buildx imagetools inspect "$1" --format '{{json .Manifest}}' | jq -r '.digest'
+          }
+
+          tags_array=()
+          IFS=',' read -ra raw_tags <<< "$TAGS"
+          for tag in "${raw_tags[@]}"; do
+            tag="${tag#"${tag%%[![:space:]]*}"}"
+            tag="${tag%"${tag##*[![:space:]]}"}"
+            [ -z "$tag" ] && continue
+            tags_array+=("$tag")
+          done
+
+          if [ "${#tags_array[@]}" -eq 0 ]; then
+            echo "::error::no tags parsed from input '$TAGS'"
+            exit 1
+          fi
+
+          copied=()
+          source_digests=()
+
+          {
+            echo "## Docker copy"
+            echo ""
+            echo "- Source: \`$SOURCE_IMAGE\`"
+            echo "- Destination: \`$DEST_IMAGE\`"
+            echo "- Dry run: \`$DRY_RUN\`"
+            echo ""
+            echo "| Tag | Source digest | Destination digest |"
+            echo "| --- | --- | --- |"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+          for tag in "${tags_array[@]}"; do
+            src="${SOURCE_IMAGE}:${tag}"
+            dest="${DEST_IMAGE}:${tag}"
+
+            echo "::group::Inspect source ${src}"
+            docker buildx imagetools inspect "$src"
+            src_digest="$(inspect_digest "$src")"
+            if [ -z "$src_digest" ] || [ "$src_digest" = "null" ]; then
+              echo "::error::could not read digest for $src"
+              exit 1
+            fi
+            echo "source digest: $src_digest"
+            echo "::endgroup::"
+            source_digests+=("${tag}@${src_digest}")
+
+            dest_digest=""
+            if [ "$DRY_RUN" = "true" ]; then
+              echo "dry-run: skipping copy of $src -> $dest"
+              dest_digest="(skipped)"
+            else
+              echo "::group::Copy ${src} -> ${dest}"
+              copy_tag "$src" "$dest"
+              dest_digest="$(inspect_digest "$dest")"
+              echo "destination digest: $dest_digest"
+              echo "::endgroup::"
+
+              if [ "$src_digest" != "$dest_digest" ]; then
+                echo "::error::digest mismatch for tag ${tag}: source=${src_digest} dest=${dest_digest}"
+                exit 1
+              fi
+              copied+=("${dest}@${dest_digest}")
+            fi
+
+            echo "| \`${tag}\` | \`${src_digest}\` | \`${dest_digest}\` |" >> "$GITHUB_STEP_SUMMARY"
+          done
+
+          IFS=','
+          echo "copied-tags=${copied[*]}" >> "$GITHUB_OUTPUT"
+          echo "source-digests=${source_digests[*]}" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/copy-tools-images-to-dockerhub.yml
+++ b/.github/workflows/copy-tools-images-to-dockerhub.yml
@@ -3,7 +3,8 @@ name: Copy aerospike-asadm images to Docker Hub
 # Copies already-published aerospike-asadm tags from JFrog to Docker Hub.
 # Does not build images and does not write anything back to Artifactory.
 # Tags: 5.0.3-5 (multi-arch index), 5.0.3-5-amd64 (linux/amd64),
-# 5.0.3-5-arm64 (linux/arm64).
+# 5.0.3-5-arm64 (linux/arm64). Extra dest tags 5.0.3 and latest
+# are aliases of 5.0.3-5.
 
 on:
   workflow_dispatch:
@@ -25,6 +26,14 @@ on:
         type: boolean
         required: false
         default: true
+      alias-from:
+        description: Source tag to copy onto extra destination tags
+        required: false
+        default: 5.0.3-5
+      alias-tags:
+        description: Extra destination tags (comma-separated) based on alias-from
+        required: false
+        default: 5.0.3,latest
 
 permissions:
   contents: read
@@ -77,6 +86,8 @@ jobs:
           DEST_IMAGE: ${{ inputs.dest-image }}
           TAGS: ${{ inputs.tags }}
           DRY_RUN: ${{ inputs.dry-run }}
+          ALIAS_FROM: ${{ inputs.alias-from }}
+          ALIAS_TAGS: ${{ inputs.alias-tags }}
         run: |
           set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


mcoberly2:
This repo’s default branch requires 1 approving review (`enforce_admins` is on), so I cannot squash-merge this the way we did for aql/asconfig/asbench/asbackup.

Please approve when you can so the copy workflow lives in this repo. Meanwhile I will copy `aerospike-asadm:5.0.3-5` (plus amd64/arm64) to Docker Hub using the same dispatcher already on `aerospike/aerospike-tools-backup`.

pvinh-spike:

**Promises:**
1. Dispatch-only workflow copying already-published asadm tags JFrog→Docker Hub; builds nothing, pushes nothing back to Artifactory.
2. Default tags derive from VERSION (rcN folded via pkg_release.sh: 5.0.3-rc5 → 5.0.3-5); aliases `<semver>` + `latest` from the release tag.
3. Explicit tags/alias inputs override the derived defaults.
4. Dry-run (default on) inspects JFrog only.
5. Follows the aerospike-tools copy-workflow pattern; divergences deliberate and sound.
6. Every copied tag is digest-verified.

| Gate | Status |
|------|--------|
| Contract | fail — promise 2's derived defaults can alias unreleased or non-rc pre-release builds to `latest` (2 findings), and promise 3 has no expressible "no aliases" value |
| Impacted surface | pass — new file only; publish-side tag computation cross-checked (same pkg_release.sh derivation); filename and `docker/` source-path match the merged siblings |
| Failure paths | pass — every failure path traced and empirically executed under set -euo pipefail: pkg_release.sh rejection kills the step via assignment-substitution, inspect failures propagate through pipefail-inherited command substitution, the jq null/empty guard is live, dry-run is read-only end to end |
| Evidence | fail — no automated check exercises this file (no actionlint/shellcheck in CI, dispatch-only), and both manual test-plan boxes are unchecked; prior PRs in this series linked green dry runs |
| Quality | pass — one structural note on inline shell vs the repo's .github/bin + bats convention |
| Backward Compat | pass — no collisions; nothing pre-existing touched |

**6 findings** posted inline (0 Critical, 2 Important, 4 Minor).

Verified sound along the way: all five action pins match their version comments (checked against tags); no `${{ }}` interpolation reaches any run: body (everything flows through env); the Hub secret is never mounted in dry-run; hyphenated GITHUB_OUTPUT keys are all referenced correctly; comma-only and whitespace inputs behave; empty `copied[*]` under set -u is safe on the runner's bash; the digest-mismatch check is a live tripwire, not dead code; the OIDC provider name matches this repo's other workflows.

Minimum to a clean pass: dispatch the dry run on this branch and link the green run (closes Evidence and settles whether `artifact.aerospike.io/docker/aerospike-asadm` — a path appearing nowhere else in this repo, which publishes to `database-docker-dev-local` — actually serves the asadm tags), and decide on a guard for the derived `latest` alias.

.github/workflows/copy-tools-images-to-dockerhub.yml
```diff
@@ -0,0 +1,304 @@
...
+          fi
+
+          alias_from="$(trim "${ALIAS_FROM_INPUT:-}")"
+          if [[ -z "$alias_from" ]]; then
+            alias_from="$RELEASE"
+          fi
+
+          alias_tags="$(trim "${ALIAS_TAGS_INPUT:-}")"
+          if [[ -z "$alias_tags" ]]; then
+            alias_tags="${SEMVER},latest"
```
pvinh-spike: From Architecture agent:

Severity: important
Problem Confidence: high
Solution Confidence: medium

## Problem

The derived defaults assume VERSION on the dispatched ref names a released iteration, but this repo's own release flow guarantees windows where it does not: Bump Version merges the next `-rcN` to master *before* that rc is built or blessed. Trigger sequence:

1. 5.0.3 ships; Bump Version merges VERSION=5.0.4-rc1 to master.
2. build-and-release publishes the rc1 images to JFrog as `5.0.4-1` — pkg_release.sh folds rcN into the iteration, so the JFrog tag already looks final — before anyone has blessed 5.0.4.
3. An operator dispatches this workflow on master (the dispatch UI's default ref), dry-run unchecked, tag inputs empty.
4. The resolve step derives tags=`5.0.4-1,…`, alias-tags=`5.0.4,latest`; the copy publishes the unblessed rc1 build as public `aerospike/aerospike-asadm:latest` and `:5.0.4`.

The reference workflow avoided this by hardcoding tags at merge time — a human asserted the released tag in a reviewed PR. This PR replaces that assertion with "whatever VERSION says right now", which is precisely the novel part, and nothing checks it. (Two agents converged on this independently.)

## Reasoning Chain

1. Publish-side tags are pkg_release.sh(VERSION, release) — the derivation matches the publish computation, so there is no drift; the gap is purely *which* VERSION.
2. VERSION moves ahead of the last release the moment the next bump merges; the derive is correct only in the window between "release dispatched" and "next bump merged".
3. The repo already has the invariant that closes this: `create-tag` is tag-last — `git tag <VERSION>` exists only after every publish step succeeded (build-and-release.yml, the create-tag job's needs list, line ~1066).

## Solution

When falling back to derived defaults, require the release to be provable: checkout with `fetch-tags: true` and refuse the derive unless `git rev-parse -q --verify "refs/tags/${VERSION}"` succeeds, with an error saying "VERSION ${VERSION} has no release tag — pass explicit tags or dispatch on the release tag". Explicit inputs bypass the guard (the operator is asserting the tags). Alternatives if the guard feels heavy: gate the `latest` alias behind an explicit boolean input defaulting false, or point the default source at a promoted-only JFrog path so an unblessed iteration cannot resolve. Whichever way, the dispatch-on-the-release-tag runbook becomes self-verifying instead of trusted.

.github/workflows/copy-tools-images-to-dockerhub.yml
```diff
@@ -0,0 +1,304 @@
...
+# Leave tags / alias-from / alias-tags empty at dispatch to use those
+# defaults, or set them to copy a different already-published set.
+
+on:
+  workflow_dispatch:
+    inputs:
+      source-image:
+        description: Source image without tag
+        required: true
+        default: artifact.aerospike.io/docker/aerospike-asadm
```
pvinh-spike, line 22: From Tests agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

No automated check exercises any of this file: the repo's CI has no actionlint or shellcheck job, a dispatch-only workflow cannot run on a PR, and both manual test-plan boxes are unchecked. Prior PRs in this series linked green dry runs; this one links nothing, and it is the one PR in the series with novel logic to prove.

## Reasoning Chain

Mapping the promises to proof: copy-without-build is proven by diff parity with the merged siblings; the VERSION-derive logic (regex, trim, defaulting, output plumbing) is proven by nothing — pkg_release.sh has its own bats, but the resolve step cannot run in CI; dry-run/inspect-only and digest verification are parity code never run against this image path. Separately, the source default `artifact.aerospike.io/docker/aerospike-asadm` matches the merged siblings' convention (aql uses `docker/aerospike-aql`), but this repo's own pipeline publishes to `artifact.aerospike.io/database-docker-dev-local/aerospike-asadm` (build-and-release.yml docker jobs) and `docker/` appears nowhere else in this repo. Whether that virtual repo serves promoted asadm images — and whether `5.0.3-5*` tags exist there today — is exactly what the dry run proves. Nothing else can.

## Solution

Before merge: dispatch the workflow on this branch with defaults (dry-run stays true), link the green run in the PR, and check the first box. That one run validates the resolve step end-to-end, the derived tag set, and the `docker/` path assumption. The real-copy box can stay for the first release use.

.github/workflows/copy-tools-images-to-dockerhub.yml
```diff
@@ -0,0 +1,304 @@
...
+          registry: artifact.aerospike.io
+          username: ${{ steps.jf.outputs.oidc-user }}
+          password: ${{ steps.jf.outputs.oidc-token }}
+
+      - name: Docker login to Docker Hub
+        if: ${{ !inputs.dry-run }}
+        uses: docker/login-action@b45d80f862d83dbcd57f89517bcf500b2ab88fb2 # v4.0.0
+        with:
+          username: ${{ vars.AEROSPIKE_DOCKER_ORG_NAME || 'aerospike' }}
+          password: ${{ secrets.AEROSPIKE_DOCKER_ORG_RW_TOKEN }}
```
pvinh-spike, line 162: From Security agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

The org RW Docker Hub token is a plain secret with no environment protection, in a workflow whose inputs let the dispatcher name any source registry and any destination Hub repo. Shared with the aerospike-tools reference — flagged because the blast radius is the public `aerospike` Hub org, and this repo already knows the mitigation.

## Reasoning Chain

1. Anyone with repo write can dispatch this workflow on any ref.
2. `source-image`/`dest-image` are free-form: source set to an attacker-controlled registry, dest to any `aerospike/<repo>` the token can write, dry-run=false — the runner copies attacker content onto a public Aerospike tag. The digest check verifies dest==source, i.e. that the attacker's content copied faithfully; it is not a provenance check.
3. Constraining inputs alone wouldn't close it: workflow_dispatch executes the workflow file from the chosen ref, so a write user could dispatch a modified copy. The secret's reachability, not the input surface, is the boundary.
4. This repo already gates its sensitive master-tag push behind `environment: deploy-to-master` (build-and-release.yml:1076); the Hub-push path here gets weaker protection than a git tag does. The actor bar is repo write ≈ maintainer, so this is hardening, not a hole.

## Solution

Bind the Docker Hub push to a protected environment (reviewers and/or branch restriction) holding the token, mirroring `deploy-to-master`; independently, scope the Hub token to the repositories this workflow legitimately writes. Worth applying to the reference workflow and merged siblings too.

.github/workflows/copy-tools-images-to-dockerhub.yml
```diff
@@ -0,0 +1,304 @@
...
+          if [[ -z "$tags" ]]; then
+            tags="${RELEASE},${RELEASE}-amd64,${RELEASE}-arm64"
+          fi
+
+          alias_from="$(trim "${ALIAS_FROM_INPUT:-}")"
+          if [[ -z "$alias_from" ]]; then
+            alias_from="$RELEASE"
+          fi
+
+          alias_tags="$(trim "${ALIAS_TAGS_INPUT:-}")"
```
pvinh-spike: From Logic agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

There is no expressible way to dispatch this workflow with aliases disabled. In the reference workflow, clearing the alias-tags field skipped the alias block entirely; here, an empty or whitespace-only alias-tags is trimmed to empty and replaced with the derived default `<semver>,latest`, so a run the operator intended as "copy these tags only" also (re)pushes the `latest` and `<semver>` aliases. The only value that skips aliases is the accidental escape hatch `alias-tags=","` — undocumented and non-obvious.

## Reasoning Chain

1. Operator re-dispatches to backfill a single tag (e.g. `tags=5.0.3-5-arm64` after a partial failure), leaves alias-tags empty, unchecks dry-run.
2. Resolve step: `trim("")` → empty → `alias_tags="${SEMVER},latest"` (executed: `trim "   "` → empty; `trim ","` → `,` kept and later filtered to zero aliases).
3. Copy step receives non-empty ALIAS_TAGS → alias block runs → `latest` and `5.0.3` are pushed even though the explicit input named only one arch tag.
4. Harm is bounded (aliases repoint to the VERSION-derived release, digest-verified, dry-run defaults true), but "explicit inputs override defaults" does not hold for the value "none".

## Solution

Recognize an explicit "no aliases" value: accept `none` (or `-`) for alias-tags in the resolve step and emit an empty output for it, and say so in the input description. A named sentinel is clearer than a lucky parse artifact.

.github/workflows/copy-tools-images-to-dockerhub.yml
```diff
@@ -0,0 +1,304 @@
...
+            echo "::error::VERSION file is missing at repo root." >&2
+            exit 1
+          fi
+          VERSION=$(tr -d '[:space:]' < VERSION)
+          if [[ -z "$VERSION" ]]; then
+            echo "::error::VERSION file is empty." >&2
+            exit 1
+          fi
+          # VERSION may be 5.0.3-rc5; Docker Hub aliases only want MAJOR.MINOR.PATCH.
+          if [[ ! "$VERSION" =~ ^([0-9]+\.[0-9]+\.[0-9]+) ]]; then
```
pvinh-spike: From Logic agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

The derived defaults are only meaningful for a plain or rcN VERSION, but the resolve step derives them for ANY VERSION pkg_release.sh accepts — including non-rc pre-releases that pkg_release.sh deliberately keeps in the version. A dispatch with default inputs on a ref where VERSION is e.g. `5.0.3-beta1` defaults alias-tags to `5.0.3,latest` pointing at the `5.0.3-beta1-1` image: Docker Hub `latest` and the bare semver would alias a beta. pkg_release.sh's own header states pre-release identifiers other than rc are preserved precisely so pre-releases can never collapse onto a release name — the alias derivation undoes that separation at the Hub. (Related to but distinct from the unreleased-rc-window finding above: a *released* beta tag would pass even a tag-existence guard.)

## Reasoning Chain

1. Executed `.github/bin/pkg_release.sh 5.0.3-beta1 release` → `5.0.3-beta1-1` (accepted, exit 0).
2. The workflow regex extracts SEMVER=5.0.3 from the same string; alias-tags defaults to `5.0.3,latest`, alias-from to `5.0.3-beta1-1`.
3. An inattentive non-dry dispatch on such a ref repoints Hub `latest` to a beta build — and the whole point of deriving from VERSION is that operators stop reading the tag values.
4. Garbage suffixes are NOT a concern (executed: `5.0.3junk`, `5.0.3-`, `5.0.3-rc0` pass the workflow regex but pkg_release.sh rejects them, and the assignment-from-failed-substitution kills the step under set -e).

## Solution

When alias-tags is being defaulted, derive it only if VERSION was plain or rcN: compare `$(.github/bin/pkg_release.sh "$VERSION" version)` to `$SEMVER` and fail otherwise with "VERSION carries a non-rc pre-release identifier; pass alias-tags explicitly (or 'none')".

.github/workflows/copy-tools-images-to-dockerhub.yml
```diff
@@ -0,0 +1,304 @@
...
+            echo "::error::VERSION '$VERSION' does not start with MAJOR.MINOR.PATCH." >&2
+            exit 1
+          fi
+          SEMVER="${BASH_REMATCH[1]}"
+          RELEASE=$(.github/bin/pkg_release.sh "$VERSION" release)
+
+          trim() {
```
pvinh-spike: From Style agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

This repo's convention is that testable shell lives in .github/bin with a bats file beside it (.github/bin/test/pkg_release.bats, version_lib_test.sh); the workflow then calls the script. The resolve step instead embeds the derive logic inline in YAML, where it can never be executed outside a live dispatch. A side effect: the file now carries two trim implementations — this trim() function and the reference's inline parameter-expansion trims in the copy step — which cannot be unified because shell functions do not cross step boundaries.

## Reasoning Chain

1. The derive logic (VERSION read, semver regex, rc fold via pkg_release.sh, empty-input defaulting, trimming) is pure string computation — exactly the shape pkg_release.sh already extracts and tests.
2. Inline in YAML it is unreachable by bats, so the derive logic stays permanently dependent on manual dispatches for proof — the Evidence gap becomes structural.
3. The copy step deliberately tracks the shared reference and should stay as-is; only the resolve step is this repo's own code.

## Solution

Extract the resolve step's body to .github/bin/resolve_copy_tags.sh taking the VERSION contents plus the three inputs and emitting the KEY=value lines; add .github/bin/test/resolve_copy_tags.bats covering rc fold, non-rc VERSION, explicit overrides, whitespace inputs, malformed VERSION. The workflow step shrinks to a call plus the GITHUB_OUTPUT append. Fine as a fast-follow if the team prefers merging the workflow shape first.

.github/workflows/copy-tools-images-to-dockerhub.yml
mcoberly2: Architecture — unreleased VERSION becoming :latest. Empty tags now require git tag VERSION; explicit tags still bypass it.

.github/workflows/copy-tools-images-to-dockerhub.yml
mcoberly2: Logic — no way to disable aliases. alias-tags=none (or -) copies tags only.

.github/workflows/copy-tools-images-to-dockerhub.yml
mcoberly2: Logic — beta/non-rc defaulting to :latest. Default aliases run only when pkg_release version equals the three-digit semver.

.github/workflows/copy-tools-images-to-dockerhub.yml
mcoberly2: Style — extract resolve_copy_tags.sh + bats. Claude called this a fast-follow. You dropped it when slimming to one file.

.github/workflows/copy-tools-images-to-dockerhub.yml
mcoberly2: Security — bind Hub writes to deploy-to-master. Also dropped as extra. Same gap exists on the sibling copy workflows.

.github/workflows/copy-tools-images-to-dockerhub.yml
mcoberly2: Tests — 
Cannot close it from the PR branch only after merge

dry-run on this branch and link a green run. That is the remaining Important item. GitHub will not let you workflow_dispatch this file until it is on master, so you cannot close it from the PR branch. After merge: run with dry-run on and empty tag inputs, then paste the run URL. That also answers whether artifact.aerospike.io/docker/aerospike-asadm actually serves 5.0.3-5.

PR #554, opened 2026-09-08
chore: move server feature floor from 8.1.3 to 8.2.0

a-spiker:
The features asadm gates on 8.1.3 are shipping in 8.2.0 instead:

- `SERVER_SINDEX_INTEGER_TYPE_FIRST_VERSION` (`numeric` renamed `integer`)
- `SERVER_SINDEX_ON_AEL_FIRST_VERSION` (server compiles AEL source itself)
- `SERVER_MEMORY_ALLOC_STATS_FIRST_VERSION` (index/sindex alloc + cgroup stats)

## Changes

- Bump the three version constants in `lib/utils/constants.py`. The user-facing
  strings (`Allocation figures require server X or later`, the sindex type
  deprecation warning, the AEL `Requires server >= X` help) all read from these,
  so they follow automatically.
- Update the comments and docstrings that spell out the release.
- Update the unit and e2e fixtures. The "unsupported" fixtures stay on 8.1.x /
  8.0.x, which are still below the new floor.
- Repoint the e2e server image at
  `artifact.aerospike.io/database-docker-test-local/aerospike-server-enterprise:8.2.0.0-20260908214755`
  (note the registry moves from `database-docker-dev-local` to
  `database-docker-test-local`). Both are still overridable via
  `ASADM_E2E_IMAGE_REPO` / `ASADM_E2E_SERVER_TAG`.

No behaviour change beyond the version each gate compares against.

## Testing

`make unit`: 2115 passed, 1 skipped, 296 subtests passed.


pvinh-spike:

**Promises:**
1. The three feature gates compare against 8.2 instead of 8.1.3, so 8.1.3–8.1.x servers are treated as not having these features.
2. All user-facing strings derived from these constants follow automatically; no hardcoded 8.1.3 is left.
3. No behaviour change beyond the version each gate compares against.
4. Unit and e2e fixtures updated consistently; "unsupported" fixtures stay below the new floor.
5. e2e runs against `8.2.0.0-20260908214755` from `database-docker-test-local`, still env-overridable.
6. `make unit` passes.

| Gate | Status |
|------|--------|
| Contract | pass |
| Impacted surface | **fail** |
| Failure paths | pass |
| Evidence | **fail** |
| Quality | **fail** |
| Backward Compat | pass |

**5 findings** posted inline (0 Critical, 2 Important, 3 Minor).

**Two things checked out cleanly, and are worth recording because they were the obvious places to doubt this PR:**

**The floor bump itself is correct.** Verified against `citrusleaf/aerospike-server-enterprise`: `index_shmem_alloc_bytes`, `cgroup_memory_limit_bytes`, `host_total_mem_kbytes`, the sindex `v=v2` list format and the `"integer"` type name are all **absent at tag `8.1.3.0`** and **present at tag `8.2.0.0`**. The `8.1.3.0` tag points at a May 2026 dependabot commit that predates the feature work entirely, and release branches stop at `hotfix/8.1.2`. Raising the floor denies nothing a released 8.1.3 server can actually do.

**The registry move is verified by CI.** The `Tests` workflow ran on `3892227` and `make coverage` executed the full e2e suite — **358 passed, 7 skipped in 22:46** — pulling from `database-docker-test-local`. So the JFrog OIDC identity does have read on the new repository, and promises 5 and 6 are evidenced by CI rather than only by a local run. Worth noting for future reviewers: e2e *is* on the PR critical path here, via `coverage` → `integration-cov`.

The findings below are about the **test suite's ability to defend this change**, not about the change being wrong.

lib/utils/constants.py
```diff
@@ -229,9 +229,9 @@ class JobType:
 SERVER_SINDEX_ON_CDT_FIRST_VERSION = "6.1"
 SERVER_SINDEX_ON_EXP_FIRST_VERSION = "8.1"
 SERVER_SINDEX_SET_INDEX_FIRST_VERSION = "8.1.2"
-SERVER_SINDEX_INTEGER_TYPE_FIRST_VERSION = "8.1.3"  # "numeric" type renamed "integer"
-SERVER_SINDEX_ON_AEL_FIRST_VERSION = "8.1.3"  # server compiles AEL source itself
-SERVER_MEMORY_ALLOC_STATS_FIRST_VERSION = "8.1.3"  # index/sindex alloc + cgroup stats
+SERVER_SINDEX_INTEGER_TYPE_FIRST_VERSION = "8.2"  # "numeric" type renamed "integer"
```
pvinh-spike, line 232: From Tests agent:

Severity: important
Problem Confidence: high
Solution Confidence: high

## Problem

The regression this PR exists to create — 8.1.3 must no longer enable the integer sindex type, the v2 `sindex-list` format, or AEL indexing — is locked down by zero tests. For two of the three constants, nothing in the suite can detect a wrong value at all.

`grep -rn "8\.1\.3"` over the head tree returns nothing. Every "old" fixture is 8.0.x/8.1.0/8.1.1/8.1.2; every "new" one is 8.2.0/8.3.0. There is no fixture anywhere in `[8.1.3, 8.2)` — precisely the interval whose classification this PR changes. So every fixture sits on the same side of *both* the old floor and the new one, and no test can tell them apart.

This was demonstrated, not assumed. Reverting only `SERVER_SINDEX_INTEGER_TYPE_FIRST_VERSION` and `SERVER_SINDEX_ON_AEL_FIRST_VERSION` to `"8.1.3"` and running the two files that own every sindex/AEL gate produces a **byte-identical pass/fail ledger**. Reverting `SERVER_MEMORY_ALLOC_STATS_FIRST_VERSION` the same way fails 11 tests — so the experiment is capable of detecting a wrong floor, and these two specifically are undefended.

The mechanism is circular assertions. Every sindex/AEL message assertion builds its expected string from the same constant the production code reads:

- `test/unit/live_cluster/test_manage_controller.py:2113`, `:2133`, `:2377`, `:2399`
- `test/e2e/live_cluster/test_manage.py:775-785` — doubly circular: it picks which branch to expect by comparing `lib.server_build()` against the same constant, *and* formats the expected warning from it.

The alloc-stats path escaped only because two assertions spell the version out as a literal. Even those are weak: `assertIn("Allocation figures require server 8.2", ...)` (`test/unit/live_cluster/test_info_controller.py:252`, `test/unit/collectinfo_analyzer/test_info_controller.py:263`) is a prefix match satisfied by `8.2`, `8.2.0`, `8.2.5`, `8.20`.

Consequence: if a rebase, bad merge, or revert puts either sindex constant back to `"8.1.3"`, CI stays green and asadm silently resumes sending `type=integer`, `sindex-list:v=v2` and `ael=` to servers that reject all three.

This is the same gap Copilot raised on this file. That thread was marked resolved without a boundary test being added.

## Reasoning Chain

1. Grepped the head tree for `8\.1\.3` and for `8\.1\.[3-9]` — zero hits in either.
2. Enumerated every 8.x build literal in `test/unit`: the fixture set jumps from 8.1.2 straight to 8.2.0.
3. Therefore no fixture distinguishes floor `8.1.3` from floor `8.2` — they classify identically under both.
4. Verified by construction: reverted the two sindex constants in a scratch copy and re-ran the owning test files; identical ledger.
5. Ran the same experiment against the memory constant as a control: 11 failures. The difference isolates exactly which constants are unprotected.
6. Traced why: read the expected-string construction at each assertion site; every sindex/AEL one interpolates the constant, while the two alloc-stats ones use literals.

## Solution

Add boundary fixtures pinned inside `[8.1.3, 8.2)`:

- `test/unit/live_cluster/client/test_node.py`, beside `test_info_sindex_v2_on_new_server`: set `self.node.build = "8.1.3.0"` and assert `self.info_mock.assert_called_once_with("sindex-list:", self.ip)`. `assert_called_once_with` is load-bearing — it proves the AEL b64 follow-up call is suppressed too, covering `SERVER_SINDEX_ON_AEL_FIRST_VERSION` at the same boundary.
- `test/unit/live_cluster/test_manage_controller.py`, beside `test_create_integer_not_supported_on_old_server` and `test_ael_not_supported`: add `{"principal": "8.1.3.0"}` cases that must still raise.
- `test/unit/utils/test_util.py::NodesMissingMemoryAllocStatsTest`: add `"boundary": "8.1.3"` to the builds dict and to the expected-unsupported list.

Separately, break the circularity: write at least one expected message per constant as a literal (`"The 'integer' sindex type requires server v. 8.2 or later"`) rather than `.format(constants.SERVER_...)`, and extend the two `assertIn` calls to `"require server 8.2 or later"` so the trailing token closes the prefix match. For the e2e test, keep the constant in the branch *selector* (it legitimately adapts to whichever server the suite targets) but make the expected warning body literal.

Cost: the next deliberate floor move must state the new floor twice, once in code and once as an expectation. That is the point of the change. These are pure test additions, so nothing in production can break.

lib/utils/util.py
```diff
@@ -655,7 +655,7 @@ def aggregate_ns_memory_stats(ns_stats, editions=None):
 
 def node_reports_memory_alloc_stats(build):
     """
-    Whether a build reports the 8.1.3 memory allocation statistics.
+    Whether a build reports the 8.2 memory allocation statistics.
```
pvinh-spike, line 658: From Data Flow agent:

Severity: important
Problem Confidence: high
Solution Confidence: medium

## Problem

Raising `SERVER_MEMORY_ALLOC_STATS_FIRST_VERSION` opens a build window `[8.1.3, 8.2)` in which the headline table and the verbose tables disagree about the same node in the same session — which is exactly what the docstring three lines below this one says cannot happen.

Only one of the three consumers is version-gated:

- `lib/utils/util.py:755` — `if ns_known and alloc_known:` suppresses `allocated_bytes` / `allocated_shmem_bytes` in the headline for a node below the floor. Gated.
- `lib/view/templates.py:3185` — `info_memory_index_sheet` ("Index and Data Memory", `info memory --verbose`) declares `from_source=("node_names", "node_ids", "ns_agg")` and projects `total_alloc_bytes` / `pi_alloc_bytes` / `si_alloc_bytes` / `set_alloc_bytes` straight out of `ns_agg`. **No build in scope, no gate.**
- `lib/view/templates.py:383`, `:482` — the `info namespace usage` Index/Sindex `Alloc` columns project `index_shmem_alloc_bytes` / `sindex_shmem_alloc_bytes` directly. **No gate.**

`CliView.info_memory` (`lib/view/view.py:298-320`) makes this concrete: it passes `dict(headline=headline, builds=builds, **node_src)` to the headline sheet, but `dict(ns_agg=ns_agg, **node_src)` to the verbose index sheet. `builds` is filtered at `:295` and then never reaches the verbose sources.

`aggregate_ns_memory_stats` is purely presence-driven (`util.py:641-649`: emit the total iff an arena key is present). So a node that *reports* the arena stats but *builds* below the floor gets a warning naming it, a blank `Allocated Total`/`Shmem` in the headline, and populated `Alloc` values in the verbose table and in `info namespace usage`.

`nodes_missing_memory_alloc_stats`' docstring claims "One predicate for the headline gate, **the verbose gate**, and the warning." There is no verbose gate. Before this PR the window was empty — the floor equalled the version the stats shipped in, so "reports the stats" and "at or above the floor" coincided. The bump decouples them.

On exposure, being precise: no *released* server in `[8.1.3, 8.2)` reports these stats (checked the server repo — tag `8.1.3.0` has none of them), so this is not a fleet-wide issue. But the pre-image of `test/e2e/lib.py` pinned `8.1.3.0-20260904184619`, a dev build that this PR's own `test/e2e/live_cluster/test_info.py:163-169` documents as reporting the arena stats. Any collectinfo archive captured from that build replays through `lib/collectinfo_analyzer/info_controller.py:145-152` and hits the split — warned about, blanked in the headline, still showing bytes elsewhere. Collectinfo archives are read long after the cluster they came from is gone.

## Reasoning Chain

1. Grepped `SERVER_MEMORY_ALLOC_STATS_FIRST_VERSION`: consumed only by `node_reports_memory_alloc_stats` and the two `do_memory` warning strings.
2. Traced that predicate to its only two uses — `nodes_missing_memory_alloc_stats` (the warning) and `alloc_known` in `derive_memory_headline` (`util.py:724`, applied at `:755`). Nothing else.
3. Read `CliView.info_memory`: `builds` goes to the headline sheet's sources and to no other sheet.
4. Read `info_memory_index_sheet` and `info_namespace_usage_sheet` `from_source` tuples: neither declares `builds`, so suppression there is presence-of-stat only.
5. Read `aggregate_ns_memory_stats` / `_accumulate_ns_memory`: no build parameter anywhere — confirming presence-only.
6. Read `MemoryTablesAgreeTest` (`test/unit/utils/test_util.py:1694`), which exists specifically to pin this invariant. Its two cases are both *aligned* — no arena stat + build 8.1.2, and arena stat + build 8.2.0. The mixed case is untested, which is why the bump passes CI.
7. Confirmed the window was empty before the bump and is non-empty after, making this a change this PR introduces rather than a pre-existing condition.

## Solution

Thread the same predicate into the verbose path. In `CliView.info_memory`, before rendering the verbose sheets, strip the alloc keys (`total_alloc_bytes`, `pi_alloc_bytes`, `si_alloc_bytes`, `set_alloc_bytes`, `data_alloc_bytes`, `shmem_alloc_bytes`) from `ns_agg[node]` for every node named by `util.nodes_missing_memory_alloc_stats(builds)`. `info namespace usage` needs `builds` threaded in from `do_namespace_usage`, which today receives none.

The cheaper alternative is to delete the gate from `derive_memory_headline` and let presence-of-stat decide everywhere, leaving the warning advisory. That makes all three tables agree by construction and removes this whole class of floor-bump skew — but it is a deliberate reversal of what `node_reports_memory_alloc_stats`' docstring was written to prevent, not a cleanup, so it needs your call rather than mine.

Either way, add the missing `MemoryTablesAgreeTest` case — arena stat present, `builds={"node1": "8.1.3"}`, assert the two tables agree. It fails today. Solution confidence is medium because which of the two directions you want is a product judgment I can't make for you; the missing test case is unambiguous either way.

test/unit/utils/test_util.py
```diff
@@ -1743,7 +1743,7 @@ def test_both_tables_publish_their_total_on_a_supported_build(self):
             util.derive_memory_stats({"node1": {"heap_allocated_kbytes": "1000"}}),
             {},
             ns_agg,
-            builds={"node1": "8.1.3"},
+            builds={"node1": "8.2.0"},
```
pvinh-spike, line 1746: From Complexity agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

The helper this test calls, `_pre_8_1_3_memory_engine_node` (`test/unit/utils/test_util.py:1702`), still names the old floor. It is the one surviving 8.1.3 reference in the tree, and it is now wrong in two ways:

- It claims to model a node below the memory-alloc floor. The floor is now 8.2, so "pre 8.1.3" no longer describes the boundary the fixture tests.
- This test — the *supported* case — calls it, adds `index_shmem_alloc_bytes`, and pairs it with `builds={"node1": "8.2.0"}`. A helper named "pre-8.1.3" now feeds the supported path as well as the unsupported one at `:1715`, so the name describes neither caller.

This is the identifier a `grep '8\.1\.3'` sweep cannot see, because the dots are underscores. That is exactly why it survived a 14-file bump that otherwise cleared every 8.1.3 reference.

## Reasoning Chain

1. `grep -rn '8\.1\.3'` over the head tree returned nothing, which appears to satisfy the promise that no hardcoded 8.1.3 remains.
2. Re-ran with the separator-free spellings (`8_1_3`, `8-1-3`) — three hits, all in `test_util.py`, all this one helper and its two call sites.
3. Read `MemoryTablesAgreeTest` (`:1694-1749`) and confirmed the helper is shared by the below-floor case (`builds={"node1": "8.1.2"}`) and this supported case (`builds={"node1": "8.2.0"}`), so the version in its name corresponds to neither.

## Solution

Rename to something version-free that describes the base fixture rather than either test — e.g. `_memory_engine_node_without_arena_stats` — and update the two call sites at `:1715` and `:1734`. Nothing else references it, so the rename is contained. Note the second caller mutates the fixture into a *with*-arena-stats node, so the name should describe the base fixture, not the scenario.

test/unit/utils/test_util.py
```diff
@@ -1668,7 +1668,7 @@ def test_stop_writes_threshold_absent_when_unconfigured(self):
 class NodesMissingMemoryAllocStatsTest(unittest.TestCase):
     def test_unsupported_builds_are_named(self):
         builds = {
-            "new": "8.1.3",
+            "new": "8.2.0",
```
pvinh-spike, line 1671: From Style agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

`"new"` and `"newer"` now hold the identical value `"8.2.0"`.

On master these were `"8.1.3"` (exactly at the floor) and `"8.2.0"` (above it) — two deliberately distinct equivalence classes in a test named `test_unsupported_builds_are_named`. The mechanical `8.1.3` → `8.2.0` substitution collapsed them. The key `"newer"` now contradicts the data underneath it, and the case it existed to cover — a build strictly above the floor — is gone.

The test still passes, so nothing flags it, but one of its eight fixtures is now dead weight and the fixture set no longer contains anything strictly above the floor.

## Reasoning Chain

1. The diff changes only the `"new"` line (`"8.1.3"` → `"8.2.0"`) and leaves `"newer": "8.2.0"` untouched.
2. Reading the head tree confirms both keys now carry `"8.2.0"`.
3. The pre-image is legible from the diff: `"new"` was the at-the-floor case, `"newer"` the above-the-floor case.
4. Post-change both land in the same class, so the fixture proves the same fact twice. The name `"newer"` describes a relationship the values no longer express — the same defect class as a comment that contradicts its code.

## Solution

Restore the two classes:

```python
"new": "8.2",
"newer": "8.2.0",
```

or `"new": "8.2.0"` / `"newer": "8.3.0"`, matching how the `("all_supported", ...)` case in `test/unit/live_cluster/test_info_controller.py:289` was updated in this same PR. Nothing else reads this dict, and the expected-names list at `:1680-1682` already excludes both keys, so no other change is needed.

test/e2e/lib.py
```diff
@@ -50,9 +50,9 @@
 
 IMAGE_REPO = os.environ.get(
     "ASADM_E2E_IMAGE_REPO",
-    "artifact.aerospike.io/database-docker-dev-local/aerospike-server-enterprise",
+    "artifact.aerospike.io/database-docker-test-local/aerospike-server-enterprise",
 )
-SERVER_TAG = os.environ.get("ASADM_E2E_SERVER_TAG", "8.1.3.0-20260904184619")
+SERVER_TAG = os.environ.get("ASADM_E2E_SERVER_TAG", "8.2.0.0-20260908214755")
```
pvinh-spike, line 55: From Architecture agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

`8.2.0.0-20260908214755` is a dated snapshot tag, not a released one, and it sits on the CI critical path for **every** push to this repo.

`.github/workflows/tests.yml:52` runs `make coverage`, which is `unit-cov` + `integration-cov` (`makefile:143-147`), and `integration-cov` runs the full e2e suite against `IMAGE_REPO:SERVER_TAG`. The workflow sets neither `ASADM_E2E_IMAGE_REPO` nor `ASADM_E2E_SERVER_TAG`, so CI uses this hardcoded value. I confirmed it actually runs: the `Tests` run on this head SHA executed 358 e2e tests in 22:46 and passed, so the credential and registry side is fine.

The durability is the exposure. When this snapshot is garbage-collected from `database-docker-test-local`, `test/e2e/lib.py:544` fails the pull, falls through to the `DOCKER_CLIENT.images.list` check, finds no local copy on a fresh runner, and raises — taking down the Tests job for every subsequent PR on this repo, not just this one. The previous pin (`8.1.3.0-20260904184619`, four days old when this PR was opened) shows the cadence these get replaced at.

The failure will also be misdiagnosed. The message at `lib.py:549-552` says the image "comes from an authenticated registry; run `docker login artifact.aerospike.io`" — so a runner that is correctly logged in but facing a GC'd tag gets sent after a credential problem that does not exist.

## Reasoning Chain

1. Read `test/e2e/lib.py:51-55` — repo and tag are plain strings, env-overridable, no digest.
2. Read `.github/workflows/tests.yml` and `makefile:122-147` — the test step is `make coverage`, which reaches `integration-cov` → `pytest test/e2e/live_cluster`, so the pull is on the CI critical path rather than developer-only.
3. Confirmed empirically from the `Tests` job log for `3892227`: the e2e suite ran and passed, so the tag currently resolves and the OIDC identity can read the new repository.
4. Read `lib.py:541-555` — on pull failure it raises unless a local image of that name already exists, which on a hosted runner it never does.
5. Compared against the tag this PR replaces (four days old) to establish that these snapshots turn over quickly rather than persisting.

## Solution

Repoint at the promoted, non-dated `8.2.0.0` tag once one exists in the release repository; land the dated snapshot only as an explicit stopgap with a comment saying so.

If a dated tag is unavoidable for now, at least make the eventual failure legible — extend the `raise` at `lib.py:548-553` to name the tag and offer both causes ("not logged in, *or* this snapshot may have been garbage-collected; set `ASADM_E2E_SERVER_TAG` to a current build"), since the two are indistinguishable from the Docker error.

Solution confidence is medium only because the right replacement tag depends on release state I can't see from here; the direction is clear.

lib/utils/constants.py
a-spiker: 8.1.3 was never released. That build was renamed 8.2.0 and the 8.1.3 line abandoned, so `[8.1.3, 8.2)` is empty by construction rather than merely empty of released servers. A fixture pinned inside it would assert asadm's behaviour against a server that does not exist, which is why there is no fixture there and why one should not be added.

On the circularity: I tried it and backed it out. Interpolating the constant is deliberate. Hardcoding `8.2` into every sindex/AEL assertion is what turns the next floor move from a 14-file change into a 40-file one, and every one of those literals is a fresh chance to miss a site. The failure being defended against is a bad merge silently reverting a constant, and that is cheaper to catch with a single equality assertion on the constants themselves than by spreading the version across every assertion site.

lib/utils/util.py
a-spiker: Same root cause as the constants thread: the window this rests on does not exist. 8.1.3 was renamed 8.2.0 before release and the 8.1.3 line abandoned, so no build reports the arena stats while sitting below the floor. "Reports the stats" and "at or above the floor" still coincide exactly as they did before the bump, and the presence gate in `aggregate_ns_memory_stats` remains the operative one.

The reading of the code is correct: there is no verbose gate, and `builds` reaches only the headline sheet. But the bump does not decouple anything, so I am not threading the predicate into the verbose sheets or into `info namespace usage` here. Worth recording that the `info namespace usage` Alloc columns are multi-source, projecting `index_shmem_alloc_bytes` alongside `index_pmem_alloc_bytes` and `index_flash_alloc_bytes`, so a build gate there would also suppress device arenas that predate 8.2.

The `nodes_missing_memory_alloc_stats` docstring naming a "verbose gate" that is not there is a fair catch, but it is pre-existing wording on master, not something this PR introduces.

test/unit/utils/test_util.py
a-spiker: Fixed. Renamed to `_memory_engine_node_without_arena_stats` and updated both call sites. The name now describes the base fixture rather than either scenario, and it carries no version at all, so it cannot go stale on the next floor move.

test/unit/utils/test_util.py
a-spiker: Fixed. `"newer"` is now `"8.3.0"`, so `"new"` sits exactly at the floor and `"newer"` strictly above it, restoring the two equivalence classes the mechanical substitution collapsed.

test/e2e/lib.py
a-spiker: Leaving the pin as is. Repointing at a promoted, non-dated `8.2.0.0` is the right end state, but it belongs with that promotion rather than with this floor bump, and the tag will be repointed then regardless.

Skipping the error-message rewording too. It does not change what happens when the snapshot is collected, and adding a stopgap comment beside a pin that is already understood to be temporary is not worth the churn.

pvinh-spike (approved):
LGTM

PR #557, opened 2026-09-14
deps: bump Python package pins and pinned GitHub Actions

a-spiker:
Two independent pin sweeps: the Python packages in the `Pipfile`, and the SHA-pinned GitHub Actions in `.github/workflows/`.

# Python packages

## What

Bump the explicitly pinned Python packages in the `Pipfile` to their latest releases and regenerate `Pipfile.lock`.

| Package | From | To |
| --- | --- | --- |
| `msgpack` | 1.2.1 | 1.2.2 |
| `setuptools` | 83.0.0 | 84.0.0 |
| `pyinstaller` | 6.20.0 | 6.22.3 |
| `black` | 26.3.1 | 26.5.1 |

cryptography and pyOpenSSL are deliberately **not** bumped. See below.

The re-lock also pulls in transitive updates: `cffi` 2.1.1, `certifi` 2026.7.22, `charset-normalizer` 3.5.1, `click` 8.5.0, `coverage` 7.16.1, `idna` 3.19, `packaging` 26.3, `platformdirs` 4.11.8, `pygments` 2.21.0, `pyinstaller-hooks-contrib` 2026.7.

## What is not changing

The other pins (`bcrypt`, `distro`, `jsonschema`, `ply`, `toml`, `yappi`, `python-dateutil`, `flake8`) and the unpinned packages (`asyncssh`, `typing-extensions`, `pytest`, `mock`, `macholib`, `aerospike`, `docker`, `parameterized`) are already at their latest published releases, so the re-lock left them where they were.

## cryptography stays at 48.0.1

The first push of this branch took cryptography to 50.0.1 and pyOpenSSL to 26.4.0. That broke the `macos-15-intel` build, so both are reverted to the versions on master.

cryptography deprecated x86_64 macOS in [47.0.0](https://cryptography.io/en/stable/changelog/) ("Support for `x86_64` macOS (including publishing wheels) is deprecated and will be removed in the next release") and removed it in 49.0.0 ("**BACKWARDS INCOMPATIBLE:** Support for `x86_64` macOS has been removed. We now only publish `arm64` wheels for macOS").

Worth being precise about what that means: upstream dropped **publishing prebuilt Intel wheels**, not the ability to build and run on Intel macOS. Without a wheel, pip falls back to a from-source Rust build on the Intel runner. That build succeeds, the "Build asadm" step is green. What fails is the resulting binary at startup:

```
ImportError: dlopen(.../cryptography/hazmat/bindings/_rust.abi3.so, 0x0002):
  Symbol not found: _SSL_get0_group_name
  Expected in: .../_internal/libssl.3.dylib
```

PyInstaller bundles `/usr/local/opt/openssl@3/lib/libssl.3.dylib`, which does not export the symbol the freshly compiled extension was linked against. Build-time and bundle-time OpenSSL disagree. That is a packaging mismatch on our side, and fixing it means pinning one OpenSSL for both the cargo build and the PyInstaller bundle.

That work belongs in its own change, together with the question of how long asadm keeps shipping an Intel macOS package, rather than inside a routine pin bump.

48.0.1 is the last cryptography with an Intel macOS wheel, and pyOpenSSL 26.2.0 is the last release accepting `cryptography<49`, so both are held at their current versions.

### Consequence

Dependabot alert #132, GHSA-g6cj-pr64-35w5 / CVE-2026-69247 (high), "PKCS#7 EnvelopedData decryption exposes a Bleichenbacher oracle through distinguishable errors and timing", stays open. It is fixed in cryptography 50.0.0, which is the release we cannot take yet. asadm does not call the PKCS#7 EnvelopedData decryption APIs, so there is no practical exposure, but the alert will remain until the Intel build question is settled.

## Risk

With cryptography held back, the remaining Python bumps are low risk: a msgpack patch, and dev-only bumps to setuptools, pyinstaller, and black. The pyinstaller bump is the only one that affects packaging, and it is covered by the build check below.

## Testing

- `black --check .` clean, 172 files unchanged.
- `pytest test/unit`: 2209 passed, 1 skipped, 296 subtests passed.
- Re-ran both after reverting cryptography, same results.
- `make one-dir` builds successfully; the resulting `asadm` and `asinfo` binaries launch and report their version.

# GitHub Actions

## What

Bring every SHA-pinned action up to its latest release. Several actions had drifted into two to four different pins across the nine workflow files; each is now on a single SHA everywhere.

| Action | From | To |
| --- | --- | --- |
| `step-security/harden-runner` | v2.13.3 / v2.16.0 / v2.19.3 / v2.19.4 | v2.21.1 |
| `actions/checkout` | v4.3.1 / v6.0.2 | v7.0.1 |
| `actions/setup-python` | v3.1.4 / v5.6.0 / v6.2.0 | v7.0.0 |
| `actions/upload-artifact` | v4.6.2 / v7.0.0 | v7.0.1 |
| `actions/download-artifact` | v4.3.0 | v8.0.1 |
| `actions/cache` | v4.3.0 / v5.0.4 | v6.1.0 |
| `github/codeql-action` | v3.28.1 | v4.38.0 |
| `docker/login-action` | v4.0.0 | v4.6.0 |
| `step-security/setup-buildx-action` | v4.0.0 | v4.2.0 |
| `step-security/action-gh-release` | v2.4.1 | v3.0.3 |
| `step-security/action-cond` | v1.2.4 | v1.2.5 |
| `kenchan0130/actions-system-info` | v1.3.0 | v1.4.0 |
| `codecov/codecov-action` | v3.1.6 | v7.0.0 |
| `snyk/actions/setup` | unlabelled 2024-09-06 master SHA | v1.0.0 |

Two comment fixes rode along:

- The three `github/codeql-action` steps carried a `# v2.28.1` comment on a SHA that is actually v3.28.1. Corrected with the bump.
- `snyk/actions/setup` was pinned to a bare master SHA with no version comment and now points at the v1.0.0 tag with a matching one. The Snyk CLI version itself is unaffected, it is pinned separately through the `snyk-version` input.

Already latest, so left alone: `actions/attest-build-provenance` v4.2.2, `step-security/setup-jfrog-cli` v5.1.0, `step-security/slack-github-action` v3.0.3.

## Deliberately not bumped

`aerospike/shared-workflows` stays at v3.7.0 and v4.1.0. Both v4.0.0 and v5.0.0 declare breaking changes that require caller-side edits, not just a new ref:

- v4.0.0 needs added `actions: write` permission grants on the calling workflow.
- v5.0.0 stops emitting `latest` and the bare major Docker tag by default, so `reusable_docker-build-deploy.yaml` callers need a `versions-override` to keep current behavior.

Those touch the release pipeline and cannot be exercised by a PR check, so they belong in their own change.

## Risk

The major jumps worth a second look, and why they should be fine:

- `codecov/codecov-action` v3 to v7. v4 made `token` mandatory; [`tests.yml`](.github/workflows/tests.yml) already passes `secrets.CODECOV_TOKEN`. The `files`, `token`, and `fail_ci_if_error` inputs all still exist in the v7 `action.yml`.
- `step-security/action-gh-release` v2 to v3. [`draft-release.yml`](.github/workflows/draft-release.yml) uses only `draft`, `generate_release_notes`, and `files`, all still present in v3.
- `github/codeql-action` v3 to v4. The pin was 20 months stale; v4 is a runner and Node bump, and the workflow uses the stock init/autobuild/analyze trio with no removed inputs.
- The remaining `actions/*` majors are Node runtime bumps. Note the repo was already running `checkout` v6, `upload-artifact` v7, `download-artifact` v8, and `setup-python` v6 in [`build-and-release.yml`](.github/workflows/build-and-release.yml), so this mostly drags the laggard workflows up to what the main pipeline already proved out.

## Testing

Every new SHA was resolved from its upstream tag through the GitHub API and then checked back the other way, confirming the SHA matches the version in its trailing comment. All nine workflow files still parse as YAML. The workflows themselves get exercised by this PR's own CI run.


pvinh-spike (approved):
LGTM

PR #558, opened 2026-09-14
ci: build cryptography against a static OpenSSL on Intel macOS

a-spiker:
Follow-up to #557, which held `cryptography` back because it broke the Intel macOS build. This is the fix that lets it move.

Takes `cryptography` to 50.0.1 and `pyOpenSSL` to 26.4.0, and makes the Intel macOS leg build cryptography in a way that survives bundling.

## Why the first attempt failed

cryptography publishes arm64-only macOS wheels from 49.0.0 on, so the Intel leg compiles it from the sdist. That compile linked OpenSSL dynamically against Homebrew's libssl, and PyInstaller vendored it into the bundle.

The mechanism is a **base-name collision**. PyInstaller flattens every vendored dylib into one `_internal/` directory keyed by base name. Python's own `_ssl` and `_hashlib` already contribute a `libssl.3.dylib` and `libcrypto.3.dylib` there:

```
_ssl.cpython-312-darwin.so     -> @rpath/libssl.3.dylib, @rpath/libcrypto.3.dylib
_hashlib.cpython-312-darwin.so -> @rpath/libcrypto.3.dylib
```

So two different `libssl.3.dylib` files compete for one slot. One wins. On the Intel runner the winner was not the one cryptography had been compiled against:

```
ImportError: dlopen(.../cryptography/hazmat/bindings/_rust.abi3.so, 0x0002):
  Symbol not found: _SSL_get0_group_name
  Expected in: .../_internal/libssl.3.dylib
```

## The fix

Link OpenSSL statically into cryptography on the Intel leg. That takes it out of the collision entirely, and it is what the published wheels already do, so the same invariant holds on every platform. Python's `_ssl` keeps its own OpenSSL, consistent with itself.

Three environment variables, set before `pipenv install --dev`, gated to `runner.arch == 'X64'`: `OPENSSL_DIR`, `OPENSSL_STATIC=1`, and `MACOSX_DEPLOYMENT_TARGET` at the shipping floor.

That last one is not incidental. `_rust.abi3.so` is a Mach-O that [`check_min_os.sh`](.github/bin/check_min_os.sh) scans, and an unpinned compile inherits the runner's SDK, which is the exact failure that script was written to catch. The floor is read out of `pkg/Makefile` rather than repeated in the workflow, so `MACOS_MIN_VERSION` stays the single source of truth.

### OpenSSL is built here, not taken from Homebrew

An earlier revision pointed `OPENSSL_DIR` at `brew --prefix openssl@3`. That was wrong, and review caught it.

Homebrew builds each bottle on its own macOS, so `openssl@3`'s `libcrypto.a` and `libssl.a` members carry that runner's minos. Linking those in leaves the final Mach-O stamped 14.0 while the OpenSSL code inside was compiled for a later release, and `check_min_os.sh` reads only the final object, so the mismatch is invisible to it. Nothing in CI closes the gap either: the x86_64 pkg only ever executes on `macos-15-intel`, and GitHub retired the last Intel macOS 14 runner, so this cannot be covered by adding a matrix leg.

Measured locally, the two cases are unambiguous:

```
Homebrew openssl@3 archive member   ->  minos 26.0   (bottle built on macOS 26)
floor-targeted source build         ->  minos 14.0   (all 1117 objects)
```

[`build_static_openssl.sh`](.github/bin/build_static_openssl.sh) compiles a pinned, checksum-verified OpenSSL 3.6.4 against the floor, static only, then proves the result by unpacking both archives and running them through `check_min_os.sh` itself rather than reimplementing that logic. The build is cached on arch, floor, and the script's own hash, so the cost lands once instead of on every Intel run.

## The guard

[`check_crypto_static_openssl.sh`](.github/bin/check_crypto_static_openssl.sh) asserts the invariant after every macOS build, so a silent return to dynamic linking fails with a clear message instead of a startup crash.

It deliberately checks **cryptography's extension specifically**, not whether OpenSSL dylibs are present in the bundle. Those dylibs are legitimate and always will be, since Python's `_ssl` and `_hashlib` bring them. My first version of this script got that wrong and would have failed every correct build.

Both `cache-index` values move to `"5"` so the fix is not masked by a bundle built before it. There is already a comment in `build-and-release.yml` about a stale cache shipping pre-fix binaries.

## What is verified, and what is not

Verified on arm64:

- The env vars take effect. `MACOSX_DEPLOYMENT_TARGET` retags the built wheel `macosx_14_0` and sets `minos 14.0` on the extension.
- The `MACOS_MIN_VERSION` extraction returns `14.0` against the real `pkg/Makefile`.
- The guard passes a statically linked build and fails a dynamically linked one with exit 1.
- cryptography rebuilt against the floor-targeted OpenSSL reports `OpenSSL 3.6.4`, not Homebrew's 3.6.3, confirming it actually relinked. `_rust.abi3.so` stays at `minos 14.0`.
- A full `make one-dir` produces working `asadm` and `asinfo` binaries. 2209 unit tests pass, `black --check` clean.

One caveat on that relink check: the first attempt reused a cached pip wheel and proved nothing. It needs `--no-cache-dir`. CI is unaffected, no pip cache is persisted between runs.

**Only one of the two copies has run.** `mac-artifact.yml` runs on `pull_request` and is green. `build-and-release.yml` is `workflow_dispatch`-only, so the near-duplicate step block that produces the shipping `.pkg` has never executed on this branch. The two transcriptions differ exactly where it matters, subdirectory checkout versus workspace root, and one of those differences was a live bug here until ba52736. Everything in the untested copy fails closed (empty or malformed floor errors, missing archive errors, guard exits 1), so the worst case is a loud release-build failure rather than a bad artifact, but a `workflow_dispatch` dry run with `release=false` is worth doing before the next release leans on it.

**Not verified locally:** the Intel flip from dynamic to static. On arm64 every configuration I could construct links statically, including with no OpenSSL environment variables set at all, so the failing case is not reproducible off an Intel runner. That leg is confirmed by this PR's own `macos-15-intel` run, and the guard makes the verdict unambiguous either way.

## The guard failed open

As first written, [`check_crypto_static_openssl.sh`](.github/bin/check_crypto_static_openssl.sh) piped `otool` into `grep`. Under `pipefail` that makes "otool could not read it" indistinguishable from "it has no OpenSSL dylib", and the count was incremented before the inspection, so a run where every inspection failed still reported success and exited 0.

Confirmed with a stub `otool` that exits 1:

```
before:  ==> cryptography links OpenSSL statically (1 extension(s) checked)   EXIT=0
after:   error: otool failed on .../_rust.abi3.so                            EXIT=1
```

`otool`'s stderr is now surfaced instead of discarded.

`OPENSSL_DIR` was also assigned through a command substitution inside an argument, where a failure cannot trip `set -e` and an empty value would reach `openssl-sys` as a real prefix. It is now assigned on its own line with the archive confirmed present, matching the convention already documented in the `Create .pkg` step.

## Lock diff

Rebased onto master after #557 merged. The lock now differs from master by exactly two entries:

```
cryptography   ==48.0.1 -> ==50.0.1
pyopenssl      ==26.2.0 -> ==26.4.0
```

The transitive dev churn an earlier revision carried is gone, #557 absorbed it.

The new OpenSSL cache steps were written before #557 landed and carried `actions/cache` v4.3.0. They are moved to v6.1.0 to match the rest of the repo rather than reintroducing the drift that change cleaned up.

## Known follow-ups

Raised in review, deliberately left out to stop this PR growing further:

- An automated staleness signal for the pinned OpenSSL. The obligation is documented next to `VERSION`/`SHA256`, but a comment is not a mechanism. The vehicle (Renovate regex manager vs a scheduled tag check) depends on org tooling.
- `bats` coverage for the guard's failure paths in `.github/bin/test/`, next to the existing `pkg_release.bats`.
- Extracting the `MACOS_MIN_VERSION` read into `.github/bin/`, so the parsing contract lives in one place instead of two. It is shape-checked in both copies now, which makes the duplication safe rather than absent.

## Ongoing cost

This adopts a build path upstream deliberately abandoned. Every future cryptography release recompiles from source on Intel against an OpenSSL we pin ourselves, adding roughly two minutes to each cold Intel build. That is the price of keeping the Intel macOS package while staying current on cryptography.

## What this closes

Dependabot #132, GHSA-g6cj-pr64-35w5 / CVE-2026-69247 (high), fixed in cryptography 50.0.0.

## The legacy provider

Installed on an Intel Mac, the first Intel build from this branch warned on every start:

```
<frozen importlib._bootstrap>:488: Warning: OpenSSL 3's legacy provider failed to load. Legacy algorithms will not be available. If you need those algorithms, check your OpenSSL configuration.
```

Without `no-module`, OpenSSL builds the legacy provider as `legacy.dylib` under `MODULESDIR`, here `$RUNNER_TEMP/openssl-static/lib/ossl-modules`. That path is compiled into `libcrypto.a`, cryptography asks for the provider at import, and the lookup only succeeds on the runner that built it. `no-module` compiles the provider into libcrypto (`STATIC_LEGACY` in OpenSSL's `providers/build.info`), so the Intel bundle behaves like the arm64 one, which loads it from cryptography's own wheels without a word. [`build_static_openssl.sh`](.github/bin/build_static_openssl.sh) now fails if `ossl_legacy_provider_init` is missing from the archive.

CI was green through all of this. The `mac-artifact.yml` sanity step runs on the build runner, where the module directory exists, and the install smoke test only checked exit status. Two guards close that:

- The sanity step now runs `asadm --version` with `OPENSSL_MODULES=/nonexistent`, which redirects the module lookup the way a user's machine does. A built-in provider never looks.
- `test_execute.bats` and `test_execute.sh` fail on the warning text. They run against the installed package on a clean runner in `build-and-release.yml`, on every Linux distro and macOS leg and against the Docker image.


.github/workflows/build-and-release.yml
```diff
@@ -393,7 +393,7 @@ jobs:
         id: cache-asadm-asinfo
         env:
           cache-name: cache-asadm-asinfo
-          cache-index: "4"
+          cache-index: "5"
```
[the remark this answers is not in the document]
a-spiker: Agreed, fixed in ba52736. `.github/bin/build_static_openssl.sh` is now part of the existing input hash, so changing the pinned OpenSSL misses this cache instead of needing another manual `cache-index` bump.

On the support floor specifically: `pkg/Makefile` is deliberately left out of the hash. A floor change is already handled without it. Raising the floor (14.0 to 15.0) leaves a restored bundle at `minos 14.0`, which still satisfies the higher floor, so it is benign. Lowering it (15.0 to 14.0) makes a restored bundle violate the new floor, and `check_min_os.sh` in `Create .pkg` fails loudly rather than shipping it. Adding `pkg/Makefile` would invalidate the bundle cache on every unrelated packaging tweak for a case that is already either harmless or caught.

.github/workflows/mac-artifact.yml
```diff
@@ -74,11 +74,50 @@ jobs:
...
+          min_os=$(sed -n 's/^MACOS_MIN_VERSION[[:space:]]*=[[:space:]]*//p' pkg/Makefile)
+          [[ -n "${min_os}" ]] || { echo "::error::MACOS_MIN_VERSION not found in pkg/Makefile" >&2; exit 1; }
+          echo "MACOS_MIN_VERSION=${min_os}" >> "$GITHUB_ENV"
+      - name: Cache static OpenSSL
+        if: runner.arch == 'X64' && steps.cache-asadm-asinfo.outputs.cache-hit != 'true'
+        id: cache-openssl
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: ${{ runner.temp }}/openssl-static
+          key: openssl-static-${{ runner.arch }}-${{ env.MACOS_MIN_VERSION }}-${{ hashFiles('.github/bin/build_static_openssl.sh') }}
```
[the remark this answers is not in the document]
a-spiker: Correct, and fixed in ba52736.

Confirmed the failure mode: the pattern matched nothing, `hashFiles` returned the empty string, and the key collapsed to the constant `openssl-static-X64-14.0-`. Once one entry existed it would be restored by every later run, so an OpenSSL CVE bump in the build script would have been silently ignored while the Intel bundle kept linking the stale archives. That is exactly the staleness the hash was there to prevent.

The pattern is now qualified with the checkout path:

    hashFiles(format('{0}/.github/bin/build_static_openssl.sh', steps.working-dir.outputs.value))

`build-and-release.yml` needed no equivalent change, since it checks out at the workspace root.

pvinh-spike:

**Promises:**
1. Upgrades cryptography to 50.0.1 / pyOpenSSL to 26.4.0, closing CVE-2026-69247.
2. Intel macOS leg builds cryptography from sdist with statically linked OpenSSL, eliminating the PyInstaller dylib base-name collision.
3. Static OpenSSL is a pinned, checksum-verified 3.6.4 source build against the macOS floor read from pkg/Makefile, proven by check_min_os.sh.
4. check_crypto_static_openssl.sh guards the static-link invariant after every macOS build, ungated on caches, and fails loudly rather than open.
5. Cache correctness: OpenSSL cache misses on build-script changes, bundle cache includes the build script, cache-index bumps to 5.
6. No behavior change for arm64 macOS or non-macOS builds.
7. Lock diff is exactly the two intended entries.

| Gate | Status |
|------|--------|
| Contract | blind spot — promises 1, 5, 6, 7 verified; the build-and-release.yml copy of the new steps has never executed (workflow_dispatch-only), so promises 2/3 are proven only for the mac-artifact.yml transcription |
| Impacted surface | pass — both macOS bundle producers covered; env timing, gating matrix, and cache isolation traced clean |
| Failure paths | fail — 1 Important: the OpenSSL 3.6.4 pin has no owner or staleness signal (see inline) |
| Evidence | blind spot — the guard's failure paths (dynamic link, otool failure, zero files) are verified only by the author's local experiments, not by anything committed |
| Quality | pass |
| Backward Compat | pass — pyOpenSSL API usage in lib/live_cluster is covered by unit tests green under the new lock; Python floor is 3.12 everywhere, so the >=3.9 marker move strands nothing |

**7 findings** posted inline (0 Critical, 1 Important, 6 Minor).

Verified sound along the way: the guard's while-read/pipefail semantics (exit codes propagate, grep-under-if doesn't trip set -e, empty find fails closed — all confirmed by execution); no otool self-reference false positive; ar-extraction member names collision-free for OpenSSL 3.x; GITHUB_ENV write/read ordering across steps; GitHub cache isolation keeps fork-PR caches out of release runs; the 3.6.4 pin and SHA256 match upstream; lock hashes and markers match PyPI; pyOpenSSL 26.4.0's cryptography <51,>=49 constraint fits 50.0.1.

.github/bin/build_static_openssl.sh
```diff
@@ -0,0 +1,70 @@
...
+# _rust.abi3.so leaves the final Mach-O stamped with MACOSX_DEPLOYMENT_TARGET
+# while the code inside it was compiled for a later release -- and
+# check_min_os.sh reads only the final object, so the mismatch is invisible to
+# it. Compiling OpenSSL ourselves against the floor is what makes the floor the
+# binary claims actually true.
+#
+# Usage: build_static_openssl.sh <prefix> <min_macos>
+set -euo pipefail
+
+VERSION="3.6.4"
```
pvinh-spike, line 28: From Security agent:

Severity: important
Problem Confidence: high
Solution Confidence: medium

## Problem

Statically linking a pinned OpenSSL moves OpenSSL patch responsibility for the Intel macOS bundle from upstream (Homebrew, or cryptography's own wheels) into this repository — and nothing in the repository will ever notice the pin going stale. `VERSION="3.6.4"` + SHA256 are hardcoded here, and `.github/dependabot.yml` tracks only `pip`, `github-actions`, and `docker` — no ecosystem watches a version string inside a shell script.

The asymmetry is the sharp edge: when the next OpenSSL 3.6.x security release lands, the arm64 bundle gets the fix automatically through the next cryptography bump (upstream wheels vendor their own current OpenSSL), while the Intel bundle keeps linking 3.6.4 — silently, with every CI gate green. Trigger sequence:

1. OpenSSL ships 3.6.5 fixing a CVE in libssl/libcrypto.
2. Dependabot bumps cryptography; arm64 wheels pick up the fixed OpenSSL.
3. The Intel leg rebuilds cryptography from sdist against the still-pinned 3.6.4 — the bundle cache even misses (Pipfile.lock changed), so a fresh build embeds the old OpenSSL.
4. The release ships an Intel macOS asadm carrying the vulnerable OpenSSL, indefinitely, until a human remembers this script exists.

The 3.6 line has shipped three security releases in 2026 alone (3.6.2 April, 3.6.3 June, 3.6.4 August), so this cadence is measured in months. And 3.6 is not an LTS line, so the pin will also outlive its support window with no signal.

The pin itself is currently sound: 3.6.4 is the current 3.6.x, the SHA256 matches upstream, and HTTPS + in-repo SHA256 is adequate protection for the download itself.

## Reasoning Chain

1. `.github/dependabot.yml` in the PR head covers pip, github-actions, docker — no manager covers this file.
2. No workflow step compares the pin against upstream; the OpenSSL cache key rewards the pin staying unchanged (permanent cache hit).
3. The script's header explains why the build exists but says nothing about the update obligation it creates — the one thing a future maintainer responding to an OpenSSL advisory would need.
4. OpenSSL 3.6.x security releases occurred every ~2 months in 2026, so the window between "pin stale" and "pin vulnerable" is short in practice.

## Solution

Minimum: a comment block next to `VERSION=`/`SHA256=` stating the obligation explicitly — "this OpenSSL ships inside the Intel macOS asadm; on any OpenSSL 3.6.x security release, bump VERSION and SHA256 here (cache keys and the bundle cache miss automatically)" — plus an entry in the release checklist. Better: automate the staleness signal — a scheduled job that fetches the latest `openssl-3.6.*` tag from github.com/openssl/openssl and warns when it is newer than the pin, or a Renovate regex manager on this file if Renovate is available org-side. Solution confidence is medium only because the right vehicle depends on org tooling; the obligation to make the failure mode visible is not in doubt.

.github/workflows/build-and-release.yml
```diff
@@ -393,15 +393,59 @@ jobs:
...
           # served a stale bundle for every source change within one rc: a
           # pyinstaller-build.spec fix landed, the key was unchanged, `make
           # one-dir` was skipped, and the pre-fix binaries shipped.
-          key: ${{ env.cache-name }}-${{ env.cache-index }}-${{ matrix.os }}-${{ runner.arch }}-${{ steps.system-info.outputs.release }}-${{ env.PYTHON_VERSION }}-${{ needs.get-version.outputs.VERSION }}-${{ hashFiles('pyinstaller-build.spec', 'Pipfile.lock', 'makefile', 'asadm.py', 'lib/**/*.py', 'asinfo/**/*.py') }}
+          # build_static_openssl.sh is a build input too: it decides which
+          # OpenSSL gets linked into _rust.abi3.so, so patching the pinned
+          # version there has to miss this cache or the old bundle ships.
+          key: ${{ env.cache-name }}-${{ env.cache-index }}-${{ matrix.os }}-${{ runner.arch }}-${{ steps.system-info.outputs.release }}-${{ env.PYTHON_VERSION }}-${{ needs.get-version.outputs.VERSION }}-${{ hashFiles('pyinstaller-build.spec', 'Pipfile.lock', 'makefile', 'asadm.py', 'lib/**/*.py', 'asinfo/**/*.py', '.github/bin/build_static_openssl.sh') }}
```
pvinh-spike: From Logic agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

This key was extended to treat `build_static_openssl.sh` as a build input, on the correct principle that whatever determines the bytes of `_rust.abi3.so` must be able to miss this cache. But the PR introduces a second such input and does not key on it: `MACOS_MIN_VERSION` in `pkg/Makefile`. That value becomes `MACOSX_DEPLOYMENT_TARGET` for the Intel cryptography compile and selects the static OpenSSL cache entry — it determines the bytes of the bundle exactly as the build script does — yet `pkg/Makefile` is absent from the hashFiles list.

Trigger sequence (floor lowered, e.g. 14.0 → 13.0, within one VERSION):

1. An rc of version V builds on macos-15-intel; the bundle, compiled at floor 14.0, is cached under a key containing V but not the floor.
2. A commit changes only pkg/Makefile's MACOS_MIN_VERSION to 13.0.
3. Next run: bundle cache HIT. The floor step, OpenSSL steps, pipenv, and make are all skipped; the guard passes (the stale bundle is statically linked).
4. "Create .pkg" runs check_min_os.sh 13.0 over the bundle: members stamped 14.0 violate the new floor. CI fails — and keeps failing on every rerun, because the rebuild that would fix it is gated behind the cache hit. The operator's only exits are bumping cache-index or the version, neither of which the error message suggests.

Floor raised (14.0 → 15.0): same cache hit, but check_min_os passes (14.0 ≤ 15.0) and a bundle compiled against the old floor ships silently — strictly more compatible than claimed, so harm is doctrinal, which is why this is minor.

mac-artifact.yml is not affected: its bundle key includes `git describe`, which changes with every commit.

## Reasoning Chain

1. The Intel compile takes its deployment target from MACOS_MIN_VERSION (the floor step + the env-export step below), so the floor is an input to `_rust.abi3.so`.
2. The OpenSSL cache key below includes `${{ env.MACOS_MIN_VERSION }}` — the PR itself acknowledges the floor as a cache-relevant input there.
3. This bundle key hashes the spec, lock, makefile, sources, and build_static_openssl.sh — pkg/Makefile is not in the list, and VERSION does not change on a floor-only commit.
4. Every rebuild step is gated on `cache-hit != 'true'`, so a hit forecloses the rebuild; the only downstream floor check is check_min_os.sh inside `make -C pkg osx-pkg`, whose pass/fail asymmetry gives the two traces above.

## Solution

Add `'pkg/Makefile'` to the hashFiles() list in this key. One token; it makes the key's own "keyed on the build inputs" comment true and removes both the wedged-red-CI path and the stale-floor-ships path. (Alternatively key on env.MACOS_MIN_VERSION, but that requires hoisting and ungating the floor-extraction step; the hashFiles addition is simpler and sufficient.)

.github/workflows/build-and-release.yml
```diff
@@ -393,15 +393,59 @@ jobs:
...
+          [[ -n "${min_os}" ]] || { echo "::error::MACOS_MIN_VERSION not found in pkg/Makefile" >&2; exit 1; }
+          echo "MACOS_MIN_VERSION=${min_os}" >> "$GITHUB_ENV"
+
+      - name: Cache static OpenSSL
+        if: runner.arch == 'X64' && steps.cache-asadm-asinfo.outputs.cache-hit != 'true'
+        id: cache-openssl
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: ${{ runner.temp }}/openssl-static
+          key: openssl-static-${{ runner.arch }}-${{ env.MACOS_MIN_VERSION }}-${{ hashFiles('.github/bin/build_static_openssl.sh') }}
```
pvinh-spike: From Data Flow agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

The static-OpenSSL cache key hashes only `build_static_openssl.sh`, but that script's validation gate is `check_min_os.sh`, invoked via `$here` at build time. The archive-member floor scan runs exactly once — at build time — and the result is then frozen into the cache. A later fix or tightening of `check_min_os.sh` never re-validates the cached libraries, and this is the one place where no downstream check can compensate: `build_static_openssl.sh`'s own header states that once the `.a` members are linked into `_rust.abi3.so`, the final object carries only the stamped deployment target — the whole-bundle `check_min_os.sh` run in `pkg/Makefile` is structurally blind to it. Both other new checks are ungated on caches by design; this one is silently gated. (Same applies to the mac-artifact.yml copy of this key.)

Trigger sequence:

1. check_min_os.sh has a scanning gap (say, a load-command layout its parser misses).
2. CI builds OpenSSL; a floor-violating member slips past; the result is cached.
3. A maintainer fixes the scanner in a later PR. build_static_openssl.sh is unchanged, so the OpenSSL cache key is unchanged — cache hit, build step skipped, the fixed scanner never sees the archive members.
4. The violating code links into _rust.abi3.so; the final-bundle scan cannot detect it; the Intel asadm ships a binary that dies on the supported macOS floor — the exact failure class this PR's machinery exists to prevent.

## Reasoning Chain

1. Both cache keys hash the single file `build_static_openssl.sh` (the mac-artifact variant is path-qualified but hashes the same one file).
2. build_static_openssl.sh executes check_min_os.sh as a hard gate — the checker is a behavioral input of the build, exactly the property this PR's own bundle-cache-key comment argues for.
3. Post-link, member minos is unrecoverable from the final Mach-O (per the build script's header), so build-time validation is the only validation.
4. Contrast with the bundle-cache treatment: check_crypto_static_openssl.sh was deliberately left ungated on caches; the floor check got no equivalent treatment for the OpenSSL cache.

## Solution

Add the checker to both keys: `hashFiles('.github/bin/build_static_openssl.sh', '.github/bin/check_min_os.sh')` here, and the two format()-qualified paths in mac-artifact.yml. Cost is one extra OpenSSL rebuild per checker change; the checker changes rarely.

.github/workflows/build-and-release.yml
```diff
@@ -393,15 +393,59 @@ jobs:
...
+          # version there has to miss this cache or the old bundle ships.
+          key: ${{ env.cache-name }}-${{ env.cache-index }}-${{ matrix.os }}-${{ runner.arch }}-${{ steps.system-info.outputs.release }}-${{ env.PYTHON_VERSION }}-${{ needs.get-version.outputs.VERSION }}-${{ hashFiles('pyinstaller-build.spec', 'Pipfile.lock', 'makefile', 'asadm.py', 'lib/**/*.py', 'asinfo/**/*.py', '.github/bin/build_static_openssl.sh') }}
+
+      - name: Get the macOS support floor
+        # Read the floor from pkg/Makefile rather than repeating it here, so the
+        # shipping bundle and the Intel compile below cannot drift apart.
+        if: runner.arch == 'X64' && steps.cache-asadm-asinfo.outputs.cache-hit != 'true'
+        run: |
+          set -euo pipefail
+          min_os=$(sed -n 's/^MACOS_MIN_VERSION[[:space:]]*=[[:space:]]*//p' pkg/Makefile)
```
pvinh-spike, line 419: From Architecture agent:

Severity: minor
Problem Confidence: medium
Solution Confidence: high

## Problem

Two related weaknesses in the floor extraction, both fixable with one move.

First, it parses Make syntax by hand, and fails OPEN on inputs it does not anticipate: the `[[ -n ]]` check catches only the empty case, but a trailing inline comment or trailing whitespace on the Makefile line passes straight through — `MACOSX_DEPLOYMENT_TARGET` becomes `14.0 # oldest supported` and the same garbage lands in the OpenSSL cache key. Whether that surfaces as a Configure error or a silently ignored deployment target is up to each toolchain; the one outcome this machinery exists for (the floor actually enforced) is the one you can no longer count on. Verified: `printf 'MACOS_MIN_VERSION = 14.0 # c\n' | sed -n 's/^MACOS_MIN_VERSION[[:space:]]*=[[:space:]]*//p'` prints `14.0 # c`. pkg/Makefile already carries a two-line comment above the assignment; someone moving it inline is an ordinary edit.

Second, this sed + validation block is duplicated byte-identical in mac-artifact.yml — the one piece of new logic with real content (the MACOS_MIN_VERSION parsing contract against pkg/Makefile) is maintained in two places, while this same PR follows the scripts-in-.github/bin convention for the build and the guard. Workflow plumbing duplication (cache keys, if: gates) matches the existing regime between these two workflows and is not the complaint; the parsing contract is.

## Reasoning Chain

1. The sed strips the prefix and prints the rest of the line, whatever it is; only empty results are rejected.
2. Non-empty garbage flows into MACOSX_DEPLOYMENT_TARGET, ./Configure, and the cache key — no loud failure guaranteed.
3. Diffed the two workflows: the floor-read run: body is byte-identical in both; drift between them would be caught only when the divergent copy misbehaves.
4. The repo's own convention for shared shell logic is .github/bin/, which this PR itself uses for the other two scripts.

## Solution

Extract the read into a `.github/bin/macos_min_version.sh` that does the sed plus a shape check, and call it from both workflows:

    [[ "$min_os" =~ ^[0-9]+(\.[0-9]+)?$ ]] || { echo "::error::MACOS_MIN_VERSION malformed: '$min_os'" >&2; exit 1; }

Now every drift in how the Makefile writes the line fails loudly instead of only the drifts you predicted, and the parsing contract lives in one place. If it becomes a script, add it to the OpenSSL cache-key hashFiles alongside build_static_openssl.sh so a parsing change misses the cache. If the team prefers zero new files, at least add the shape check in both copies.

.github/workflows/build-and-release.yml
```diff
@@ -393,15 +393,59 @@ jobs:
...
           # served a stale bundle for every source change within one rc: a
           # pyinstaller-build.spec fix landed, the key was unchanged, `make
           # one-dir` was skipped, and the pre-fix binaries shipped.
-          key: ${{ env.cache-name }}-${{ env.cache-index }}-${{ matrix.os }}-${{ runner.arch }}-${{ steps.system-info.outputs.release }}-${{ env.PYTHON_VERSION }}-${{ needs.get-version.outputs.VERSION }}-${{ hashFiles('pyinstaller-build.spec', 'Pipfile.lock', 'makefile', 'asadm.py', 'lib/**/*.py', 'asinfo/**/*.py') }}
+          # build_static_openssl.sh is a build input too: it decides which
+          # OpenSSL gets linked into _rust.abi3.so, so patching the pinned
+          # version there has to miss this cache or the old bundle ships.
+          key: ${{ env.cache-name }}-${{ env.cache-index }}-${{ matrix.os }}-${{ runner.arch }}-${{ steps.system-info.outputs.release }}-${{ env.PYTHON_VERSION }}-${{ needs.get-version.outputs.VERSION }}-${{ hashFiles('pyinstaller-build.spec', 'Pipfile.lock', 'makefile', 'asadm.py', 'lib/**/*.py', 'asinfo/**/*.py', '.github/bin/build_static_openssl.sh') }}
+
+      - name: Get the macOS support floor
```
pvinh-spike, line 413: From Tests agent:

Severity: minor
Problem Confidence: high
Solution Confidence: medium

## Problem

The PR's green macos-15-intel run proves only the mac-artifact.yml copy of this machinery. build-and-release.yml triggers on workflow_dispatch alone, so this near-duplicate step block — the one that produces the shipping .pkg — has never executed. The two copies differ in exactly the places that bite: working-directory handling and the hashFiles path qualification. "Confirmed by CI" in the PR body overstates what ran.

## Reasoning Chain

1. build-and-release.yml `on:` is workflow_dispatch only; nothing in this PR's checks ran its build-mac-artifacts job.
2. mac-artifact.yml runs on pull_request and is green — but it is a separate transcription of the steps, not the same code.
3. The transcriptions diverge deliberately (subdir checkout vs. root), so one being green is not proof of the other.
4. Everything in the untested copy fails closed (empty min_os errors, missing libcrypto.a errors, guard exits 1), so the worst case is a loud release-build failure, not a bad artifact. That bounds severity at minor.

## Solution

Do a workflow_dispatch dry run (release=false) of build-and-release before this is relied on for a release, rather than discovering a transcription error mid-release. Worth noting the caveat in the PR body too.

.github/bin/check_crypto_static_openssl.sh
```diff
@@ -0,0 +1,55 @@
...
+	exit 1
+fi
+
+checked=0
+violations=0
+while IFS= read -r f; do
+	# Keep "otool could not read it" apart from "it has no OpenSSL dylib".
+	# Piped straight into grep, pipefail makes both look identical, and this
+	# guard would then report success on a build it never actually inspected.
+	if ! out=$(otool -L "$f" 2>&1); then
```
pvinh-spike, line 29: From Tests agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

The guard's entire value is its failure path — dynamic link found, otool broken, no extension found — and none of those paths is exercised by any automated test. CI only ever proves the pass path. The evidence for "fails loudly" is the stub-otool experiment described in the PR body, which lives nowhere but the PR description.

The repo already has the harness for exactly this: `.github/bin/test/pkg_release.bats` tests pkg_release.sh. This is not a demand for a convention the repo lacks — the convention exists; the closest peer (check_min_os.sh) merely predates it untested.

## Reasoning Chain

1. A fail-closed guard that regresses to fail-open (say, a future edit pipes otool into grep and pipefail eats the distinction — the exact trap this script's own comment warns about) passes every green CI run silently.
2. The three failure modes are cheap to pin: a PATH-stubbed otool emitting a libssl.3.dylib line, a failing otool, an empty directory.
3. The author already wrote this test by hand once; it evaporates unless committed.

## Solution

Add `.github/bin/test/check_crypto_static_openssl.bats` with a stub otool on PATH: assert exit 1 + message on a dynamic dep, exit 1 on otool failure, exit 1 on zero files found, exit 0 on a clean static extension. Wire it in next to the existing bats runs.

.github/workflows/build-and-release.yml
```diff
@@ -393,15 +393,59 @@ jobs:
...
+        # cryptography publishes arm64-only macOS wheels from 49.0.0 on, so the
+        # Intel leg builds it from the sdist instead. PyInstaller flattens
+        # vendored dylibs by base name, where Python's own _ssl and _hashlib
+        # already put a libssl.3.dylib, so a dynamically linked cryptography
+        # collides with them and loses. Linking it statically keeps it out of
+        # that collision entirely.
+        if: runner.arch == 'X64' && steps.cache-asadm-asinfo.outputs.cache-hit != 'true'
+        run: |
+          set -euo pipefail
+          # Assigned on its own line: a failing command substitution inside an
```
pvinh-spike: From Style agent:

Severity: minor
Problem Confidence: high
Solution Confidence: high

## Problem

The comment justifies the code against a hazard the code does not have. It says "a failing command substitution inside an argument does not trip `set -e`" — but the line it annotates is `openssl_dir="$RUNNER_TEMP/openssl-static"`, a plain parameter expansion. There is no command substitution anywhere in this step, in either workflow. `$RUNNER_TEMP` cannot "fail"; at worst it's empty, and the next line's `[[ -f ... ]]` guard is what actually catches that.

This is a leftover from the earlier draft that did `openssl_dir="$(brew --prefix ...)"`. As it stands, the comment narrates careful engineering around a nonexistent trap, and the next reader will burn time hunting for the substitution it warns about. And it's pasted into both workflows, so it's wrong twice.

## Reasoning Chain

1. Read the step body in both build-and-release.yml and mac-artifact.yml: the only assignment is `openssl_dir="$RUNNER_TEMP/openssl-static"`.
2. Grepped the step for `$(`: none. No command substitution exists in the step.
3. The comment's stated rationale therefore cannot apply to any line present.
4. The real safety mechanism is the `[[ -f "${openssl_dir}/lib/libcrypto.a" ]]` guard on the following line, which the comment doesn't describe.

## Solution

Delete the three comment lines in both workflows, or replace them with the one thing that's load-bearing:

    # Guard the prefix exists before exporting it: an empty or missing
    # OPENSSL_DIR would reach openssl-sys as a real prefix.

Same edit in both files.

.github/workflows/build-and-release.yml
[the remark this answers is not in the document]
a-spiker: Correcting myself: the reasoning in my reply above was wrong about half of this, and `pkg/Makefile` is now in the key as of e441437.

I argued a floor change was "benign or loudly caught". The raising case is benign as described, but the lowering case is worse than "loud" — it wedges. The bundle cache hits, so every rebuild step is gated off, and `check_min_os.sh` then fails on the stale bundle on every rerun. The rebuild that would fix it is exactly what the cache hit forecloses, so the only exits are bumping `cache-index` or the version, neither of which the error message suggests.

A permanently red build that cannot self-heal is not an acceptable resting state, and one token in the hash removes it. Thanks for the push here.

.github/bin/build_static_openssl.sh
a-spiker: Agreed, and this is the best finding on the PR. Fixed in e441437, partially.

The asymmetry you describe is the part I had not thought through: the arm64 bundle gets OpenSSL fixes for free through the next cryptography bump, so the Intel bundle can sit on a vulnerable pin indefinitely with every gate green and nothing anywhere pointing at it. That is a worse failure shape than a pin that is simply old.

Taken now: the obligation is stated next to `VERSION`/`SHA256`, where someone responding to an OpenSSL advisory is actually standing, including the note that 3.6 is not an LTS line and the pointer to upstream checksums. The cache keys already invalidate on a bump, so the mechanical part needs no ceremony.

Not taken here: the automated staleness signal. I agree it is the real answer and a comment is not a substitute for one. The vehicle depends on org tooling (Renovate regex manager vs a scheduled job querying the `openssl-3.6.*` tags), which is not mine to pick inside this PR. Tracking it separately rather than pretending the comment closes it.

.github/workflows/build-and-release.yml
a-spiker: Fixed in e441437. `pkg/Makefile` added to the hashFiles list.

The wedge analysis is right and better than mine: I had replied on another thread that a floor change was "benign or loudly caught", having only worked through the raising case. The lowering case forecloses its own repair, since the rebuild that would fix the stale bundle is precisely what the cache hit skips. Corrected that thread too.

.github/workflows/build-and-release.yml
a-spiker: Fixed in e441437. Both OpenSSL cache keys now hash `check_min_os.sh` alongside `build_static_openssl.sh`.

The point that carries it is that member minos is unrecoverable post-link, so the build-time scan is not one check among several, it is the only one that will ever look at those archives. Freezing its verdict into a cache keyed on everything except the checker itself was the gap. One extra OpenSSL rebuild per checker change is a fair price.

.github/workflows/build-and-release.yml
a-spiker: Half taken in e441437.

The fail-open parse is real and now closed. Confirmed your repro, then the fix against it:

    14.0                      ACCEPT
    "14.0 # oldest supported" REJECT
    "14.0   "                 REJECT
    ""                        REJECT
    14, 14.0.1                ACCEPT

Same class as two other bugs already fixed on this branch (the guard passing when otool failed, the cache key collapsing to a constant), so "only the empty case is rejected" was the right thing to pull on.

Not taken: extracting to `.github/bin/macos_min_version.sh`. I agree the parsing contract living in two places is the more durable complaint, and the shape check makes the duplication safe rather than absent. Holding it back only because this PR has already grown well past its original scope; it is a better fit for the follow-up that also picks up the bats coverage.

.github/workflows/build-and-release.yml
a-spiker: Correct, and fixed in e441437. Leftover from the draft that did `openssl_dir="$(brew --prefix openssl@3)"`; I changed the line and kept the comment, so it justified the code against a hazard the code no longer had. Replaced in both workflows with what the next line actually does.

.github/workflows/build-and-release.yml
a-spiker: Accepted as a caveat, not fixed. You are right that "confirmed by CI" overstated it: the green run proves the mac-artifact.yml transcription, and the two copies diverge in exactly the spots that bite, subdir checkout and the hashFiles qualification. One of those divergences was in fact a live bug on this branch until ba52736.

Noting it in the PR body rather than leaving the claim standing, and a workflow_dispatch dry run with release=false before this is relied on for a release is the right gate. That needs dispatch rights on the release workflow, so flagging it for whoever cuts the next one rather than silently assuming it works.

.github/bin/check_crypto_static_openssl.sh
a-spiker: Agreed, not taken here. The point that the stub-otool experiment "evaporates unless committed" is exactly right, and this guard has already regressed to fail-open once on this branch, which is the argument for pinning its failure paths rather than its happy path.

Deferring only on scope: `.github/bin/test/` with the bats harness is the right home, and it belongs in the same follow-up as extracting the floor read. Recording it there rather than letting it disappear with the PR description.

pvinh-spike (approved):
LGTM

PR #562, opened 2026-09-17
fix: TOOLS-4445 label 1024-based byte sizes with IEC suffixes (KiB/MiB/GiB/TiB/PiB)

a-spiker:
[TOOLS-4445](https://aerospike.atlassian.net/browse/TOOLS-4445)

## Problem

asadm formats every byte-valued field by dividing by powers of 1024, but labels the result with decimal SI suffixes (`KB`/`MB`/`GB`/`TB`/`PB`). The correct labels for 1024-based divisors are the IEC binary suffixes.

A customer reconciling `License Usage Latest` from a collectinfo read `38.144 TB` as 10^12 bytes and was off by about 10%. The numbers asadm prints have always been correct. Only the unit labels were wrong.

## Change

Byte output is now labelled `KiB`/`MiB`/`GiB`/`TiB`/`PiB`. **Numeric values are unchanged.**

`lib/utils/file_size.py` is the single source of truth, so this reaches every byte-valued surface at once:

- every sheet with a byte column (`info ...`, `show ...`)
- `summary` cluster and namespace storage lines
- license usage lines (latest/min/max/avg)
- the `converted` field in `--json` output
- object size distribution bucket headers on pre-4.2 servers, e.g. `1.000KiB-2.000KiB`

Supporting changes:

- **`" B "` becomes `" B  "`.** Every suffix in the table must be the same width, because sheet column widths are derived from `len()` of the converted string and cells are right-justified. Without the extra pad, `512.000 B` and `1.000 KiB` no longer line up on the decimal point.
- **`byte_verbose` deleted.** It had no callers in any commit in this repo's history, only a slot in `systems`. Removing it also makes the singular/plural branch in `size()` and the nested suffix loop in `is_file_size()` unreachable, so those go too.
- **`is_file_size()` still accepts the old suffixes**, via a new `legacy_byte_suffixes` tuple kept deliberately outside `systems` so it can never be passed to `size()`.
- **`SummaryLicenseLine` Min/Max/Avg now stripped.** It was the only unstripped `file_size.size()` call site, and the wider pad would otherwise have rendered a double space.

## Compatibility

**Nothing in asadm parses asadm-rendered text, so ingesting old data is unaffected.** Verified path by path:

| Path | Reads |
| --- | --- |
| collectinfo analysis | `ascinfo.json` and `.conf` only; the text-section parsers were removed in 2022 |
| log analysis | server logs only, gated on the server log timestamp pattern |
| health checks | raw numerics; the HQL lexer has no unit token |
| on-disk state | none; no pickle or cache, history holds typed commands only |
| `manage` commands | no size-valued argument exists |

Old collectinfo bundles do contain `GB` strings inside the captured text logs, but asadm ignores those files apart from a regex that extracts the version, so mixed-vintage bundles analyse fine.

Worth noting that `is_file_size()` is currently unreachable at runtime. Its only caller sits in the horizontal branch of `Table._update_column_metadata`, and both `Table` instantiations pass vertical orientation. The legacy-suffix handling is therefore defensive rather than load-bearing, but it costs nothing and the function is public.

## Breaking change, needs a release note

Customer tooling that greps asadm output for `" GB"` or `" TB"` will need updating.

`--json` consumers reading the `raw` numeric field are unaffected. Only those reading `converted` as text see the new labels.

Docs on the asadm site show sample output with the old suffixes and should be refreshed separately.

## Testing

```
pytest --disable-warnings test/unit     2241 passed, 1 skipped, 296 subtests
make format-check                       173 files unchanged
flake8                                  4 warnings removed, 0 added
```

Added `test/unit/utils/test_file_size.py`, which covers each tier, the equal-width suffix invariant, and `is_file_size()` for both new and legacy suffixes. That function had no direct coverage before.

Updated the existing unit-string assertions in `test_sheet.py`, `test_templates.py` and `test_view.py`.

Alignment across mixed tiers, rendered from a sample sheet:

```
Node|       Used|      Total
n0  |512.000 B  |  1.000 KiB
n1  |  1.000 MiB|  2.000 GiB
n2  | 38.144 TiB|100.000 TiB
n3  |  0.000 B  |  3.000 PiB
```

Not verified: a live-cluster smoke test. No Aerospike node or Docker was available in my environment, so `summary` and `info namespace usage` against a real cluster are untested. The e2e suite has no byte-string assertions, so it should be unaffected.


dwelch-spike (approved):
LGTM

PR #563, opened 2026-09-17
feat: TOOLS-4157 always collect partition-info in collectinfo

a-spiker:
[TOOLS-4157](https://aerospike.atlassian.net/browse/TOOLS-4157)

## Why

`partition-info` was only collected when someone passed `asadm collectinfo --pmap`, and almost nobody does. A support engineer opening a customer bundle and running `show pmap` got silent empty output: no table, no warning, exit 0. That costs a round-trip to the customer for a re-collection.

pmap is the fastest way to spot four things nothing else in the bundle shows directly: split brain (Cluster Key differing between nodes), data loss (`Dead > 0`), availability outage (`Unavailable > 0`), and partition imbalance.

## What changed

- **Always collected.** pmap moved into gather batch 3 of `_get_collectinfo_data_json`, and the `CollectinfoController.get_pmap` class attribute is gone. Batch 3 rather than 1 or 2 because `_get_pmap_data` is synchronous and blocks the event loop for seconds on a large cluster, and because batch 1 already runs the real statistics sweep that `get_pmap` duplicates.
- **`--pmap` is now a silent no-op.** Still accepted so existing scripts and runbooks keep working, no longer advertised in `--help`. The dead `_confdefault["asadm"]["pmap"]` entry is removed; it never worked from a config file anyway, since `_confspec` has `additionalProperties: false` and no `pmap` key, so `pmap = true` in `astools.conf` has always failed validation.
- **Full error wiring.** New `CollectinfoSection.PMAP`, `_get_as_pmap(nodes, ledger)` records per-node `partition-info` failures in the `NodeErrorLedger`, and pmap gets a retry-table row. Note `_ErrorRecordingCluster` only wraps `info_*` methods, and `get_pmap` reaches `partition-info` through bare `cluster.info(...)`, so the wrapper alone records nothing useful here. That is why `get_pmap` also gained a `keep_exceptions` path, mirroring `GetLatenciesController.get_all`.
- **`ascollectinfo.log` unchanged.** The gate is deleted, but pmap deliberately stays out, honoring TOOLS-1530 (`dc9f192`). Including it would have cost a second full pmap sweep per bundle, since that file is generated by running live `ShowController` commands.
- **Old bundles now say so.** `show pmap` on a pre-5.1.0 bundle logs a warning instead of printing nothing. Follows the existing precedent in the same file (`ShowStatisticsXDRController._check_ns_stats_and_warn`). Warning, not error, so exit code stays 0.

## Two fixes folded in

**Regression this change would otherwise have introduced.** `as_stat_has_aerospike_data` treats any non-`meta_data` stanza with content as proof the node returned data. With pmap always present, a node that lost statistics, config, metadata, histogram and latency but answered one `partition-info` call would have read as healthy, silencing the TOOLS-3596 no-data warning and hiding the node from the analyzer's empty-node diagnostics. pmap is now excluded via `COLLECTINFO_SUPPLEMENTAL_STANZAS`. The fix is in `as_stat_has_aerospike_data`, not `stanza_has_server_data`, because the latter is also used to label a section partial vs empty in `collectinfo_diagnostics`.

**Latent bug this change would have made visible.** `_get_pmap_data` guarded `node_ids[_node]` by key presence only, but the value can be an `Exception` from `cluster.info("node")`. It then flowed into `node_id == working_master`, which is permanently `False`, so a server >= 3.15 node reported 0 master partitions and up to 4096 proles. Rare and opt-in before; it would now be wrong numbers in every bundle. Two-line guard, with a test that fails without it.

## Cost

Per sweep, `partition-info` is roughly 184 KB per namespace per node on the wire, and the parse is ~3 us per row. At 100 nodes and 8 namespaces that is ~147 MB and ~3.3M rows, about 10-13 s of event-loop-blocking CPU per snapshot. Bundle growth is negligible: ~260 bytes per node per namespace, so ~208 KB on that same cluster, under 0.5% of a typical bundle, and it gzips to near nothing.

The duplicate statistics sweep inside `get_pmap` is ~3.7% of the bytes pmap adds. Left alone deliberately: removing it would force pmap out of the gather batches and create a second path into `_get_pmap_data` that live `show pmap` never exercises. If pmap collection time becomes a complaint, the worthwhile follow-up is making the parse non-blocking, not deduplicating stats.

## Testing

2218 unit tests pass, both e2e suites pass (`test_collectinfo.py` 37 passed, `test_all.py` 166 passed), `black --check` clean.

`test_collectinfo.py` now builds its bundle without `--pmap` and still compares live `show pmap` against bundle `show pmap`, which makes the acceptance criterion an assertion. `show pmap` is also no longer excluded from `NOT_IN_CI_MODE` in `test_all.py`.

Verified manually against a live single-node server with 4096 partitions:

1. `collectinfo` with no flags produces `as_stat.pmap` with `master_partition_count: 4096` and the correct cluster key
2. `grep -c "Partition Map Analysis" *ascollectinfo.log` returns 0
3. Bundles collected with and without `--pmap` have identical `as_stat` key sets, and the flag prints no message
4. `show pmap` against the bundle renders the table
5. A bundle with the pmap stanza stripped warns and exits 0
6. `pmap = true` in `astools.conf` is still rejected, unchanged

## Compatibility

Additive only. `ascinfo.json` is unversioned, older analyzers skip unknown `as_stat` keys, and `collectinfo_meta.json` is untouched, so no `meta_format_version` bump. No provenance entry was added to `collection.flags`, since those record what a run was asked to do and pmap is no longer askable.

[TOOLS-4157]: https://aerospike.atlassian.net/browse/TOOLS-4157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/live_cluster/get_controller.py
```diff
@@ -1178,7 +1178,9 @@ def _get_pmap_data(self, pmap_info, ns_info, cluster_keys, node_ids):
 
         return pmap_data
 
-    async def get_pmap(self, nodes="all"):
+    async def get_pmap(self, nodes="all", keep_exceptions=False):
+        """keep_exceptions preserves per-node partition-info Exception values instead
+        of dropping the node key, so collectinfo can record which nodes failed."""
         getter = GetStatisticsController(self.cluster)
         service_stats = asyncio.create_task(getter.get_service(nodes=nodes))
```
dwelch-spike: From Logic agent:

`get_service` drops a node whose `info_statistics` failed, so that node is missing from `cluster_keys`. If its `info_namespace_statistics` succeeded, `_get_namespace_data` raises `KeyError` on `cluster_keys[node]`. This predates the PR, but `get_pmap` now runs in batch 3 of every collectinfo, where the exception escapes the `gather` and the whole snapshot is lost (no `ascinfo.json` at all on snapshot 1). Passing `keep_exceptions=True` makes the existing `"N/E"` branch below reachable: the node keeps its partition counts with cluster key `N/E`, and `_ErrorRecordingCluster` already records the service failure under `pmap`/`service`. Worth adding a test where only service stats fail; `NoDataWarningProductionShapeTest` fails both calls for B, so it doesn't cover this case.

```suggestion
        service_stats = asyncio.create_task(
            getter.get_service(nodes=nodes, keep_exceptions=True)
        )
```

lib/live_cluster/get_controller.py
```diff
@@ -1066,7 +1066,7 @@ def _get_pmap_data(self, pmap_info, ns_info, cluster_keys, node_ids):
             ck = cluster_keys[_node]
             node_id = node_ids[_node]
 
-            if isinstance(partitions, Exception):
+            if isinstance(partitions, Exception) or isinstance(node_id, Exception):
```
dwelch-spike: From Logic agent:

When `info("node")` fails for a node, this guard drops the node without recording why. `cluster.info` isn't an `info_*` method, so `_ErrorRecordingCluster` never sees it, and the `keep_exceptions` loop in `get_pmap` only puts back `partition-info` exceptions. The node silently disappears from `as_stat.pmap` with no ledger entry, so a timeout is never retried and the diagnostics never mention it. Surface `node_ids` exceptions through the `keep_exceptions` path as well, and have `get_pmap` report which call failed so `_get_as_pmap` doesn't hardcode `detail="partition-info"`.

lib/collectinfo_analyzer/show_controller.py
```diff
@@ -1428,17 +1428,24 @@ def __init__(self):
...
+            rendered = True
             self.view.show_pmap(
                 pmap_data[timestamp],
                 self.log_handler.get_cinfo_log_at(timestamp=timestamp),
                 timestamp=timestamp,
             )
 
+        if not rendered:
+            logger.warning(
+                "Partition map data was not written to collectinfo files by asadm before 5.1.0."
```
dwelch-spike: From Logic agent:

Now that pmap is always collected, an empty pmap in a 5.1.0+ bundle means collection failed on every node, but this message blames the bundle's asadm version. Use neutral wording, e.g. "No partition map data in this bundle. asadm before 5.1.0 did not collect it by default.", or choose the message based on the bundle's asadm version or its recorded errors.

a-spiker: Done

lib/live_cluster/get_controller.py
a-spiker: Done

dwelch-spike (approved):
Nice! LGTM

